PR #7, opened 2024-09-20
Fix infrahub/#8

PhillSimonds:
Make infrahubctl transform and render commands use InfrahubClient to communicate with GQL API.


PhillSimonds:
Happy to add some unit/integration tests here if beneficial.

infrahub_sdk/ctl/cli_commands.py
```diff
@@ -187,19 +195,34 @@ def render_jinja2_template(template_path: Path, variables: dict[str, str], data:
 
 
 def _run_transform(
-    query: str,
+    query_name: str,
+    client: InfrahubClient,
```
wvandeun: Does it makes sense to pass an InfrahubClient instance into this method? The only reason we seem to need this is for the graphql query. If we use the `run` method of the transform instead, then it will handle querying the data for us.

The type you are using here is the async version of the client, but this is a sync function, and we are calling a sync method on the `client`, so it should probably be `InfrahubClientSync`. When you are calling this function, you seem to be passing in the sync version of the client.

infrahub_sdk/transforms.py
```diff
@@ -92,8 +92,20 @@ async def run(self, data: Optional[dict] = None) -> Any:
 
 
 def get_transform_class_instance(
-    transform_config: InfrahubPythonTransformConfig, search_path: Optional[Path] = None
+    transform_config: InfrahubPythonTransformConfig,
+    search_path: Optional[Path] = None,
+    client: Optional[InfrahubClient] = None,
```
wvandeun: `InfrahubClientSync`

infrahub_sdk/transforms.py
```diff
@@ -108,7 +120,8 @@ def get_transform_class_instance(
         transform_class = getattr(module, transform_config.class_name)
 
         # Create an instance of the class
-        transform_instance = transform_class()
+        transform_instance = asyncio.run(transform_class.init(client=client, branch=branch))
```
wvandeun: We are deprecating the usage of the `init` method on different classes in the SDK. We should probably not use this method any longer (we cannot remove it yet, as the backend Infrahub code is still leveraging it). Instead we should consider changing the __init__ method.

wvandeun (changes requested):
@PhillSimonds, first of all,  Thank you for your contribution!
I left some comments. If they are not clear to you, please reach out.

infrahub_sdk/ctl/cli_commands.py
PhillSimonds: I originally went to implement the run method on the transform, but it ended up querying a stored graphql query endpoint (`f"{self.address}/api/query/{name}"`) which doesn't exist when you are running the command from the CLI on a git repo that hasn't been synced with infrahub. 

https://github.com/opsmill/infrahub-sdk-python/blob/ba76af1011579d25320b1d87bd988d5c3af1f513/infrahub_sdk/client.py#L774

The `.execute_graphql` method on the InfrahubClient object queries the graphQL API directly, rather than using a stored query.

https://github.com/opsmill/infrahub-sdk-python/blob/ba76af1011579d25320b1d87bd988d5c3af1f513/infrahub_sdk/client.py#L207

I don't think we can guarantee the stored query path will exist when running the infrahubctl command directly, but I may be making bad assumptions.

infrahub_sdk/transforms.py
PhillSimonds: Acknowledged. Looking at the method, there is actually no asynchronous code in the async `init` method. The only thing the `init` method adds over `__init__` is the creation of a client.

I added a `@cached_property` (from functools) of `client` that initializes the client in the same way this init method does. This will cause it to be created when called, so we don't have to include it in the init logic.

I then instantiated the `TransformClass` directly using it's __init__ method synchronously so that we don't call that `init` method.

The only oddity here is that the client is a `InfrahubClient` instead of an `InfrahubClientSync`. I added a block of code in the `_run_transform` method to determine which type of object we're dealing with (as the `infrahubctl render` command uses the synchronous version) and run it correctly.

PhillSimonds:
Thanks for the comments Wim. I've added to the comments and updated per your feedback. I also renamed the town crier doc to reflect the new issue number now that the issue has been moved to this repo.

Let me know your thoughts on changing the transform run logic vs doing something else. Happy to update if there is a better way.

Also, this CLI command was broken when I picked up the PR. Would you like me to add an integration test to protect this in the future?

infrahub_sdk/ctl/cli_commands.py
PhillSimonds: I ended up updating the transform so it tries the stored query API endpoint first then tries the graphQL API directly afterwards.

PhillSimonds:
Wim. I updated the InfrahubTransform so that it will try to query the stored graphql query path first and query the graphql API directly in the event the stored query doesn't exist. That allows us to just call .run() on the transform class. I think that's the most elegant way to do this.

Another thought here -- there seems to be some confusion in the InfrahubTransform class's understanding of "branch". It's getting the branch name from the local repo if a branch name isn't passed in, but the branch name is actually used to ask for data from the specified branch in *infrahub*. I believe the repo branch and infrahub branch are completely separate, so we probably shouldn't be using a repo object to look at the local branch then pass that in the API request. the `.branch_name` property in the InfrahubTransform is only used in calls to the infrahub API from everything I see inside the class. I guess some external component could be using it to introspect the branch name of the local branch. Not sure if this should be remediated.

petercrocker:
@opsmill/backend please review this one from @PhillSimonds !

ogenstad:
Hey @PhillSimonds, thank you for your contribution!

> Also, this CLI command was broken when I picked up the PR. Would you like me to add an integration test to protect this in the future?

More testing overall would be great :) The CTL specifically is missing test coverage. In what way was the command currently broken? 

Looking through the code to initialize transforms I realize that we need to do some refactoring with regards to how it's done as we have multiple ways of doing it at present. There is a small caveat with the PR you've done with regards to how the ctl currently expects a client to be setup in the way that the config would look a bit different if we used the client returned by a call to `initialize_client()`

As this currently targeting the `stable` branch and I think larger changes are needed I think it would be good to lower the scope and number of changes on this one. I'll open another issue to describe some thoughts I had in mind for these changes but they should probably be done against `develop` as this is more of a small bugfix.

It would be impossible to know but we're going to remove the part with `server_url` as that's never used within the codebase today.

I think what we need to do here is to change the initializer: 

```python
class InfrahubTransform:
    name: Optional[str] = None
    query: str
    timeout: int = 10

    def __init__(self, branch: str = "", root_directory: str = "", server_url: str = "", client: Optional[InfrahubClient] = None):
        self._client: InfrahubClient

```

Then have a property for .client which currently acts as a setter/getter (later the client will be a required argument). Here the same thing could be done with `self.git` as it looked like you find a bug around that if that was the issue?

Then here:

```python
    try:
        transform_instance = get_transform_class_instance(transform_config=transform_config)
    except InfrahubTransformNotFoundError as exc:
        console.print(f"Unable to load {transform_name} from python_transforms")
        raise typer.Exit(1) from exc
```

We'd add a:

```python
transform_instance.client = the-client-object
``` 

A problem here with the current code could be that initialize_client is an async function even though there's no need for it to be. We should be able to change it to a sync function. Alternatively we can cheat and run initialize_client_sync and convert it:

```python
sync_client = initialize_client_sync()
async_client = InfrahubClient(config=sync_client.config)
```

Does this make sense to you?

PhillSimonds:
Happy to contribute :).

> More testing overall would be great :) The CTL specifically is missing test coverage.

Awesome, I'll add some tests.

> In what way was the command currently broken?

Execution of the graphQL query was failing because the query wasn't loaded into the repository config object, and thus `get_query($name)` could not get the query.

*Further calls up the stack trace omitted for brevity
```
  File ".../infrahub-sdk-python/infrahub_sdk/ctl/utils.py", line 106, in
execute_graphql_query
    query_object = repository_config.get_query(name=query)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../infrahub-sdk-python/infrahub_sdk/schema.py", line 228, in
get_query
    return self._get_resource(resource_id=name, resource_type=InfrahubRepositoryGraphQLConfig)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../infrahub_sdk/schema.py", line 192, in
_get_resource
    raise KeyError(f"Unable to find {resource_id!r} in {RESOURCE_MAP!r}")
KeyError: "Unable to find 'tags_query' in 'queries'"
```

> Here the same thing could be done with self.git as it looked like you find a bug around that if that was the issue?

The `.git` attribute wasn't actually causing an issue on execution of code because we were passing a branch name in, so the block that would have caused an issue with accessing `.git` wasn't being hit even though it's a bug that could blow things up. There are two things I think we should consider with this object:

1. a `.git` attribute doesn't always exist (the bug I found). Because of this the `branch_name` property on the `InfrahubTransform` would have raised an error if no branch_name was passed in at object instantiation because it would have tried to access .git in order to acquire it. I fixed this by modifying the branch_name property such that it uses `hasattr(self, git)` to determine if the git attribute is specified before trying to determine if it's set to None (and then instantiate a new Repo() object if this is the case). This is probably better fixed by the solution you recommended, making git an optional parameter we can pass in at object instantiation, then using the property as a getter/setter.

2. The logic behind this git object doesn't make sense to me and seems like it *could* cause an issue with code execution at some point. The .git attribute is a local `Repo()` object, but it is being used as if it were the branch_name to query in infrahub. To me, these should be two completely separate things -- but I may be misunderstanding something here.

**I filed a separate issue for the git stuff, #49. I am thinking it might make sense to create a separate PR for its full resolution, but I'm happy to integrate into this one if you think it would be better.

> Does this make sense to you?

I think I'm following. I'll make some updates here shortly then ping for further review :)

PhillSimonds:
@ogenstad I've made the following updates:

- InfrahubTransform now takes an optional client argument (of type `infrahubClient`) in it's __init__ method. This is assigned to `._client`
- I've modified `.client` from a cached_property to a standard property. It gets `._client` if `._client` exists, otherwise it sets it to a new instance of `InfrahubClient` using the server URL
- the `initialize_client` utility has been modified to make it synchronous. I changed all places in the code where this was called by removing the await keyword. This was done using grep. Mypy also flagged errors in the event the newly synchronous function call was being awaited instead of called directly. These two checks in mind -- I'm pretty sure I caught and modified all calls to `initialize_client`
- We now create a client using `initialize_client` in the `infrahubctl transform` command, then pass that client into `InfrahubTransform` at instantiation. I like doing it this way more then assigning the `.client` attribute post-instantiation as we're not overriding the getter/setter this way.

infrahub_sdk/transforms.py
```diff
@@ -25,33 +27,49 @@ class InfrahubTransform:
...
+        if not self._client:
+            self._client = InfrahubClient(address=self.server_url)
+
+        return self._client
+
     @classmethod
     async def init(cls, client: Optional[InfrahubClient] = None, *args: Any, **kwargs: Any) -> InfrahubTransform:
         """Async init method, If an existing InfrahubClient client hasn't been provided, one will be created automatically."""
+        warnings.warn(
+            "InfrahubClient.init has been deprecated and will be removed in Infrahub SDK 0.14.0 or the next major version",
```
ogenstad: Sorry about the delay around reviewing this will take a closer look soon! Just a quick note about this which looks like a typo. Even if it's a minor detail this shouldn't be `InfrahubClient.init` it should be either `InfrahubTransform.init` or `{cls.__class__.__name__}.init`

infrahub_sdk/ctl/cli_commands.py
```diff
@@ -187,19 +194,35 @@ def render_jinja2_template(template_path: Path, variables: dict[str, str], data:
...
     repository_config: InfrahubRepositoryConfig,
 ):
+    """
+    Query GraphQL for the required data then run a transform on that data.
+
+    Args:
+        query_name: Name of the query to load (e.g. tags_query)
+        variables: Dictionary of variables used for graphql query
+        transformer_func: The function responsible for transforming data received from graphql
+        transform: A function used to transform the return from the graphql query into a different form
```
ogenstad: The arguments provided here don't match the function signature, I'm guessing due to renaming things, we should replace "transformer_func" & "transform" -> transform_func that is used in the signature.

infrahub_sdk/ctl/cli_commands.py
```diff
@@ -187,19 +194,35 @@ def render_jinja2_template(template_path: Path, variables: dict[str, str], data:
...
 
     try:
         response = execute_graphql_query(
-            query=query, variables_dict=variables, branch=branch, debug=debug, repository_config=repository_config
+            query=query_name, variables_dict=variables, branch=branch, debug=debug, repository_config=repository_config
         )
+
+        if debug:
+            message = ("-" * 40, f"Response for GraphQL Query {query_name}", response, "-" * 40)
+            console.print("\n".join(message))
```
ogenstad: I think this might be copied from some other location but unfortunately it's not valid :( 

I wish the linters would catch this but haven't dug into it. But the join expects an `Iterable[str]` or similar and `response` is a dict. An example of what this looks like when it fails:

```python
>>> query_name = "my_query"
>>> response = {"data": {"ok": True}}
>>> message = ("-" * 40, f"Response for GraphQL Query {query_name}", response, "-" * 40)
>>> print("\n".join(message))
TypeError: sequence item 2: expected str instance, dict found
>>>
```

I think for now we can perhaps just remove this extra `if debug:` and come back to it in another PR and start by fixing the broken examples.

infrahub_sdk/ctl/cli_commands.py
```diff
@@ -288,30 +322,39 @@ def transform(
...
-    result = _run_transform(
-        query=transform_instance.query,
-        variables=variables_dict,
-        transformer=transformer,
-        branch=branch,
-        debug=debug,
-        repository_config=repository_config,
-    )
+    # Load query config
+    query_config_obj = InfrahubRepositoryGraphQLConfig(name=transform.query, file_path=Path(transform.query + ".gql"))
```
ogenstad: I don't think we can assume that the file_path will match the name of the query exactly. This would work if the query lives in the root of the repo. But not if it lived under other directories such as ./queries, ./transforms etc.

infrahub_sdk/transforms.py
ogenstad: Also we should bump the version a bit here perhaps just to 0.15.0 as we are just releasing 0.14.0. Preferably we should also add a towncrier newsfragement to indicate that this is now marked for deprecation.

infrahub_sdk/transforms.py
```diff
@@ -72,17 +90,40 @@ def branch_name(self) -> str:
...
-    async def run(self, data: Optional[dict] = None) -> Any:
+        # Try getting data from stored graphql query endpoint
+        try:
+            return await self.client.query_gql_query(name=self.query, branch_name=self.branch_name, variables=variables)
+        # If we run into an error, the stored graphql query may not exist. Try to query the GraphQL API directly instead.
+        except httpx.HTTPStatusError:
+            if not self.repository_config:
+                raise
+            query_str = self.repository_config.get_query(name=self.query).load_query()
+            return await self.client.execute_graphql(query=query_str, variables=variables, branch_name=self.branch_name)
```
ogenstad: I'd like to revert to the old behaviour here so that the focus of this PR is only to add the missing `InfrahubClient` that was the target of #8.

While I can see that the collect_data is missing some information as it is now I'm not sure that we should add a `InfrahubRepositoryConfig` object to the transform itself in order to load a missing query. I think we should handle all that in a different way. I'm curious to the root cause of this if you had problems running with a local query if it had to do with the fact that the query wasn't defined within your `.infrahub.yml` file? This is my assumption based on the `repository_config.queries.append(query_config_obj)` could this be correct?

ogenstad (changes requested):
For now I'd like to remove some of the additions in this PR and focus on the main cause we're trying to solve with the bug and instead raise other issues to discuss other improvements such as the loading of queries if that is currently problematic.

Feel free to reach out if you want a quick sync around this!

infrahub_sdk/transforms.py
PhillSimonds: The query was indeed not in my infrahub.yml file (hence the loading of the query in L#352-354 of `cli_commands.py`, which can be removed with this in mind) -- but that isn't the root of the issue with running the transform here.

The InfrahubTransform object in stable only has access to the query name, not to the query string. The query name is able to be used in the front end because a stored query at the path `f"{self.address}/api/query/{name}" ` pre-exists in infrahub which returns data, so we don't actually need the query string to make the query, just the name of the query. I think this stored query path is created as part of syncing a repo (though I'm not sure). This is not the case when the `infrahubctl transform` command is run, because the stored query path does not pre-exist in infrahub (unless I'm doing something wrong here). This in mind, we need the full query string to execute the query with `infrahubctl transform`, not just the query name.

```
HTTP communication failure: Client error '403 Forbidden' for url 'http://localhost:8000/api/query/tags_query?branch=main&update_group=false'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403 on POST to 
http://localhost:8000/api/query/tags_query?branch=main&update_group=false
```

I can add a client object to the transform (as per #8), but I'll have to call the client object's `.execute_graphql` method directly with the query string in order to acquire data, circumventing the transform's run method for data acquisition. I can then run the transform with that data. Is this what you're thinking would be best to resolve #8 with a more modest touch?

infrahub_sdk/ctl/cli_commands.py
PhillSimonds: Ack, I'll remove :)

infrahub_sdk/ctl/cli_commands.py
PhillSimonds: Aye good catch. Sorry, re-wrote that a few times.

infrahub_sdk/transforms.py
PhillSimonds: I updated per my comment there. If this isn't what you're thinking (or I'm missing something) LMK and I can modify.

infrahub_sdk/ctl/cli_commands.py
```diff
@@ -187,19 +193,35 @@ def render_jinja2_template(template_path: Path, variables: dict[str, str], data:
...
     debug: bool,
     repository_config: InfrahubRepositoryConfig,
 ):
+    """
+    Query GraphQL for the required data then run a transform on that data.
+
+    Args:
+        query_name: Name of the query to load (e.g. tags_query)
+        variables: Dictionary of variables used for graphql query
+        transformer_func: The function responsible for transforming data received from graphql
```
ogenstad: This one still has a slight typo. transformer_func -> transform_func.

ogenstad (commented):
Thanks for the updates. I left a minor comment about a slight typo.

I also opened #54, which I think might have been some source of confusion for you.

ogenstad:
Regarding the debug: #56

PhillSimonds:
@ogenstad I've added some tests for the `infrahubctl transform` command. I believe this should be good to go, let me know if I need to add or remove something more.

ogenstad (approved):
Thank you for the changes!

PR #41, opened 2024-09-21
Remove local validation of query filters when calling client.filters()

### LINKED ISSUE #9: feature: Remove the local validation of the query filters when calling client.filters()
### Component

Python SDK

Currently in the python SDK we are automatically validating the query filters passed to `client.filters()` based on the information provided in the schema.

This local validation has been the root cause of a few issues recently because the filters provided as part of the schema aren't  100% identical to the filters supported by the GraphQL api.

Since the GraphQL API will validate the query filters anyway the proposal is to completely remove the local validation of the filters as currently implemented. 

In the future if we want to have some local validation, we should probably investigate how to retrive the graphql schema directly and do a validation of the graphql query based on the schema 

### Describe the Use Case

Improve the user experience and ensure that all filters available in the graphql api are available in the python SDK

### Additional Information

_No response_

dgarros:
Fixes #9 #10 #30

Remove the local validation of the query filters when calling client.filters()

> I completely removed the code since the filters will be removed from infrahub in the next version


wvandeun (approved):

ogenstad (approved):
LGTM, can we also delete the `FilterNotFoundError` exception entirely from the exceptions file?

dgarros:
> LGTM, can we also delete the `FilterNotFoundError` exception entirely from the exceptions file?

Good point, I removed `FilterNotFoundError` completely

PR #47, opened 2024-09-29
Fix `execute_graphql` method not respecting clien's `default_branch` property

### LINKED ISSUE #46: bug: `execute_graphql` does not consider `default_branch` setting on `InfrahubClient`
### Component

Python SDK

### Infrahub SDK version

0.13.1

### Current Behavior

The `execute_graphql` method of `InfrahubClient` does not consider the `default_branch` setting on `InfrahubClient`.

### Expected Behavior

`execute_graphql` executes the the graphql query in the branch that is set as `default_branch`

### Steps to Reproduce

- Spin up an instance of Infrahub
- Create a branch `test`
- Create a `BuiltinTag` node in the `test` branch
- Run the following script
```python
from infrahub_sdk import InfrahubClientSync, Config
client = InfrahubClientSync(config=Config(default_branch="test"))
client.execute_graphql(query="query { BuiltinTag { count }}") # result count will be 0
client.execute_graphql(query="query { BuiltinTag { count }}", branch_name="test") # result count will be 1
```

### Additional Information

The `branch_name` argument that you can use seems to deviate from how we typically name this argument in other methods in the SDK. Typically the name would be `branch`.

wvandeun:
fixes #46

### DIFF
--- a/changelog/46.fixed.md
+++ b/changelog/46.fixed.md
@@ -1,0 +1,1 @@
+`execute_graphql` method for InfrahubClient(Sync) now properly considers the `default_branch` setting
--- a/infrahub_sdk/client.py
+++ b/infrahub_sdk/client.py
@@ -575,6 +575,7 @@ class InfrahubClient(BaseClient):
             _type_: _description_
         """
 
+        branch_name = branch_name or self.default_branch
         url = self._graphql_url(branch_name=branch_name, at=at)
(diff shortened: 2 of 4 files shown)


ogenstad (approved):
LGTM, can you add a newsfragment that indicates that the bug was fixed?

PR #61, opened 2024-10-05
use DiffTreeSummary instead of deprecated DiffSummary query

ajtmccarty:
IFC-727 
infrahub PR https://github.com/opsmill/infrahub/pull/4546

use the new DiffTreeSummary query instead of the deprecated DiffSummary graphql query in `get_diff_summary`

### DIFF
--- a/infrahub_sdk/diff.py
+++ b/infrahub_sdk/diff.py
@@ -1,0 +1,128 @@
+from __future__ import annotations
+
+from typing import (
+    Any,
+)
+
+from typing_extensions import NotRequired, TypedDict
+
+
+class NodeDiff(TypedDict):
+    branch: str
+    kind: str
+    id: str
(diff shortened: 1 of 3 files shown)


infrahub_sdk/diff.py
```diff
@@ -0,0 +1,128 @@
+from __future__ import annotations
+
+from typing import (
+    Any,
+)
+
+from typing_extensions import NotRequired, TypedDict
+
+
+class NodeDiff(TypedDict):
```
ogenstad, line 10: FYI we have a:

```python
from infrahub_sdk.client import NodeDiff  # noqa: TCH002
```

In infrahub/message_bus/types.py, it would still work since we are importing NodeDiff into infrahub_sdk/client but it gets a bit weird.

PR #67, opened 2024-10-11
Upload coverage reports to codecov

ogenstad:
* Add reporting to codecov. We'll check to see how much comments codecov provides on the PR and potentially disable or limit that later if it's verbose. It seems like this needs to be merged into the default branch in order to properly test it.


dgarros (approved):

fatih-acar:
cc @wartraxx51

PR #71, opened 2024-10-11
Add Python 3.13

ogenstad:


dgarros:
Would be good to update this PR to see if it's working now
Since 18.0 PyArrow added the support for 3.13

ogenstad:
The pipeline succeeds now after allowing later versions of pyarrow. Doesn't look like we have much testing around this area of the code though.

pyproject.toml
```diff
@@ -43,7 +44,7 @@ numpy = [
     { version = "^1.24.2", optional = true, python = ">=3.9,<3.12" },
     { version = "^1.26.2", optional = true, python = ">=3.12" },
 ]
-pyarrow = { version = "^14", optional = true }
+pyarrow = { version = ">=14,<19", optional = true }
```
dgarros: For the SDK I think it's best if we don't add a upper limit for our dependencies unless we have a good reason to.

gmazoyer (approved):
[This](https://github.com/opsmill/infrahub-sdk-python/blob/9de2be34a618995b0eff493ed29f7381ee12cc51/infrahub_sdk/transfer/importer/json.py#L58) is the only place where we use pyarrow. So it should be fairly safe to upgrade. We used to have [integration tests](https://github.com/opsmill/infrahub-sdk-python/blob/develop/tests/integration/test_export_import.py) to cover this, but I don't think we run these tests for now.

pyproject.toml
ogenstad: This was just mainly to keep the current type of behaviour but I agree we could use this instead.

```toml
pyarrow = { version = ">=14", optional = true }
```

ogenstad:
Changed the versioning constraint and added a newsfragment.

dgarros (approved):

PR #73, opened 2024-10-12
Add `infrahubctl menu load` command to load menu items from a file

dgarros:
This PR adds a new cli command `infrahubctl menu load` to load menu items from one or multiple files
> This effort is part of the upcoming release of Infrahub and it is required to load the new menu items

The menu file need to have the following format.

in Summary
- The first 4 lines are mandatory
- The format under `data` must be compatible with `CoreMenuItem`
- If `kind` is provided, the path will be automatically generated to `/objects/<kind>`

```yaml
---
apiVersion: infrahub.app/v1
kind: Menu
spec:
  data:
  - namespace: Location
    name: MainMenu
    label: Location
    icon: "mingcute:location-line"
    children:
      data:
      - namespace: Location
        name: ContinentMenu
        label: Continent
        kind: LocationContinent
        icon: "jam:world"
        order_weight: 1000
```

## More than just the menu

Because, it felt silly to create yet another specific command for the menu, I tried to make it as generic as possible.
With a tiny difference, this format should work for all type of objects in the near future.
I've also added the command `infrahubctl object load` (hidden for now) to explore this feature.

I was able to successfully load this file with `location` data but more work is required to properly support all types of relationships

```yaml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: LocationContinent
  data:
  - name: North America
    children:
      kind: LocationCountry
      data:
      - name: "United States of America"
      - name: "Canada"
  - name: South America
    children:
      kind: LocationCountry
      data:
      - name: "Mexico"
      - name: "Brazil"
```

### DIFF
--- a/infrahub_sdk/ctl/menu.py
+++ b/infrahub_sdk/ctl/menu.py
@@ -40,13 +40,8 @@ async def load(
     files = load_yamlfile_from_disk_and_exit(paths=menus, file_type=MenuFile, console=console)
     client = await initialize_client()
 
-    default_kind = "CoreMenuItem"
-
     for file in files:
         file.validate_content()
-        if not file.spec.kind:
-            file.spec.kind = default_kind
-
         schema = await client.schema.get(kind=file.spec.kind, branch=branch)
 
         for idx, item in enumerate(file.spec.data):
@@ -55,6 +50,6 @@ async def load(
                 schema=schema,
                 data=item,
                 branch=branch,
-                default_schema_kind=default_kind,
+                default_schema_kind=file.spec.kind,
                 context={"list_index": idx},
             )
--- a/infrahub_sdk/ctl/menu.py
+++ b/infrahub_sdk/ctl/menu.py
@@ -1,0 +1,60 @@
+import logging
+from pathlib import Path
+
+import typer
+from rich.console import Console
+
+from infrahub_sdk.async_typer import AsyncTyper
+from infrahub_sdk.ctl.client import initialize_client
+from infrahub_sdk.ctl.utils import catch_exception, init_logging
+from infrahub_sdk.spec.menu import MenuFile
+
+from .parameters import CONFIG_PARAM
+from .utils import load_yamlfile_from_disk_and_exit
+
+app = AsyncTyper()
+console = Console()
+
+
+@app.callback()
+def callback() -> None:
+    """
+    Manage the menu in a remote Infrahub instance.
+    """
+
+
+@app.command()
+@catch_exception(console=console)
+async def load(
+    menus: list[Path],
+    debug: bool = False,
+    branch: str = typer.Option("main", help="Branch on which to load the menu."),
+    _: str = CONFIG_PARAM,
+) -> None:
+    """Load one or multiple menu files into Infrahub."""
+
+    init_logging(debug=debug)
+
+    logging.getLogger("infrahub_sdk").setLevel(logging.INFO)
+
+    files = load_yamlfile_from_disk_and_exit(paths=menus, file_type=MenuFile, console=console)
+    client = await initialize_client()
+
+    default_kind = "CoreMenuItem"
+
+    for file in files:
+        file.validate_content()
+        if not file.spec.kind:
+            file.spec.kind = default_kind
+
+        schema = await client.schema.get(kind=file.spec.kind, branch=branch)
+
+        for idx, item in enumerate(file.spec.data):
+            await file.spec.create_node(
+                client=client,
+                schema=schema,
+                data=item,
+                branch=branch,
+                default_schema_kind=default_kind,
+                context={"list_index": idx},
+            )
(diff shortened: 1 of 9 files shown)


infrahub_sdk/ctl/menu.py
```diff
@@ -0,0 +1,60 @@
...
+
+    for file in files:
+        file.validate_content()
+        if not file.spec.kind:
+            file.spec.kind = default_kind
+
+        schema = await client.schema.get(kind=file.spec.kind, branch=branch)
+
+        for idx, item in enumerate(file.spec.data):
+            await file.spec.create_node(
```
ogenstad, line 48: Not related to the PR but just a note that it's very annoying the way codecov comments on every line like this instead of specifying a range of changed lines that aren't covered by tests. I like the idea but the execution :/ Hopefully that's configurable.

infrahub_sdk/ctl/object.py
```diff
@@ -0,0 +1,48 @@
...
+
+    init_logging(debug=debug)
+
+    files = load_yamlfile_from_disk_and_exit(paths=paths, file_type=ObjectFile, console=console)
+    client = await initialize_client()
+
+    for file in files:
+        file.validate_content()
+        if not file.spec.kind:
+            raise FileNotValidError(name=str(file.location), message="kind must be specified.")
```
ogenstad: Can we have this happen within .validate_content() instead?

infrahub_sdk/spec/object.py
```diff
@@ -0,0 +1,128 @@
...
+class InfrahubObjectFileData(BaseModel):
+    kind: Optional[str] = None
+    data: list[dict[str, Any]] = Field(default_factory=list)
+
+    @classmethod
+    def enrich_node(cls, data: dict, context: dict) -> dict:
+        return data
+
+    @classmethod
+    async def create_node(
```
ogenstad, line 19: In what way has this been used? I think it might be fine to create the menu but I'm wondering if we need to rethink the structure for other object types where the object we create is related to other objects. If those objects doesn't yet exist I think we'd need to create them first before creating the "base" object. Should be doable by looking at the relationships and which side if any is mandatory. Though it would also be nice if we could just feed everything to the backend as a "fix this in a single transaction" type of way. Not a blocker here and it could be that I don't completely understand how this gets used but I see some situations where it seems problematic now depending on how the relationships are defined.

infrahub_sdk/spec/object.py
```diff
@@ -0,0 +1,128 @@
...
+                raise ValueError(
+                    f"Relationship {rel_schema.name} doesn't have the right format {rel_schema.cardinality} / {type(rel_data)}"
+                )
+
+
+class ObjectFile(InfrahubFile):
+    _spec: Optional[InfrahubObjectFileData] = None
+
+    @property
+    def spec(self) -> InfrahubObjectFileData:
```
ogenstad, line 119: Just thinking out loud here if we're later adding a spec property to other object types and want to look at the spec for a generic `InfrahubFile` object it might get problematic if we always need to override the returned data in a different way for each given subclass (in this case `ObjectFile` which would return an `InfrahubObjectFileData`. Depending on what we do in the future it might not be a problem, alternatively we could look at generics and see if we could fix it with that.

infrahub_sdk/yaml.py
```diff
@@ -23,3 +46,52 @@ def load_content(self) -> None:
         if not self.content:
             self.error_message = "Empty YAML/JSON file"
             self.valid = False
+
+    def validate_content(self) -> None:
+        pass
+
+    @classmethod
+    def load_from_disk(cls, paths: list[Path]) -> list[Self]:
```
ogenstad, line 54: We currently read most of the data from disk as a sync function, i.e. the jinja2 templates when used from within the git-agent and also even the loading of .infrahub.yml files. I see this being used within the git-agents as well. Even though we don't have the file loaders as async now I'm wondering if this should be an async function from the start, or if we just introduce a new one that is async once we actually read the files from disk in an async way.

ogenstad (approved):
LGTM. Added some comments which is mostly me just thinking out loud.

infrahub_sdk/spec/object.py
dgarros: Yes I fully agree, this is not the final implementation
My end goal is to be able to create all these objects in memory and to have a engine that will know in which order they need to be created based on the schema.
This is very similar to what we need to build better generators.

infrahub_sdk/spec/object.py
dgarros: Agreed, not a fan of this implementation as well, happy to consider a better/cleaner solution in the future

infrahub_sdk/yaml.py
dgarros: Since it's sync for now, my vote is to keep it as a sync function but it would be good to start using `anyio.Path` to properly support Async for all disk access. I'll open an issue to track that

PR #80, opened 2024-10-15
Convert import paths to use relative Package name

dgarros:
Related to https://github.com/opsmill/infrahub/issues/4234

In order to use a different package name within Infrahub, we need to convert the internal hardcoded import paths into relative paths.

### DIFF
--- a/infrahub_sdk/__init__.py
+++ b/infrahub_sdk/__init__.py
@@ -2,12 +2,12 @@ from __future__ import annotations
 
 import importlib.metadata
 
-from infrahub_sdk.analyzer import GraphQLOperation, GraphQLQueryAnalyzer, GraphQLQueryVariable
-from infrahub_sdk.batch import InfrahubBatch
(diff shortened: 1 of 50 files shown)


ogenstad:
I think we might run into issues with this:

https://github.com/opsmill/infrahub-sdk-python/blob/develop/infrahub_sdk/schema.py#L111-L112

When we import the generators from the Git Agent and run this check on the imported objects we're probably going to get a mismatch. I still think we can move forward with this PR, but potentially we'll need to switch to using a `protocol` definition instead of pointing to the expected base class.

ogenstad (approved):

dgarros:
> I think we might run into issues with this:
> 
> https://github.com/opsmill/infrahub-sdk-python/blob/develop/infrahub_sdk/schema.py#L111-L112
> 
> When we import the generators from the Git Agent and run this check on the imported objects we're probably going to get a mismatch. I still think we can move forward with this PR, but potentially we'll need to switch to using a `protocol` definition instead of pointing to the expected base class.

This is a good point, I'll check that

PR #96, opened 2024-10-17
Add specific timeouts per request per #25

### LINKED ISSUE #25: feature: Add support for specific timeout per request
### Component

Python SDK

Currently it's possible configure a default_timeout but at the client level but it's not possible to configure a specific timeout on most request. 

Would be good to allow the user to control the timeout directly when he/she is calling the following functions
- `Client.get`
- `Client.all`
- `Client.create`
- `Client.filters`
- `Node.save`

> `query_gql_query` and `execute_graphql` already support a specific timeout

### Describe the Use Case

Allow the user to have more control over the timeout

### Additional Information

_No response_

PhillSimonds:
This PR adds specific timeouts per request to the InfrahubClient and InfrahubNode objects per issue #25.

fixes #25

### DIFF
--- a/changelog/25.added.md
+++ b/changelog/25.added.md
@@ -1,0 +1,1 @@
+Add support for specific timeout per request on InfrahubClient and InfrahubNode function calls.
--- a/infrahub_sdk/client.py
+++ b/infrahub_sdk/client.py
@@ -324,6 +324,7 @@ class InfrahubClient(BaseClient):
         raise_when_missing: Literal[False],
         at: Optional[Timestamp] = ...,
         branch: Optional[str] = ...,
+        timeout: Optional[int] = ...,
         id: Optional[str] = ...,
         hfid: Optional[list[str]] = ...,
         include: Optional[list[str]] = ...,
@@ -341,6 +342,7 @@ class InfrahubClient(BaseClient):
         raise_when_missing: Literal[True],
         at: Optional[Timestamp] = ...,
         branch: Optional[str] = ...,
+        timeout: Optional[int] = ...,
         id: Optional[str] = ...,
         hfid: Optional[list[str]] = ...,
         include: Optional[list[str]] = ...,
@@ -358,6 +360,7 @@ class InfrahubClient(BaseClient):
         raise_when_missing: bool = ...,
         at: Optional[Timestamp] = ...,
         branch: Optional[str] = ...,
+        timeout: Optional[int] = ...,
         id: Optional[str] = ...,
         hfid: Optional[list[str]] = ...,
         include: Optional[list[str]] = ...,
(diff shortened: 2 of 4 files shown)


PhillSimonds:
I've added timeouts per what is outlined in #25.

The issue requests the addition of a timeout to the `InfrahubClient's` `create()` method. I added a timeout here as the method acquire a schema from the server in order to instantiate an `InfrahubNode` object in python memory.

This method creates an `InfrahubNode` object in python memory, but it does not make calls to the API to create the object on the server. From my understanding of the code, I would expect `.create()` on the client to call the API and create my Node -- but I may be misguided here. Can you confirm what's currently happening (no creation in the server) is the intended behavior of the `InfrahubClient's` `.create()` method?

dgarros:
Thanks @PhillSimonds, really appreciate the help with this PR

> This method creates an InfrahubNode object in python memory, but it does not make calls to the API to create the object on the server. From my understanding of the code, I would expect .create() on the client to call the API and create my Node -- but I may be misguided here. Can you confirm what's currently happening (no creation in the server) is the intended behavior of the InfrahubClient's .create() method?

Yes this is the expected behavior for `create` currently, 
Once you have the node "created" you can save it and then it will make the call to the API
```python
obj = self.client.create(kind="BuiltinTag", name="Blue")
obj.save()
```
we should probably change the name for something more explicit

One of the main benefit of this approach is that we can use a batch to execute all the API calls in parallel

```python
batch = await client.create_batch()
for tag in ["red", "green", "blue", "yellow", "orange"]:
     obj = client.create(kind="BuiltinTag", name=tag)
     batch.add(task=client.save)

async for obj, result in batch.execute():
    print(f"Tag {obj.name.value} created")
```

Having said that I totally see your point that someone might expect `client.create()` to actually create the object in infrahub, maybe we should add an option to control that (in a different PR). any thoughts from @opsmill/customer-success on this one ?

ogenstad:
Might be that the name is misleading and it should be `client.create` -> `client.stage` or similar.

wvandeun:
Personally I like the workflow the way it is, ie you create the node in memory and then save it. 
Which is aligned with how you update a node, you retrieve the node, make your changes and then save these changes.

That being said, I understand the point.
We could add an argument (save) to the create method, which defaults to `False`? But then you get a different workflow compared to updating a node. 
```python
device = client.create(InfraDevice, name="my device", save=True) 
```
But then we might as well add an argument that allows you to pass in a batch, to add the save action to a batch instead.
```python
batch = client.create_batch()
device = client.create(InfraDevice, name="my device", save_with_batch=batch) 
```

An alternative could be to rename the `create` method. Maybe `new_node` or `stage_node`?

PhillSimonds:
> Thanks @PhillSimonds, really appreciate the help with this PR...

Happy to help! Thanks for the clarifications :)

ogenstad (approved):
Thanks Phillip!

PR #107, opened 2024-10-25
Update protocols to match latest core schema

dgarros:
Related to https://github.com/opsmill/infrahub/pull/4716

### DIFF
--- a/infrahub_sdk/protocols.py
+++ b/infrahub_sdk/protocols.py
@@ -7,7 +7,8 @@ from typing import TYPE_CHECKING
 from .protocols_base import CoreNode, CoreNodeSync
 
 if TYPE_CHECKING:
-    from .node import RelatedNode, RelatedNodeSync, RelationshipManager, RelationshipManagerSync
+    from infrahub_sdk.node import RelatedNode, RelatedNodeSync, RelationshipManager, RelationshipManagerSync
+
     from .protocols_base import (
         URL,
         Boolean,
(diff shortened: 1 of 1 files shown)


infrahub_sdk/protocols.py
```diff
@@ -397,7 +399,6 @@ class CoreNumberPool(CoreResourcePool, LineageSource):
 
 
 class CoreObjectPermission(CoreBasePermission):
-    branch: String
```
dgarros, line 400: @gmazoyer is this change expected ? it wasn't part of my change

ogenstad: That should be fine as we decided to remove the "branch" parts as part of the object permissions with the argument that users probably wouldn't want to setup permissions on specific branches and instead have settings for "allow on default branch" or "allow on other branches" on in the future "allow on branches that I have created".

ogenstad (approved):

infrahub_sdk/protocols.py
```diff
@@ -7,7 +7,8 @@
 from .protocols_base import CoreNode, CoreNodeSync
 
 if TYPE_CHECKING:
-    from .node import RelatedNode, RelatedNodeSync, RelationshipManager, RelationshipManagerSync
+    from infrahub_sdk.node import RelatedNode, RelatedNodeSync, RelationshipManager, RelationshipManagerSync
```
ogenstad, line 10: Was there some problem with the relative import here?

dgarros: I haven't touched this part, not sure why the import was relative in the first place
this is what the current code is generating

PR #123, opened 2024-11-02
Move test from integration -> unit

ogenstad:
This test was added after the repo was split and shouldn't be in the integration folder as we're only mocking the data. Also given the config of the conftest fixtures that gets automatically loaded it wouldn't work as written as part of the integration tests when we re-enable them.

Also moved the utils into the helpers directory.

I've also marked four of the tests as ok to fail, this wasn't noticed earlier but looks to be related to Python 3.9. I still think it's better to have it in this state to ease the reactivation of integration tests where these tests previously didn't exist.


tests/unit/ctl/test_transform_app.py
```diff
@@ -64,6 +65,7 @@ def test_transform_not_exist_in_infrahub_yml(tags_transform_dir: str) -> None:
             assert output.exit_code == 1
 
     @staticmethod
+    @pytest.mark.xfail(reason="Seems problematic in Python 3.9")
```
dgarros: we could use skipif here to exclude these tests in 3.9 but keep them for the other version of Python
https://docs.pytest.org/en/stable/how-to/skipping.html#id1

dgarros (approved):

tests/unit/ctl/test_transform_app.py
ogenstad: Ah yes good call, updated to do this instead.

PR #126, opened 2024-11-06
Change ctl check to not use InfrahubCheck.init()

ogenstad:
This work relates to https://github.com/opsmill/infrahub/issues/3932 as a goal to have a consistent way of loading modules that behaves the same way. An issue that might come up is that of circular imports as the check.py is importing InfrahubClient.

We're removing the use of InfrahubCheck.init() from the CTL as it doesn't provide much value and also marking that method as deprecated.

### DIFF
--- a/infrahub_sdk/checks.py
+++ b/infrahub_sdk/checks.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import importlib
 import os
+import warnings
 from abc import abstractmethod
 from typing import TYPE_CHECKING, Any, Optional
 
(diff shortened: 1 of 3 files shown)


infrahub_sdk/checks.py
```diff
@@ -81,6 +82,11 @@ def client(self, value: InfrahubClient) -> None:
     @classmethod
     async def init(cls, client: Optional[InfrahubClient] = None, *args: Any, **kwargs: Any) -> InfrahubCheck:
         """Async init method, If an existing InfrahubClient client hasn't been provided, one will be created automatically."""
+        warnings.warn(
+            "InfrahubCheck.init has been deprecated and will be removed in the version in Infrahub SDK 1.1.0",
```
dgarros: This code will ship for the first time with 1.1 so this can't be accurate
To be clean let's say 2.0

dgarros (approved):

infrahub_sdk/checks.py
ogenstad: I'd like to get rid of the `InfrahubClient()` call below as it requires us to specifically import InfrahubClient. Are you fine with saying "after 1.1.0" instead? Alternatively we can change it to 2.0 and use importlib or similar here until that point. I don't think anyone outside of infrahubctl or the git-agent ever used this.

ogenstad: I've pushed an update to schedule this method to be removed in SDK 2.0 instead and at the same type we change how InfrahubClient is used within that module in order to change the InfrahubCheck module to follow the same logic as used for Transforms and generators in #128.

PR #128, opened 2024-11-06
Fix relative module imports

### LINKED ISSUE #81: bug: importing functions from other modules in repository doens't work consistently
### Component

Python SDK, infrahubctl

### Infrahub SDK version

0.14.0

### Current Behavior

When you want to import functions (or classes, variables) from other python modules into python transformations or generators then there is inconsistency on how this can be achieved.

| resource  | infrahubctl (relative) | infrahubctl (absolute) | Infrahub (relative) | Infrahub (absolute) |
|:----------|:-----------------------|:-----------------------|:--------------------|:--------------------|
| transform | ❌                     | ❌                     | ✅                  | ❌                  |
| generator |   ❌                  | ✅                      | ✅                  | ❌                  |

### Expected Behavior

Ideally we can import functions / classes variables from modules using relative imports when doing this in tranformations and generators, running locally or within Infrahub.

This way common code in transformations and generators can be shared in modules that are located anywhere within the repository.

### Steps to Reproduce

- load an instance of Infrahub with the demo schema
- create a BuiltinTag node 
- create an external repository
- define the following` .infrahub.yml` file
```yaml
---
queries:
  - name: test
    file_path: test.gql

python_transforms:
  - name: configuration
    file_path: transformers/configuration.py
    class_name: Configuration
```
- create a directory transformers in the root of the repository
- create a file `transformers/configurations.py` with the following contents
```python
from typing import Any, Dict
from infrahub_sdk.transforms import InfrahubTransform

from .utils import get_value

class Configuration(InfrahubTransform):
    query: str = "test"

    async def transform(self, data: Dict[str, Any]) -> Dict[str, Any]:
        return {
            "utils.get_value": get_value(),
        }
```
- create a file `transformers/utils.py` with the following contents
```python
def get_value() -> str:
    return "value"
```
- create a file `test.gql` at the root of the repository with the following contents
```graphql
query test($name: String!) {
  BuiltinTag(name__value: $name) {
    edges {
      node {
	name {
	  value
	}
      }
    }
  }
}
```
- stage all the newly created files and commit them
- try running the transformation using `infrahubctl transform configuration` > issue importing the get_value function
- push the changes to a git remote
- import the repository into Infrahub
- run the transformation through the REST API > works

### Additional Information

_No response_

ogenstad:
Fixes #81

* Removed InfrahubTransform.init() method that was marked for deprecation (to avoid circular import for InfrahubClient)
* Change the import_module function to support reloads if changes are noticed as some tests highlighted problems with the current implementation (I think we want to spend some more time on this soon and clean it up some more such as having the `sys.path` part within a context manager so that we remove the path we just added after importing the module)
* Change transform to be imported using the .load_class method tied to the repository object instead of using `get_transform_class_instance()` for consistency between modules (also related to #129)
* Refactor how we find the generator_definition or transform within each function, the linter was complaining about too many local variables so I modified this to instead use `.get_generator_definition()` and `.get_python_transform()`, due to this I also changed the exception that's raised when we fail to find the requested object from `KeyError` to a new `ResourceNotDefinedError`.
* Changed test output related to the refactoring

We should do the same changes for InfrahubCheck but we'll have to get rid of `InfrahubCheck.init()` first.

### DIFF
--- a/infrahub_sdk/_importer.py
+++ b/infrahub_sdk/_importer.py
@@ -2,19 +2,25 @@ from __future__ import annotations
 
 import importlib
 import sys
+from pathlib import Path
 from typing import TYPE_CHECKING, Optional
 
 from .exceptions import ModuleImportError
 
 if TYPE_CHECKING:
-    from pathlib import Path
     from types import ModuleType
+
+module_mtime_cache: dict[str, float] = {}
 
 
 def import_module(
     module_path: Path, import_root: Optional[str] = None, relative_path: Optional[str] = None
 ) -> ModuleType:
     import_root = import_root or str(module_path.parent)
+
+    file_on_disk = module_path
+    if import_root and relative_path:
+        file_on_disk = Path(import_root, relative_path, module_path.name)
 
     if import_root not in sys.path:
         sys.path.append(import_root)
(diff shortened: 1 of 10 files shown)


dgarros (approved):

infrahub_sdk/_importer.py
```diff
@@ -2,20 +2,26 @@
...
+
 
 def import_module(
     module_path: Path, import_root: Optional[str] = None, relative_path: Optional[str] = None
 ) -> ModuleType:
     import_root = import_root or str(module_path.parent)
 
+    file_on_disk = module_path
+    if import_root and relative_path:
+        file_on_disk = Path(import_root) / Path(relative_path) / module_path.name
```
gmazoyer: Maybe something like this `Path(import_root, relative_path, module_path.name)`?

ogenstad: Ah, yes good catch. Fixed!

PR #134, opened 2024-11-11
sets branch on InfrahubClient when running `infrahubctl transform` `infahubctl check` and `infrahubctl generator`

### LINKED ISSUE #133: bug: `infrahubctl transform` instantiates InfrahubClient without setting the branch
### Component

infrahubctl

### Infrahub SDK version

1.0.0

### Current Behavior

When you run `infrahubctl transform --branch test`, then the InfrahubClient instantiated by the tranform is not setup for that branch.

### Expected Behavior

The InfrahubClient object instantiated by the transform, should be setup with the correct branch.

### Steps to Reproduce

- set up instance of Infrahub
- load the demo schema and data
- set up the infrahub-demo-edge repository
- create a branch `test`
- create a device in that branch
- insert a breakpoint on line 8 in `openconfig.py` in the demo-edge-repository
- run the following command
`infrahubctl transform OCInterfaces device=test --branch test`
- in the debugger print the default branch of the client, it will be set to main
`p self.client.default_branch`

### Additional Information

_No response_

wvandeun:
fixes #133 

additionally fixes the same issue for `infrahubctl check` and `infrahubctl generator` commands.

### DIFF
--- a/changelog/133.fixed.md
+++ b/changelog/133.fixed.md
@@ -1,0 +1,1 @@
+Fixes issues where InfrahubClient was not properly configured for a branch when running the `infrahubctl transform` command.
--- a/changelog/133.fixed.md
+++ b/changelog/133.fixed.md
@@ -1,1 +1,1 @@
-Fixes issues where InfrahubClient was not properly configured for a branch when running the `infrahubctl transform` command.
+Fixes issues where InfrahubClient was not properly configured for a branch when running the `infrahubctl transform`, `infrahubctl check` and `infrhubctl generator` commands.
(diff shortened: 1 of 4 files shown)


dgarros (approved):
Looks good, would be good to add a newsfragment to include it in the release note

ogenstad (approved):
We're sending the branch when initializing the transport class a bit further down.

I'd imagine that some parts work now but that it might load the schema from the wrong branch etc without this fix. Looks like we might have the same issue for Checks and Generators just base on how the client is initialized.

```python
        transform = get_transform_class_instance(
            transform_config=transform_config,
            branch=branch,
            client=client,
        )
```

ogenstad (approved):

PR #142, opened 2024-11-13
Add wait_until_completion to InfrahubBranchManager.create

LucasG0:
Related PR: https://github.com/opsmill/infrahub/pull/4917

### DIFF
--- a/infrahub_sdk/branch.py
+++ b/infrahub_sdk/branch.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Optional, Union
+import warnings
+from typing import TYPE_CHECKING, Any, Literal, Optional, Union, overload
 from urllib.parse import urlencode
 
 from pydantic import BaseModel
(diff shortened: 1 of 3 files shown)


infrahub_sdk/branch.py
```diff
@@ -90,7 +101,8 @@ async def create(
 
         query = Mutation(mutation="BranchCreate", input_data=input_data, query=MUTATION_QUERY_DATA)
         response = await self.client.execute_graphql(query=query.render(), tracker="mutation-branch-create")
-
+        if not wait_until_completion:
+            return BranchData(**response["BranchCreate"]["task"])
```
LucasG0: - We cannot return branch object as it has not be created yet, does returning `task_id` make sense? Downside is that `create` return signature becomes `BranchData | str`
- Currently, `background_execution` has no effect but is available sdk side, so this is somehow a small breaking change even though I do not expect of this parameter to be used

infrahub_sdk/branch.py
```diff
@@ -77,10 +78,19 @@ async def create(
         branch_name: str,
         sync_with_git: bool = True,
         description: str = "",
-        background_execution: bool = False,
-    ) -> BranchData:
+        wait_until_completion: bool = False,
```
LucasG0: Should we add this parameter for other branch methods (`merge`, ...)?

infrahub_sdk/branch.py
ogenstad: > * We cannot return branch object as it has not be created yet, does returning `task_id` make sense? Downside is that `create` return signature becomes `BranchData | str`

That's unfortunate. But for now I think we should go with `BranchData | str` I don't remember if that works on Python 3.9 or if we have to do a `Union[BranchData, str]`. What we can do is to use an @overload decorator which we use in a few other places. We can use that to indicate the returned data type depending on the input params used.

infrahub_sdk/branch.py
```diff
@@ -91,6 +121,10 @@ async def create(
         query = Mutation(mutation="BranchCreate", input_data=input_data, query=MUTATION_QUERY_DATA)
         response = await self.client.execute_graphql(query=query.render(), tracker="mutation-branch-create")
 
+        # Make sure server version is recent enough to support background execution, as previously
```
dgarros, line 125: not sure if this is so useful because if the server doesn't support `wait_until_completion` it will blow up when you execute the query 2 lines up

LucasG0: As discussed, let's use deprecated `background_execution` as query input field so query does not break, and we will switch to `wait_until_completion` the day `background_execution` is completely removed from the graphql query.

dgarros (approved):

PR #143, opened 2024-11-15
Remove task_report.py

LucasG0:
Related to IFC-917


ogenstad:
We'd have to create a pipeline for the Infrahub project where we updated the submodule commit to use this branch and verify that it works within Infrahub itself first. I think we're still using this task_report object in a number of places there.

ogenstad:
There's a merge conflict on this PR, but I think we should be able completely remove the task_report from here now?

ogenstad (approved):

PR #149, opened 2024-11-22
Fix GraphQL query sent with hierarchical nodes

LucasG0:


ogenstad:
This looks good to me, it might be a bit weird that we start to use fragments without actually requesting fragments.

I.e. even if someone did this:

```python
from infrahub_sdk import InfrahubClientSync

client = InfrahubClientSync()
nodes = client.all(kind="LocationSite", prefetch_relationships=True, populate_store=True, fragment=False)
```

We'd still use fragments which might seem confusing. Having said given how these hierarchial models work I think it's a good compromise and most users wouldn't care about the fact that these objects require fragments. Thoughts @gmazoyer?

It would be good to have a pipeline complete with this change in the main infrahub repo. Can you open a PR there that updates the submodule to this commit?

gmazoyer:
> We'd still use fragments which might seem confusing. Having said given how these hierarchial models work I think it's a good compromise and most users wouldn't care about the fact that these objects require fragments. Thoughts @gmazoyer?

Actually the idea of using a fragment came up during our quick sync with. We thoughts that going down that path would simplify the code (instead of going through several if clauses) and probably would not hurt the user in the process. So I agreed with Lucas' choice, and thought it was a pretty good idea.

tests/integration/test_infrahub_client.py
```diff
@@ -293,3 +293,28 @@ async def test_create_branch(self, client: InfrahubClient, db: InfrahubDatabase,
     async def test_create_branch_async(self, client: InfrahubClient, db: InfrahubDatabase, init_db_base, base_dataset):
         task_id = await client.branch.create(branch_name="new-branch-2", wait_until_completion=False)
         assert isinstance(task_id, str)
+
+    # See issue #148.
```
gmazoyer, line 297: I asked myself this once: should we just put the entire link to the issue as a comment to a test that solves it? I think it makes sense to have it/

LucasG0: It depends on our preferencies, actually, I think I prefer full link too. It's a bit more verbose but it allows quicker access to corresponding issue, and it is even more relevant as we have now multiple repositories (at least sdk/server), and it should be easy to find/replace these links in case repositories names change.

LucasG0:
Main infrahub repo related PR: https://github.com/opsmill/infrahub/pull/5023

ogenstad (approved):

gmazoyer (approved):

PR #160, opened 2024-12-01
Cleanup and restructure Pydantic models to manage the Schema

dgarros:
This PR is a first step toward cleaning up and restructuring the Pydantic models to manage the schema in the SDK

The main changes is the separation of models for what a user can provide and what we are getting from the SDK
Currently most fields are shared but overtime the goal is to generate the models automatically to have the proper type defined (Optional or not).

In the process, I've split the file schema.py into multiple files to separate the models we are using for the repository config from the models for the main schema.

### DIFF
--- a/infrahub_sdk/checks.py
+++ b/infrahub_sdk/checks.py
@@ -17,7 +17,7 @@ if TYPE_CHECKING:
     from pathlib import Path
 
     from . import InfrahubClient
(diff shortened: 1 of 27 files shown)


ogenstad (approved):
I think this looks great, it's something that I've been thinking about for some time will be a nice cleanup.

PR #176, opened 2024-12-19
bugfix: Process nodes in generators CTL when no variables are passed in

FragmentedPacket:
Add processing of nodes when no variables are passed into CLI so the nodes get added into the store.

This added `generator.process_nodes` call that the `generator.collect_data` method typically handles, but since we're collecting data outside, `.process_nodes` was never called within `infrahubctl`.

### DIFF
--- a/infrahub_sdk/ctl/generator.py
+++ b/infrahub_sdk/ctl/generator.py
@@ -64,7 +64,6 @@ async def run(
             infrahub_node=InfrahubNode,
         )
         await generator._init_client.schema.all(branch=generator.branch_name)
-        await generator.process_nodes(data=data)
         await generator.run(identifier=generator_config.name, data=data)
 
     else:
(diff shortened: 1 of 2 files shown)


infrahub_sdk/ctl/generator.py
```diff
@@ -101,6 +101,7 @@ async def run(
                 repository_config=repository_config,
             )
             await generator._init_client.schema.all(branch=generator.branch_name)
+            await generator.process_nodes(data=data)
```
LucasG0: So it seems `generator.process_nodes` needs to be ran prior running `genetator.run`.
Would it be more consistent to call `generator.process_nodes` directly within `generator.run` when data is filled?
It would avoid to forget to use `generator.process_nodes` if/when we reuse `generator.run`. Considering this fix, this call to `process_nodes` should be removed too as it would be then already called within `run`: https://github.com/opsmill/infrahub-sdk-python/pull/176/files#diff-1a54e970abf259fa925f5e73f4d7d7984bdec3afa50b7e752cda368d859f2691R67

LucasG0 (approved):

infrahub_sdk/ctl/generator.py
FragmentedPacket: Yes this is a way better solution and keeps the interface the same regardless of CTL or SDK.

Thanks for the feedback!

PR #183, opened 2024-12-24
Add testing directory based on infrahub_testcontainers

dgarros:
This PR moves some tests utilities previously part of `infrahub.testing` into the SDK under `infrahub_sdk.testing`.

`infrahub_sdk.testing` is designed to work with the package `infrahub_testcontainers` that makes it possible to run integration tests for all projects using the Python SDK 
As part of this PR, I fixed 2 integration tests as part of `tests/integration/test_node.py`, there is still a lot of work to fix all of them.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,565 +1,563 @@
-import asyncio
+# import asyncio
 import os
-from typing import Any, Optional
 
-import httpx
-import pytest
-import ujson
-from fastapi.testclient import TestClient
-from infrahub import config
-from infrahub.components import ComponentType
-from infrahub.core.initialization import first_time_initialization, initialization
-from infrahub.core.node import Node
-from infrahub.core.utils import delete_all_nodes
-from infrahub.database import InfrahubDatabase, get_db
-from infrahub.lock import initialize_lock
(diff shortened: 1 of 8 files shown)


tests/integration/conftest.py
```diff
@@ -1,565 +1,563 @@
...
+#         self, message: InfrahubMessage, routing_key: str, delay: Optional[MessageTTL] = None, is_retry: bool = False
+#     ) -> None:
+#         self.messages.append(message)
+#         if routing_key not in self.messages_per_routing_key:
+#             self.messages_per_routing_key[routing_key] = []
+#         self.messages_per_routing_key[routing_key].append(message)
+
+#     @property
+#     def seen_routing_keys(self) -> list[str]:
+#         return list(self.messages_per_routing_key.keys())
```
ajtmccarty, line 563: are you hoping to uncomment this file in a later PR or should it be deleted?

ajtmccarty (approved):
are these supposed to run as part of the CI eventually?

tests/integration/conftest.py
dgarros: we need to finish the conversion of all the integration tests, probably best to keep it for now

PR #185, opened 2024-12-27
Fix pytest plugin relative import

### LINKED ISSUE #166: bug: pytest plugin does not support the usage of relative imports
### Component

Python SDK

### Infrahub SDK version

1.1.0

### Current Behavior

You are unable to use the pytest plugin for user provided resources that use a relative import.

```bash
tests/test_resources.yml::infrahub_check__check_tag__smoke-test FAILED                                                 [100%]

========================================================== FAILURES ==========================================================
______________________________________ resource: infrahub_check__check_tag__smoke-test _______________________________________
attempted relative import with no known parent package
================================================== short test summary info ===================================================
FAILED tests/test_resources.yml::infrahub_check__check_tag__smoke-test
===================================================== 1 failed in 0.03s ======================================================

```

### Expected Behavior

The pytest plugin works for user provided resources that use relative imports

### Steps to Reproduce

- spin up an instance of Infrahub
- create a local repository and initialize it as a repository
- create the file `check_tag.py` in the directory
```python
from infrahub_sdk.checks import InfrahubCheck
from .utils import something

class MyCheck(InfrahubCheck):
    query = "check_tag"

    async def validate(self, data: dict) -> None:
        a = something()
        return
```
- create the file `utils.py` in the directory
```python
def something() -> str:
    return "something"
```
- create the file `check_tag.gql` in the directory
```graphql
query check_tag($name: String!) {
  BuiltinTag(name__value: $name) {
    edges {
      node {
	id
      }
    }
  }
}
```
- create the file `.infrahub.yml` in the directory
```yaml
---
check_definitions:
  - name: check_tag
    class_name: MyCheck
    file_path: "check_tag.py"
    targets: tags
    parameters:
      name: "name__value"
```
- create a tests directory in the repository
- create a file `tests/test_resources.yml`
```yaml
---
version: "1.0"
infrahub_tests:
  - resource: Check
    resource_name: check_tag
    tests:
      - name: smoke-test
        expect: PASS
        spec:
          kind: check-smoke
```
- in the root of the respository execute the following command
```bash
pytest -v
```

### Additional Information

_No response_

LucasG0:
Fixes #166 

[Current PR](https://github.com/opsmill/infrahub-sdk-python/pull/171) for this fix breaks checks running through proposed changes pipeline with `FileNotFoundError /source/checks/...`. This is due to the fact that, while running a proposed changed pipeline, pytest is not invoked from imported local git repository, leading to a further wrong import. I tried to keep using `import_module` as suggested in original PR, but even playing with `sys.path` I did not end up to make it work. Finally, setting `__package__` accordingly on git repository modules did the job, see https://github.com/opsmill/infrahub-sdk-python/pull/185/files#diff-f79b55143786fc44e214de36e3ccdad8a4841ca3491ad9eaf467a4a6924afb06R26.

Other changes are minor refactoring.

### DIFF
--- a/changelog/166.fixed.md
+++ b/changelog/166.fixed.md
@@ -1,0 +1,1 @@
+Fix relative imports for the pytest plugin, note that the relative imports can't be at the top level of the repository alongside .infrahub.yml. They have to be located within a subfolder.
--- a/infrahub_sdk/_importer.py
+++ b/infrahub_sdk/_importer.py
@@ -16,6 +16,15 @@ module_mtime_cache: dict[str, float] = {}
 def import_module(
     module_path: Path, import_root: Optional[str] = None, relative_path: Optional[str] = None
 ) -> ModuleType:
+    """Imports a python module.
+
+    Attributes:
+        module_path (Path): Absolute path of the module to import.
+        import_root (Optional[str]): Absolute string path to the folder.
+        relative_path (Optional[str]): Relative string path between module_path and import_root.
+                                       TODO Compute `relative_path` here instead of having it as a parameter?
+    """
+
     import_root = import_root or str(module_path.parent)
 
     file_on_disk = module_path
@@ -35,6 +44,8 @@ def import_module(
         module_name = relative_path.replace("/", ".") + f".{module_name}"
 
     try:
+        # We hold a mapping of imported modules. If a module is already loaded and does not have recent changes,
+        # then we do not reload/import this module.
         if module_name in sys.modules:
             module = sys.modules[module_name]
             current_mtime = file_on_disk.stat().st_mtime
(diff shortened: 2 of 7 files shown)


dgarros (approved):

ogenstad (changes requested):
When doing validation tests with this branch I'm not able to get it to work. I think we need some additional changes here before we can actually get it to work as desired.

The check I'm validating with looks like this:

```python
from infrahub_sdk.checks import InfrahubCheck
from .utils import something

class MyCheck(InfrahubCheck):
    query = "check_tag"

    async def validate(self, data: dict) -> None:
        a = something()
        return
```

If I remove the `from .utils import something` part it works.

ogenstad:
This got implemented in with some additional changes in #171 instead.

PR #186, opened 2024-12-30
Prep for 1.3 and add release note

dgarros:


CHANGELOG.md
```diff
@@ -11,6 +11,50 @@ This project uses [*towncrier*](https://towncrier.readthedocs.io/) and the chang
...
+    @pytest.fixture(scope="class")
+    async def test_create_tag(self, default_branch: str, client: InfrahubClient) -> None:
+        obj = await client.create(kind="BuiltinTag", name="Blue")
+        await obj.save()
+        assert obj.id
+```
+
+### Changed
+
+- The Pydantic models for the schema have been split into multiple versions to align better with the different phase of the lifecycle of the schema.
```
ogenstad, line 48: Should we highlight this a bit more to indicate that if people were importing the repository classes from `infrahub_sdk.schema` they would need to change those imports? While I don't see why end users would have to import these classes you never know.

dgarros: I'm planning to send a quick note in Discord as well ... as you mentioned I don't think a lot of users are using these models today
also, the import should still work ... the main difference is that some attributes have been removed from the models

PR #196, opened 2024-12-31
adding integration tests

wartraxx51:


.github/workflows/ci.yml
```diff
@@ -171,3 +171,45 @@ jobs:
           codecov --flags python-filler-${{ matrix.python-version }}
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+
+  # ------------------------------------------ Integration Tests  ------------------------------------------
+  integration-tests:
+    strategy:
+      matrix:
+        python-version:
+          - "3.9"
```
dgarros: not sure that we need to run the integration tests for all versions of Python ...
I think the latest version should be enough

Instead it would be good to have a matrix with the version of Infrahub but it's not a blocker for this PR, we can revisit that later

pyproject.toml
dgarros, line 1: Please can you rebase your branch with the latest `develop`, these changes shouldn't be here, this is surprising
the version of the package is already `"1.3.0"` in both `develop` and `stable`

.github/workflows/ci.yml
wartraxx51: removed

pyproject.toml
wartraxx51: i rebased

dgarros (approved):
thank you

PR #205, opened 2025-01-02
Change behavior of include to only include the requested fields

### LINKED ISSUE #192: task: Change behavior of `include` to only include the requested fields
### Component

Python SDK

### Task Description

Currently it's possible to specify `include` and `exclude` when querying a node or a group of nodes from Infrahub.

Both methods are currently working on top of the default set of fields, meaning that
- `include` can be used to add fields that are not included by default
- `exclude` can be used to remove fields that are included by default

The issue with the current behavior is that the performance of a given query change over time if the schema changes and if more fields are being queried by default. there is no way to control that. this can be especially impacting if the queyr is also using `prefetch_relationships`

The proposal is to change the behavior of `include` to only include the fields that have been specified in `include` this way a query using `include` will be predictable and won't be impacted by future schema changes

dgarros:
Fixes #192

### DIFF
--- a/changelog/192.changed.md
+++ b/changelog/192.changed.md
@@ -1,0 +1,2 @@
+When using include as part of a query, the default behavior has changed.
+Only the fields defined in `include` will be returned now, where previously the Python SDK would add the fields provided to the default one.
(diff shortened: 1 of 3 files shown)


ajtmccarty (approved):

ogenstad (approved):

BaptisteGi:
@ogenstad @dgarros @ajtmccarty is it good to be merged?

dgarros:
> @ogenstad @dgarros @ajtmccarty is it good to be merged?

Not yet, I need to add some tests

PR #208, opened 2025-01-03
Adds `infrahubctl info` command

### LINKED ISSUE #109: feature: infrahubctl info command
### Component

infrahubctl CLI

Add a new command to `infrahubctl` to quickly display the main information about the system it's connected to.
- Server Address
- Check if the authentication is working
- Connected user and eventually level of permissions
- Version running on the server
- Version of the SDK

If `infrahubctl` is currently not configured (missing address or missing token) it should display the appropriate instructions on how to configure these parameters to the user.

Regarding the name of the command, I would suggest `infrahubctl info` but I'm open to other suggestions

### Describe the Use Case

When using infrahubctl I often want to understand quickly what is the current state of the system and which instance of infrahub I'm connected to

As an example I can have 2 instances of infrahub running, one for production and one for development and before loading a new schema I want to make sure I'm connected to the right instance

### Additional Information

An equivalent function should be implemented in the SDK, I'm sure it will be useful for other tools

minitriga:
Fixes: #109 
Screenshot 2025-01-03 at 16 43 55

Detail View
Screenshot 2025-01-03 at 16 44 15

Error state
Screenshot 2025-01-03 at 16 44 41

@dgarros shall I add recommendations based on `AuthenticationError` and `ServerNotReachableError`

Something like: `Please check api key or username/password` and `Ensure you have the correct infrahub address defined`?

### DIFF
--- a/infrahub_sdk/ctl/cli_commands.py
+++ b/infrahub_sdk/ctl/cli_commands.py
@@ -4,6 +4,7 @@ import asyncio
 import functools
 import importlib
 import logging
+import platform
 import sys
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable
@@ -12,7 +13,11 @@ import jinja2
 import typer
 import ujson
 from rich.console import Console
+from rich.layout import Layout
 from rich.logging import RichHandler
+from rich.panel import Panel
+from rich.pretty import Pretty
+from rich.table import Table
 from rich.traceback import Traceback
 
 from .. import __version__ as sdk_version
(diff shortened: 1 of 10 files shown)


dgarros:
Thanks @minitriga 
I think we could remove the groups from the summary view and add the username instead.

minitriga:
I will also add some unit tests to cover `infrahubctl version/info` commands.

infrahub_sdk/ctl/cli_commands.py
```diff
@@ -392,11 +398,176 @@ def protocols(
...
+
+    info: dict[str, Any] = {
+        "error": None,
+        "status": ":x:",
+        "infrahub_version": "N/A",
+        "user_info": {},
+        "groups": {},
+    }
+    try:
+        client = initialize_client_sync()
```
dgarros, line 419: Would be good to move this code into specific functions that other users could reuse outside of the CLI
```python
client.get_version()
client.get_user_permissions()
```

or similar

dgarros (approved):

PR #212, opened 2025-01-05
Fix regression with Node update

dgarros:
This PR fixes an issue introduced by #199 that would prevent a node with relationship of cardinality one from being updated.

_As a workaround, it's recommended to enable the property when querying the object (property=True)_

This is good reminder that : 
1. We need more integration tests
2. We need to refactor how the SDK manages partial update today (`_strip_unmodified`)

_I've also included the RN and the version upgrade to 1.4.1 in this PR_

### DIFF
--- a/tests/integration/test_node.py
+++ b/tests/integration/test_node.py
@@ -143,9 +143,11 @@ class TestInfrahubNode(TestInfrahubDockerClient, SchemaCarPerson):
         initial_schema: None,
         manufacturer_mercedes,
         person_joe,
+        person_jane,
         car_golf,
         tag_blue,
         tag_red,
+        tag_green,
     ):
         car_golf.color.value = "White"
         await car_golf.tags.fetch()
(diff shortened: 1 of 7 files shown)


tests/integration/test_node.py
```diff
@@ -143,48 +143,31 @@ async def test_node_update(
...
-
-    #     nodedb = await NodeManager.get_one(id=node.id, db=db, include_owner=True, include_source=True)
-    #     repodb = await nodedb.repository.get_peer(db=db)
-    #     assert repodb.id == repo99.id
-
-    #     tags = await nodedb.tags.get(db=db)
-    #     assert sorted([tag.peer_id for tag in tags]) == sorted([tag_green.id, tag_blue.id])
+        car3 = await client.get(kind=TESTING_CAR, id=car_golf.id)
+        await car3.tags.fetch()
+        assert sorted([tag.id for tag in car3.tags.peers]) == sorted([tag_green.id, tag_blue.id])
```
ajtmccarty, line 170: you can use `set`s for these instead of sorting

tests/unit/sdk/test_node.py
```diff
@@ -176,15 +182,19 @@ async def test_init_node_data_user_with_relationships(client, location_schema: N
     assert node.primary_tag.id == "pppppppp"
 
 
-@pytest.mark.parametrize("client_type", client_types)
-async def test_init_node_data_graphql(client, location_schema: NodeSchemaAPI, location_data01, client_type):
+@pytest.mark.parametrize("client_type, property_test", list(product(client_types, property_tests)))
+async def test_init_node_data_graphql(
```
ajtmccarty, line 186: you can also use `@pytest.mark.parametrize` multiple times and pytest will do the `product(client_types, property_tests)` part for you

tests/unit/sdk/test_node.py
```diff
@@ -1561,22 +1642,84 @@ async def test_update_input_data_with_relationships_02(client, location_schema,
...
+    }
+
+    expected_result = (
+        expected_result_without_property if property_test == WITHOUT_PROPERTY else expected_result_with_property
+    )
+
+    assert location._generate_input_data(exclude_unmodified=True)["data"] == expected_result
+
+
+# @pytest.mark.xfail(reason="FIXME: this function shouldn't return name and type since they are not being modified")
```
ajtmccarty: can be deleted?

ajtmccarty (approved):
2 "you can do it if you want to" comments and one commented-out line that can maybe be removed

tests/unit/sdk/test_node.py
dgarros: Oh nice, I was looking for it but I couldn't find it

tests/integration/test_node.py
dgarros: Indeed, not sure how i missed that, I'll fix in another PR

tests/unit/sdk/test_node.py
dgarros: I moved it in a comment inside the test ... there are a few things that need to be investigated further but I didn't want to include too much in this PR

PR #218, opened 2025-01-06
Resolves #130: Replace GitPython with dulwich

### LINKED ISSUE #130: feature: Replace GitPython with dulwich
### Component

Python SDK, infrahubctl CLI

Today, some features related to Git in the python SDK is dependent on having the `git` binary available on the system but the installation is left to the user. 

The proposal is to replace `GitPython` with the library `dulwich`.
`dulwich` is built in pure python and doesn't require the `git` binary to be available.

### Describe the Use Case

Improve the user experience

### Additional Information

Related to #16

minitriga:
Fixes: #130


pyproject.toml
```diff
@@ -51,6 +50,7 @@ typer = { version = "^0.12.3", optional = true }
 pytest = { version = "*", optional = true }
 pyyaml = { version = "^6", optional = true }
 eval-type-backport = { version = "^0.2.2", python = "~3.9" }
+dulwich = "^0.22.7"
```
dgarros: Do you know if there is a minimum version required for the feature that we are using ?
Best to keep the dependencies as loose as possible for the SDK

infrahub_sdk/repository.py
```diff
@@ -0,0 +1,73 @@
...
+        """Get the name of the current active branch."""
+        if not self.git:
+            raise ValueError("Repository is not initialized.")
+
+        # Read the symbolic reference of HEAD to get the current commit SHA
+        head_ref = self.git.refs[b"HEAD"]
+        if head_ref:
+            commit_sha = head_ref.decode("utf-8")  # Commit SHA from HEAD
+            # Now look for the branch that points to this commit SHA
+            for ref in self.git.refs.as_dict().keys():
```
dgarros: Not sure If I understand this code correctly /// 
how is it gonna work if someone creates a new branch that has the same commit as the main branch ? can it match on the main branch instead of the current branch ?

infrahub_sdk/repository.py
```diff
@@ -0,0 +1,73 @@
...
+                if ref.startswith(b"refs/heads/"):
+                    branch_ref = ref.decode("utf-8")
+                    # Check if this branch points to the same commit SHA
+                    if self.git.refs[ref] == commit_sha.encode("utf-8"):
+                        return branch_ref.split("/")[-1]
+            return None  # If no matching branch is found
+        raise ValueError("No HEAD reference found, cannot determine active branch.")
```
dgarros: we could bring it up and exit early if there is an issue, it would make the code a bit easier to read
```python
if not head_ref:
    raise ValueError("No HEAD reference found, cannot determine active branch.")
```

dgarros (commented):
Great work, thank you for taking care of this one

ogenstad:
Perhaps a bit unrelated to this PR but I'm wondering if we should care about git within the Checks and Generators or if we should just use the settings from the SDK config if the `default_branch_from_git` setting was used. Thoughts?

infrahub_sdk/repository.py
minitriga: I tried to match the current functionality of GitPython which returning `active_branch` returned the current `HEAD` that was active. 

https://gitpython.readthedocs.io/en/stable/reference.html#git.repo.base.Repo.active_branch

But yes I see the issue creating a new branch will cause the refs to have two matches in which It will match only the first. Ill double check this tomorrow.

dgarros (approved):

Issue #158, opened 2024-11-29
feature: add count method to InfrahubClient


wvandeun:
### Component

Python SDK

### Describe the Feature Request

Add a count method that allows you to count the kind of objects that match your query:

```python
from infrahub_sdk import InfrahubClient
client = InfrahubClient()
num_devices = await client.count(kind="InfraDevice", site__name__value="alt1")
```

### Describe the Use Case

In scenarios where I would need to count the amount of objects that match a query, I have 2 options today:
- retrieve the objects and take the length of the resulting list, which could be expensive, because we would automatically fetch attributes and relationships of cardinality one
- construct a raw GraphQL query

The above approaches are not ideal and we could make this more simple and less expensive.

wvandeun:
Today, we can achieve the same using a GraphQL query:
```python
from infrahub_sdk import InfrahubClientSync
client = InfrahubClientSync()
query = "query { BuiltinTag { count }}"
resp = client.execute_graphql(query=query)
print(resp["BuiltinTag"]["count"])
```

PR #219, opened 2025-01-06
resolves #159: Add batch argument to fetch data in batches

### LINKED ISSUE #159: feature: provide a way to use pagination concurrently to retrieve objects using the SDK
### Component

Python SDK

When you execute a query that retrieves a large amount of nodes from the db, using the filters or all method, then the SDK will leverage pagination to break the query into smaller pages.

The retrieval of this pages happens in a serial way, which is ideal. We can do this better and faster if we retrieve the pages concurrently.

Pseud code of what it could look like

```python

client = InfrahubClient()
resp = await client.get("query { LocationSuite { count }}")

count = int(resp["LocationSuite"]["count"])

batch = await client.create_batch()

num = 1
page_size=50
while num < count:
    batch.add(task=client.all, kind="nodekind", offset=num, limit=page_size)
    num += page_size
```

### Describe the Use Case

Retrieval of a large number of nodes using a GraphQL query is taking some time, since we are retrieving the pages one by one in a serial way. Being able to get the pages in a concurrent way should improve the speed.

### Additional Information

_No response_

minitriga:
Resolves: #159


infrahub_sdk/client.py
```diff
@@ -727,14 +729,49 @@ async def filters(
...
-
-            page_number += 1
+            return response, process_result
+
+        async def process_batch(schema_kind: str) -> tuple[list[InfrahubNode], list[InfrahubNode]]:
+            """Process queries in batch mode."""
+            nodes = []
+            related_nodes = []
+            batch_process = await self.create_batch()
+            resp = await self.execute_graphql(query=f"query {{ {schema_kind} {{ count }} }}")
```
dgarros: Would be good to move this into a dedicated `count` method, there's already an issue to track that https://github.com/opsmill/infrahub-sdk-python/issues/158

infrahub_sdk/client.py
```diff
@@ -623,6 +628,7 @@ async def filters(
         prefetch_relationships: bool = ...,
         partial_match: bool = ...,
         property: bool = ...,
+        batch: bool = ...,
```
dgarros: Thinking out loud
I'm not sure if `batch` will be clear enough for the end user ? 
As an alternative we could use `parallel`
any preference @opsmill/customer-success ?

dgarros (approved):
Looks good overall

infrahub_sdk/client.py
dgarros: Looks like @gmazoyer is already on it https://github.com/opsmill/infrahub-sdk-python/pull/225

minitriga: I have now moved it over to using `client.count()`.

infrahub_sdk/client.py
minitriga: I have renamed to `parallel` for now.

dgarros (approved):

PR #220, opened 2025-01-07
Upgrade ruff=0.8.6

ogenstad:


dgarros (approved):
Looks good but I think it would be best to update it in `stable` otherwise we'll need to manage 2 different versions until 1.2 is out

ogenstad:
Ah ok, closing and opening #221 instead.

PR #223, opened 2025-01-07
Order of retrieved nodes can be disabled

LucasG0:


infrahub_sdk/types.py
```diff
@@ -64,3 +66,7 @@ def exception(self, event: str | None = None, *args: Any, **kw: Any) -> Any:
 
 
 InfrahubLoggers = Union[InfrahubLogger, Logger]
+
+
+class OrderModel(BaseModel):
```
dgarros: I think we should call it just `Order` to keep it simpler for the user and easier to read

dgarros (approved):
Looks good

dgarros:
Please can you add an entry in the changelog directory to include something in the release note

PR #232, opened 2025-01-10
Add possibility to use filters in the SDK client's count method

wvandeun:


gmazoyer (approved):

tests/unit/sdk/test_client.py
```diff
@@ -83,6 +83,16 @@ async def test_method_count(clients, mock_query_repository_count, client_type):
     assert count == 5
 
 
+@pytest.mark.parametrize("client_type", client_types)
```
dgarros, line 86: Not sure if this test is really checking anything since the payload is static
ideally we should add an integration tests for it

dgarros (approved):

dgarros (approved):

PR #243, opened 2025-01-17
align environment variables between the SDK and `infrahubctl`

wvandeun:
`infrahuctl` was using other environment variables to achieve the same goal as other envrionment variables that we leverage in the SDK. This PR aligns the usage of them.

INFRAHUBCTL_TIMEOUT > INFRAHUB_TIMEOUT
INFRAHUBCTL_CONCURRENT_EXECUTION > INFRAHUBCTL_MAX_CONCURRENT_EXECUTION

### DIFF
--- a/infrahub_sdk/ctl/exporter.py
+++ b/infrahub_sdk/ctl/exporter.py
@@ -26,9 +26,9 @@ def dump(
     concurrent: int = typer.Option(
         4,
         help="Maximum number of requests to execute at the same time.",
-        envvar="INFRAHUBCTL_CONCURRENT_EXECUTION",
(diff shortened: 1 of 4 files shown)


infrahub_sdk/ctl/exporter.py
```diff
@@ -26,9 +26,9 @@ def dump(
     concurrent: int = typer.Option(
         4,
```
FragmentedPacket, line 27: Do we want to set the default to `None` and it'll take the SDK defaults? (There might be other changes required for this, but mainly the idea of these options should be overrides when defined, but match SDK defaults if not provided.

dgarros (approved):

PR #244, opened 2025-01-18
InfrahubCTL: Repository command changes

FragmentedPacket:
These changes allow us to set the ref which is used to define what release/branch/commit to uae

- Do not create/set blank credentials if no username or password are provided
- Change `commit` to `ref` option and update code to reflect the same (Breaking Change)

Is `commit` supposed to be an available field to update or one that gets set on synchronization?


infrahub_sdk/ctl/repository.py
```diff
@@ -70,7 +70,7 @@ async def add(
     description: str = "",
     username: str | None = None,
     password: str = "",
-    commit: str = "",
+    ref: str = "",
```
FragmentedPacket, line 74: Do we need to keep commit but make it mutually exclusive with ref?

BaptisteGi: So `commit` is indeed automatically populated by the sync. It's an internal field that user shouldn't change.

BaptisteGi: I think still there is a subtle difference here between *read only repository* and regular *core repository*. Ref only applies to read only, for core repository it's called `default_branch`.

FragmentedPacket: Thank you! I think removing the `commit` option makes sense and I added logic to change the `input_data` based on whether it's read-only or not based on your feedback as well to ensure we're sending the right payload.

BaptisteGi (approved):
There is a conflict to resolve otherwise sounds good!

PR #246, opened 2025-01-22
Run CI tests against multiple versions

LucasG0:
This PR enables CI tests to run against both latest infrahub server version and local version of stable/develop branch. Version is set using `INFRAHUB_TESTING_IMAGE_VER` environment variable (that we already use within python_testcontainers, and related logic should be simplified when [this PR](https://github.com/opsmill/infrahub/pull/5550) is cleaned/finalized).

Ideally, we would also like to have an `---infrahub-version` cli argument for users convenience. Adding such a parameter requires to define `pytest_adoption` within a `conftest.py` file. However, having this `conftest.py` being loaded when a user runs a test requires either of the tests to be located in the same folder/subfolders of `conftest.py` - and it should not be the case. So, we need a more elaborated pytest plugin/hook mechanism that would deserve a dedicated PR.

Note that we may later want to run sdk tests at least against pinned major infrahub versions.

### DIFF
--- a/infrahub_sdk/testing/docker.py
+++ b/infrahub_sdk/testing/docker.py
@@ -1,10 +1,40 @@
+from __future__ import annotations
+
+import os
+
 import pytest
 from infrahub_testcontainers.helpers import TestInfrahubDocker
+from packaging.version import InvalidVersion, Version
 
 from .. import Config, InfrahubClient, InfrahubClientSync
 
+INFRAHUB_VERSION = os.getenv("INFRAHUB_TESTING_IMAGE_VER", "latest")
+
+
+def skip_version(min_infrahub_version: str | None = None, max_infrahub_version: str | None = None) -> bool:
+    """
+    Check if a test should be skipped depending on infrahub version.
+    """
+    try:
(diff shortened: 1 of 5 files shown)


infrahub_sdk/testing/docker.py
```diff
@@ -1,10 +1,40 @@
...
 import pytest
 from infrahub_testcontainers.helpers import TestInfrahubDocker
+from packaging.version import InvalidVersion, Version
 
 from .. import Config, InfrahubClient, InfrahubClientSync
 
+INFRAHUB_VERSION = os.getenv("INFRAHUB_TESTING_IMAGE_VER", "latest")
+
+
+def skip_version(min_infrahub_version: str | None = None, max_infrahub_version: str | None = None) -> bool:
```
ogenstad, line 14: When is it that this skip_version function gets used?

dgarros (approved):

infrahub_sdk/testing/docker.py
LucasG0: Whenever a SDK test should be running prior/since a given infrahub version. We do not have such tests for now, but should have some in the future

Issue #251, opened 2025-01-28
bug: TypeError when executing CTL


FragmentedPacket:
### Component

infrahubctl

### Infrahub SDK version

1.4.0-1.7.0

### Current Behavior

Whenever attempting to use InfrahubCTL commands with Python 3.9, we encounter Typing issues on execution such as executing `infrahubctl version` encounters the following error:

```python
❯ infrahubctl version
╭──────────────────────────────────────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ /Users/Mikhail/cloned/infrahub-sdk-python/.venv/bin/infrahubctl:6 in <module>                                                                                                                                                                      │
│                                                                                                                                                                                                                                                    │
│   3 from infrahub_sdk.ctl.cli import app                                                                                                                                                                                                           │
│   4                                                                                                                                                                                                                                                │
│   5 if __name__ == '__main__':                                                                                                                                                                                                                     │
│ ❱ 6 │   sys.exit(app())                                                                                                                                                                                                                            │
│   7                                                                                                                                                                                                                                                │
│                                                                                                                                                                                                                                                    │
│ /Users/Mikhail/cloned/infrahub-sdk-python/.venv/lib/python3.9/site-packages/typer/main.py:340 in __call__                                                                                                                                          │
│                                                                                                                                                                                                                                                    │
│ /Users/Mikhail/cloned/infrahub-sdk-python/.venv/lib/python3.9/site-packages/typer/main.py:323 in __call__                                                                                                                                          │
│                                                                                                                                                                                                                                                    │
│ /Users/Mikhail/cloned/infrahub-sdk-python/.venv/lib/python3.9/site-packages/typer/main.py:362 in get_command                                                                                                                                       │
│                                                                                                                                                                                                                                                    │
│ /Users/Mikhail/cloned/infrahub-sdk-python/.venv/lib/python3.9/site-packages/typer/main.py:344 in get_group                                                                                                                                         │
│                                                                                                                                                                                                                                                    │
│ /Users/Mikhail/cloned/infrahub-sdk-python/.venv/lib/python3.9/site-packages/typer/main.py:481 in get_group_from_info                                                                                                                               │
│                                                                                                                                                                                                                                                    │
[...]

PR #256, opened 2025-01-30
Replace pendulum with whenever

### LINKED ISSUE #255: task: Replace pendulum with whenever
### Component

Python SDK

### Task Description

The project is currently using pendulum to parse dates but the library appears to not be supported anymore and it's blocking the support of Python 3.13

After some investigation, the best alternative appears to be https://github.com/ariebovenberg/whenever

dgarros:
fixes #255 

This PR replaces `pendulum` with `whenever` to manage datetime, mainly as part of the Timestamp class.

This change allows us to re-enable the support for Python 3.13, since we haven't shipped a release without 3.13 support, I've updated the changelog for #251 

In the process the Timestamp object has been slightly refactored
- Direct access to obj has been deprecated, instead it's recommended to use `to_datetime()`
- `add_delta()` has been deprecated, in favor of `add()` and `substract()` that are more inline with `whenever`

### DIFF
--- a/changelog/255.fixed.md
+++ b/changelog/255.fixed.md
@@ -1,0 +1,1 @@
+Refactor Timestamp to use `whenever` instead of `pendulum` and extend Timestamp with add(), subtract(), and to_datetime().
--- a/infrahub_sdk/timestamp.py
+++ b/infrahub_sdk/timestamp.py
@@ -1,9 +1,15 @@
 from __future__ import annotations
 
 import re
+import warnings
+from datetime import datetime, timezone
+from typing import Literal
 
-import pendulum
-from pendulum.datetime import DateTime
+from whenever import Date, Instant, LocalDateTime, Time, ZonedDateTime
+
+from .exceptions import TimestampFormatError
+
+UTC = timezone.utc  # Required for older versions of Python
 
 REGEX_MAPPING = {
     "seconds": r"(\d+)(s|sec|second|seconds)",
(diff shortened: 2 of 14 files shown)


changelog/255.fixed.md
```diff
@@ -0,0 +1 @@
+Refactor Timestamp to use `whenever` instead of `pendulum` and extend Timestamp with add(), subtract(), and to_datetime(). Direct access to `obj` and `add_delta` have been deprecated and will be removed in a future version.
```
ogenstad: I think we could split this up into two fragments where we also use the special one for things marked for deprecation.

infrahub_sdk/timestamp.py
```diff
@@ -16,76 +20,171 @@ class TimestampFormatError(ValueError): ...
...
+            self._obj = ZonedDateTime.now("UTC").round(unit="microsecond")
+
+    @property
+    def obj(self) -> ZonedDateTime:
+        warnings.warn(
+            "Direct access to obj property is deprecated. Use to_string(), to_timestamp(), or to_datetime() instead.",
+            UserWarning,
+            stacklevel=2,
+        )
+        return self._obj
```
ogenstad, line 41: I'd suggest creating a follow up issue for when this should be removed and tie it into a future milestone so we know when to remove it.

infrahub_sdk/timestamp.py
```diff
@@ -16,76 +20,171 @@ class TimestampFormatError(ValueError): ...
...
-                params[key] = int(match.group(1))
+                params[key] = float(match.group(1))
 
-        if not params:
-            raise TimestampFormatError(f"Invalid time format for {value}")
+        if params:
+            return ZonedDateTime.now("UTC").subtract(**params)  # type: ignore[call-overload]
 
-        return DateTime.now(tz="UTC").subtract(**params)
+        raise TimestampFormatError(f"Invalid time format for {value}")
```
ogenstad, line 79: I know it was like this before but I think the error definition should live in exceptions.py like the other core SDK exceptions.

ogenstad (approved):
We should also have a news fragment stating that the installation on Python 3.13 was fixed.

infrahub_sdk/utils.py
```diff
@@ -367,3 +369,29 @@ def get_user_permissions(data: list[dict]) -> dict:
...
+
+
+def calculate_time_diff(value: str) -> str | None:
+    """Calculate the time in human format between a timedate in string format and now."""
+    try:
+        time_value = Timestamp(value)
+    except TimestampFormatError:
+        return None
+
+    delta: TimeDelta = Timestamp()._obj.difference(time_value._obj)
```
ajtmccarty: shouldn't these use `to_datetime()`?

ajtmccarty (approved):

infrahub_sdk/utils.py
dgarros: Not `datetime`, the goal is to access the `ZonedDateTime` here
I added a new `get_obj()` instead of accessing it directly

infrahub_sdk/timestamp.py
dgarros: good point, I moved it

changelog/255.fixed.md
dgarros: Good point, I forgot about the deprecated section

infrahub_sdk/timestamp.py
dgarros: will do

PR #263, opened 2025-02-06
Protocols command consumes default branch from env

### LINKED ISSUE #104: bug: Infrahubctl protocols command requires explicit branch
### Component

infrahubctl

### Infrahub SDK version

0.14.1

### Current Behavior

```shell
> infrahubctl protocols
HTTP communication failure: Client error '400 Bad Request' for url 'http://localhost:8000/api/schema?branch=None'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400 on GET to http://localhost:8000/api/schema?branch=None
```

### Expected Behavior

I would expect the command to get the default branch from my environment.

The same way I could do `infrahubctl schema load schema.yml` without having to specify the branch

### Steps to Reproduce

Doesn't work:

```shell
infrahubctl protocols
```

Works:

```shell
infrahubctl protocols --branch main
```

### Additional Information

_No response_

BaptisteGi:
- `protocols` command properly gets default branch setting from environment variable 

resolves: #104

### DIFF
--- a/infrahub_sdk/ctl/cli_commands.py
+++ b/infrahub_sdk/ctl/cli_commands.py
@@ -383,6 +383,7 @@ def protocols(
 
     else:
         client = initialize_client_sync()
+        branch = branch or client.default_branch
         schema.update(client.schema.fetch(branch=branch))
 
     code_generator = CodeGenerator(schema=schema)
--- a/infrahub_sdk/ctl/cli_commands.py
+++ b/infrahub_sdk/ctl/cli_commands.py
@@ -129,7 +129,7 @@ async def run(
     method: str = "run",
     debug: bool = False,
     _: str = CONFIG_PARAM,
-    branch: str = typer.Option("main", help="Branch on which to run the script."),
+    branch: str = typer.Option("main", help="Branch on which to run the script."),  # TODO: Replace main by None
     concurrent: int | None = typer.Option(
         None,
         help="Maximum number of requests to execute at the same time.",
--- a/changelog/104.fixed.md
+++ b/changelog/104.fixed.md
@@ -1,0 +1,1 @@
+- `protocols` CTL command properly gets default branch setting from environment variable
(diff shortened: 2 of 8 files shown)


infrahub_sdk/ctl/cli_commands.py
```diff
@@ -129,7 +129,7 @@ async def run(
     method: str = "run",
     debug: bool = False,
     _: str = CONFIG_PARAM,
-    branch: str = typer.Option("main", help="Branch on which to run the script."),
+    branch: str = typer.Option("main", help="Branch on which to run the script."),  # TODO: Replace main by None
```
FragmentedPacket, line 132: I'm not understanding the change here, should it be setting the default to `None`?

ogenstad: In my opinion throughout `infrahubctl` the branch parameter should be an optional string defaulting to `None` and if not specified we should either not specify the branch or use the default branch provided by the SDK which will default to "main", alternatively something else if the SDK is configured that way with some environment variable.

BaptisteGi: Yep with the current setup you can't use environment variable `INFRAHUB_DEFAULT_BRANCH`.

```console
 ➜  export INFRAHUB_DEFAULT_BRANCH="test1"
 ➜  infrahubctl schema load some.yml   
 The schema in Infrahub was is already up to date, no changes were required
 1 schema processed in 0.138 seconds.
```

=> This loads schema to main branch

BaptisteGi: Here is the issue to fix all remaining commands https://github.com/opsmill/infrahub-sdk-python/issues/264 I guess it could be breaking someone's workflow ...

ogenstad: It might, although hopefully it wouldn't be that impactful as any changes would be done against a non main branch. But worth including in a changelog to indicate this updated behaviour. Not sure if we need all those todo comments when we have #264, but they might be easier to find I guess.

ogenstad (approved):

FragmentedPacket (approved):

PR #273, opened 2025-02-14
Add required change for object templates

gmazoyer:
Depends on https://github.com/opsmill/infrahub/pull/5610


dgarros (commented):
I think we need to create a new branch (`infrahub-next` or similar) for all changes that requires the next version of Infrahub, otherwise it will block us from publishing a new release of the SDK until the next version of Infrahub has been released

dgarros (approved):

PR #274, opened 2025-02-15
Add `infrahub.tasks` as built-in logger to `InfrahubGenerator` class

FragmentedPacket:
There have been a few requests to implement logging within generators within the CI pipeline output.

This adds an option to pass in your own logger or default to `infrahub.tasks`.

I also added `init_logging` in the generator CTL command so we can take advantage of the logging on the console as well.

Please let me know if there is a more optimal way, but this might be a good thing to include in a few of the classes that run within the CI pipeline.


dgarros (approved):

dgarros (commented):
LGTM, please can you add a news fragment to include something in the release note

PR #275, opened 2025-02-19
Move the function `read_file` from the ctl module to the SDK.

dgarros:
Small fix to avoid having `infrahub_sdk/yaml.py` import a function from the `ctl` directory.


ogenstad (approved):
LGTM, though I think we have a `FileNotValidError` exception class that we can delete from infrahub_sdk/ctl/exceptions.py as well as a cleanup.

ajtmccarty (approved):

PR #276, opened 2025-02-21
Add support for Tasks

### LINKED ISSUE #190: feature: Add support for Tasks
### Component

Python SDK

We need to improve the support for Tasks in the SDK to make it easy to :
- Query the list of tasks based on various filters (branch, related node, status, flow, tags etc)
- Query the list of logs associated with a given task
- Wait for a given task to complete
- Wait for all tasks that can impact the database to complete on a given branch (aka until the branch has converged)

In Infrahub 1.1, the usage of tasks has been improved and now there are a few actions that can take a long time (Branch merge,  branch rebase etc ..) that can be done asynchronously and that will return a task id. Ideally we should always execute these actions asynchronously in the SDK and let the user decide if the function should return immediately or not.

### Describe the Use Case

Improve the support for tasks in the Python SDK

### Additional Information

_No response_

dgarros:
Fixes #190


infrahub_sdk/task/models.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class Task(BaseModel):
+    id: str
+    title: str
+    state: TaskState
+    progress: float | None = None
+    workflow: str | None = None
+    branch: str | None = None
+    # start_time: datetime # Is it still required
```
gmazoyer, line 39: So it is not required, it seems.

gmazoyer (approved):

ogenstad (approved):

PR #283, opened 2025-02-24
ctl: Transform to not dump to JSON string if the return of the transform is text.

FragmentedPacket:
If the transform returns a string, do not dump to a JSON string.


dgarros (approved):
LGTM please can you add a newsfragment to include something in the release note

ogenstad (approved):

PR #287, opened 2025-02-27
fix: build for python 3.13

fatih-acar:
#256  was supposed to reintroduce 3.13 support, but according to CI logs:

image

Edit: need to investigate why `poetry install` is 20x slower with 3.13...


dgarros (approved):

ogenstad (approved):
LGTM, but based on that screenshot it seems like we also need to end with a test that validates that the running Python version is the one that we expect. :/ Without that I don't see how we could trust these tests.

PR #288, opened 2025-02-28
Raise URLNotFoundError instead of infinite retry

### LINKED ISSUE #5800: bug: git-repository-pull-read-only tasks stays in running state after you delete the branch in which you started the pull
### Component

Git Integration

### Infrahub version

1.1.7

### Current Behavior

When you update the ref in a read-only repository in a branch, and then immediately delete that branch, the git-repositor-pull-read-only task will have started and be in a running state. 

The task will stay in a running state though and not show any error message.

When you then update the ref of the read-only repository in main, a new additional git-repository-pull-read-only task will be created and enter in running state, but it doesn't seem to be doing anything.

The prefect flow doesn't show any logs, nor can you find any related log messages for the task-workers.

You can cancel the flow in prefect, but it will stay in cancelling state.

The only way to really get out of this state, seems to be restarting the server, task workers and redis containers. 

### Expected Behavior

The git-repository-pull-read-only task should fail with an error message, after which new pull tasks for this repository can be executed.

### Steps to Reproduce

- start instance of Infrahub
- add this repository as a read-only repository
- https://github.com/wvandeun/infrahub-transform
- create a branch test
- in the branch update the ref of the repository to `6f8576ac23894fcdf53ffeba4527c8402b32b7b8`
- immediately delete the test branch
- navigate to the repository in the main branch
- you can see the `git-repository-pull-read-only-task` in running state
- update the ref of the repository to `6f8576ac23894fcdf53ffeba4527c8402b32b7b8` in main
- a new task `git-repository-pull-read-only-task` will have been created and be in running state

### Additional Information

_No response_

LucasG0:
Fixes [#5800](https://github.com/opsmill/infrahub/issues/5800).

- Raise a URLNotFoundError on 404 errors, typically while trying to perform a graphql query on an unexisting branch.
- Adds an arbitrary 300 seconds maximum duration to avoid infinite loop.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -181,6 +181,6 @@ def generate_infrahubctl(context: Context) -> None:
 
 
 @task(name="generate-sdk")
-def generate_python_sdk(context: Context) -> None:
+def generate_python_sdk(context: Context) -> None:  # noqa: ARG001
     """Generate documentation for the Python SDK."""
-    _generate_infrahub_sdk_configuration_documentation(context=context)
+    _generate_infrahub_sdk_configuration_documentation()
--- a/docs/docs/python-sdk/reference/config.mdx
+++ b/docs/docs/python-sdk/reference/config.mdx
@@ -150,6 +150,15 @@ The following settings can be defined in the `Config` class
 **Default value**: False<br />
 **Environment variable**: `INFRAHUB_RETRY_ON_FAILURE`<br />
 <!-- vale off -->
+## max_retry_duration
+
+**Property**: max_retry_duration<br />
+<!-- vale on -->
+**Description**: Maximum duration until we stop attempting to retry if enabled.<br />
+**Type**: `integer`<br />
+**Default value**: 300<br />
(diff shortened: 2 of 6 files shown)


ajtmccarty (approved):

tasks.py
```diff
@@ -181,6 +181,6 @@ def generate_infrahubctl(context: Context) -> None:
 
 
 @task(name="generate-sdk")
-def generate_python_sdk(context: Context) -> None:
+def generate_python_sdk(context: Context) -> None:  # noqa: ARG001
```
dgarros, line 184: you can remove context if we don't use it

dgarros (commented):
so today we're retry indefinitely ?

LucasG0:
> so today we're retry indefinitely ?

If we delete a branch while a task in running on this branch and triggers this path, typically `git-repository-pull-read-only`, yes

tasks.py
LucasG0: It seems to be required by `invoke` tasks

dgarros (approved):

PR #292, opened 2025-03-03
Allow useof HFID to create a related node on generic relationship

LucasG0:


gmazoyer (approved):

infrahub_sdk/node.py
```diff
@@ -255,6 +256,9 @@ def _generate_input_data(self, allocate_from_pool: bool = False) -> dict[str, An
             data["id"] = self.id
         elif self.hfid is not None:
             data["hfid"] = self.hfid
+            if self._kind is not None:
+                # We might need `kind` for https://github.com/opsmill/infrahub/issues/4649
```
ajtmccarty: I don't think it is sustainable to add comments referencing GH issues
if we're worried about someone removing this line and breaking some piece of functionality, then we should have a unit test that covers it
for me, this seems like a simple enough change that a comment is not necessary at all

infrahub_sdk/testing/schemas/animal.py
```diff
@@ -80,6 +80,8 @@ def schema_cat(self) -> NodeSchema:
             namespace=NAMESPACE,
             include_in_menu=True,
             inherit_from=[TESTING_ANIMAL],
+            # Different hfid than Animal one, for testing https://github.com/opsmill/infrahub-sdk-python/issues/277
```
ajtmccarty: I'd say this comment should be in the test that uses it and the test can have an `assert cat_luna.hfid == ...` if you're concerned about someone changing this and invalidating the test
but I don't think we can add a comment every time we alter a fixture

LucasG0: Makes sense!

infrahub_sdk/node.py
LucasG0: Agree referencing issues should be either within dedicated test or for bigger changes, I'll remove this one

PR #296, opened 2025-03-07
Add support for request context

ogenstad:
Adds support for adding a request context to the SDK. This can be done at the root level of the client object or on individual requests. The idea here is that when running a generator we'd want to simplify this process for end users so that we add the context to the cloned SDK instance when running within the pipeline where as we wouldn't use this option when using `infrahubctl`.

Within the Infrahub pipeline we'd still need to be able to specify that specific objects should be saved using a context, such as when creating a validator or a check, then we are not able to set the context for the entire SDK instance.

There are still some calls missing (such as when creating branches etc) and tests would need to be written.

It would be used either like this:

```python
from infrahub_sdk.context import RequestContext

context = RequestContext(account={"id": "182a86a2-05f0-37f8-58df-1746c80a346e"})

client = InfrahubClient()
client.request_context = context

data = {
    "name": "blue-tag",
}
obj = await client.create(kind="BuiltinTag", data=data)
await obj.save()
```

Alternatively like this:

```python
from infrahub_sdk.context import RequestContext

context = RequestContext(account={"id": "182a86a2-05f0-37f8-58df-1746c80a346e"})

client = InfrahubClient()

data = {
    "name": "blue-tag",
}
obj = await client.create(kind="BuiltinTag", data=data)
await obj.save(request_context=context)
```

### DIFF
--- a/infrahub_sdk/client.py
+++ b/infrahub_sdk/client.py
@@ -58,6 +58,8 @@ from .utils import decode_json, get_user_permissions, is_valid_uuid
 if TYPE_CHECKING:
     from types import TracebackType
 
+    from .context import RequestContext
+
 
 SchemaType = TypeVar("SchemaType", bound=CoreNode)
 SchemaTypeSync = TypeVar("SchemaTypeSync", bound=CoreNodeSync)
@@ -139,6 +141,7 @@ class BaseClient:
         self.identifier = self.config.identifier
         self.group_context: InfrahubGroupContext | InfrahubGroupContextSync
         self._initialize()
+        self._request_context: RequestContext | None = None
 
     def _initialize(self) -> None:
         """Sets the properties for each version of the client"""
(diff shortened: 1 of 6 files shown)


dgarros (commented):
With this implementation, will it be possible to define the context globally when we initialize the client ? I think this is important to support this option in addition to specifying a specific context for a specific request

ogenstad:
> With this implementation, will it be possible to define the context globally when we initialize the client ? I think this is important to support this option in addition to specifying a specific context for a specific request

If you check the first example in the description that's very close to this. The scenario I was picturing was specifically for use with the generators, where we'd want to set the context on a cloned client after it was cloned. The use case I'd see to have it directly on the initialize would be when we support running jobs within an isolated Docker container or pod, but perhaps you have some other use case that I'm missing?

dgarros (approved):
> If you check the first example in the description that's very close to this. The scenario I was picturing was specifically for use with the generators, where we'd want to set the context on a cloned client after it was cloned. The use case I'd see to have it directly on the initialize would be when we support running jobs within an isolated Docker container or pod, but perhaps you have some other use case that I'm missing?

My bad I missed it, this is exactly what I had in mind too

ogenstad:
I've also updated the protocols that were falling behind, will revisit with an integration test once we have an image to test against. This will also be used in [#5985](https://github.com/opsmill/infrahub/pull/5985).

PR #299, opened 2025-03-09
Add integration tests for Task and add support for Enum

### LINKED ISSUE #18: feature: support enums in the SDK
### Component

Python SDK

Follows opsmill/infrahub#918 

The GraphQL API can now support enums based on our schema correctly. The Python SDK will need some changes to how it generates mutations for an attribute using an enum behind the scenes.

### Describe the Use Case

For an example, let's use the `CoreAccount.role` attribute, which uses this enum `["admin", "read-only", "read-write"]`.
The `CoreAccountCreate` mutation currently looks like this when creating an account with a particular role
```
mutation CreateAccount {
  CoreAccountCreate(
    data:{
      name: {value: "userperson"},
      password: {value: "infrahub"},
      role: {value: "read-only"}
    }
  ) {
    ok
    object {
      role { value }
    }
  }  
}
```

With correct enum support in the GraphQL API, `"read_only"` becomes `READ_ONLY`
```
mutation CreateAccount {
  CoreAccountCreate(
    data:{
      name: {value: "userperson"},
      password: {value: "infrahub"},
      role: {value: READ_ONLY}
    }
  ) {
    ok
    object {
      role { value }
    }
  }  
}
```

### Additional Information

you can turn enum support in the GraphQL API on or off using the environment variable `INFRAHUB_EXPERIMENTAL_GRAPHQL_ENUMS=true/false`

dgarros:
Fix #18 

This PR adds some integration tests for Task and fixes various issues related to task
In order to fully support task, this PR also adds the support for Enum to GraphQL query and mutation (#18)

By default, an enum will be rendered as an Enum but it's also possible to convert Enum to their respective value.

### DIFF
--- a/infrahub_sdk/graphql.py
+++ b/infrahub_sdk/graphql.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from enum import Enum
 from typing import Any
 
 from pydantic import BaseModel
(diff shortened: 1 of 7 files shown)


infrahub_sdk/graphql.py
```diff
@@ -1,25 +1,37 @@
...
+from enum import Enum
 from typing import Any
 
 from pydantic import BaseModel
 
 VARIABLE_TYPE_MAPPING = ((str, "String!"), (int, "Int!"), (float, "Float!"), (bool, "Boolean!"))
 
 
-def convert_to_graphql_as_string(value: str | bool | list) -> str:
+def convert_to_graphql_as_string(value: str | bool | list | BaseModel | Enum | Any, convert_enum: bool = False) -> str:  # noqa: PLR0911
```
ogenstad, line 11: In some other places for PLR0911 we've assigned the return value within multiple if-statements / match cases like this to avoid having a high number of returns.

ogenstad (approved):
I think we'll need to take a closer look at enums in general as there are different variations. We have our own Enums (which doesn't yet behave as GraphQL enums) within the schema that are different than GraphQL enums. I'm not sure this comes across anywhere. Than I think there might be more to be done with regards to the generation of Python protocols when it comes to enums.

PR #310, opened 2025-03-20
Add parameter to only optionally refresh the schema hash on schema.all

### LINKED ISSUE #152: feature: Add ability to view the schema hash of each branch within the SDK
### Component

Python SDK

Currently the SDK stores a cache of the schema for each branch that it has interacted with. For long running processes this can lead to an issue when the schema of a given branch is updated and you try to interact with new objects. For instance if you add attributes to a node and the SDK has the old schema hashed it could lead to an error when you try to access an attribute that doesn't exist in the cached version of the schema.

The current workaround for this would be to refresh the schema in the SDK:

```python
await client.schema.all(branch="my_branch", refresh=True)
```

A problem here is that it could be an expensive operation as it's typically not required. Each branch within Infrahub has a hash that identifies the schema configuration, it would be good if we could store this hash as we are loading the schema to the SDK cache. That way I'd be able to validate if the hash matches what I'd expect to see and if not I'd refresh the schema with the above command. If the cached branch has the same hash as the expected one there's no reason to refresh the cache.

### Describe the Use Case

Lower overhead and avoid refetching the schema so often.

### Additional Information

_No response_

ogenstad:
The goal of this PR is primarily for use within computed attributes, currently for the Jinja2 template based computed attributes we are constantly refreshing the schema of the SDK for each attribute. This is because we have no way of knowing if the schema in the client is up to date or if a branch we'd previously cashed has updates. With this in place we can send in the current hash stored in the registry to the schema.all method and only call the `/api/schema` endpoint if any updates have actually been made.

As a follow up to this PR I'm thinking that we should mark the public schema.fetch() method as deprecated. I don't think anyone uses it and instead use schema.get() or schema.all().

Fixes #152.

### DIFF
--- a/changelog/152.added.md
+++ b/changelog/152.added.md
@@ -1,0 +1,1 @@
+Add 'schema_hash' parameter to client.schema.all to only optionally refresh the schema if the provided hash differs from what the client has already cached.
--- a/infrahub_sdk/schema/__init__.py
+++ b/infrahub_sdk/schema/__init__.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import asyncio
-from collections import defaultdict
 from collections.abc import MutableMapping
 from enum import Enum
 from time import sleep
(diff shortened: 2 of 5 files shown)


dgarros (approved):
LGTM
Not sure about marking `fetch` as deprecated ... it's probably not used extensively but it could be useful.

PR #321, opened 2025-03-25
Restraint `store.get()` when a Kind is provided

BeArchiTek:
When we provided a Kind to the `store.get()` with `raise_when_missing=False` but there isn't any matching node, instead of returning `None` it will check all the nodes in the Store. So if there is another Node with the same key, it will return it.

This is causing an issue on `infrahub-sync`


dgarros (approved):

dgarros:
@BeArchiTek I addressed this issue in #323, please can you review it and let me know if that works for you

BeArchiTek:
https://github.com/opsmill/infrahub-sdk-python/pull/323 solved the issue as well, closing this PR

PR #323, opened 2025-03-26
Refactor internal store

dgarros:
This PR refactor the internal store to be easier to use and to cover more use cases
The main changes is the introduction of a new `internal_id` on all nodes to simplify how the nodes are organized within the store event before have been created in Infrahub.
Also with this change the store is now branch-aware and by default objects from 2 different branches will be isolated from each other. 
And last but not least, it's now easier to search node based on their HFID both with the flat version or the structured one.


infrahub_sdk/node.py
```diff
@@ -678,6 +690,11 @@ def __init__(self, schema: MainSchemaTypesAPI, branch: str, data: dict | None =
         self._branch = branch
         self._existing: bool = True
 
+        # Generate a unique ID only to be used inside the SDK
+        # The format if this ID is purposely different from the ID used by the API
+        # This is done to avoid confusion and potential conflicts between the IDs
+        self._internal_id = shortuuid.uuid()
```
ogenstad: Instead of installing an extra package just for this can we add a function such as:

```python
import uuid
import base64

def short_id() -> str:
    return base64.urlsafe_b64encode(uuid.uuid4().bytes).rstrip(b'=').decode('ascii')
```

infrahub_sdk/store.py
```diff
@@ -20,130 +23,336 @@ def get_schema_name(schema: str | type[SchemaType] | None = None) -> str | None:
...
-    """
+class NodeStoreBranch:
+    def __init__(self, name: str) -> None:
+        self.branch_name = name
 
-    def __init__(self) -> None:
-        self._store: dict[str, dict] = defaultdict(dict)
-        self._store_by_hfid: dict[str, Any] = defaultdict(dict)
+        self._objs: dict[str, InfrahubNode | InfrahubNodeSync | CoreNode | CoreNodeSync] = {}
+        self._hfids: dict[str, dict[tuple, str]] = defaultdict(dict)
```
ogenstad: Do we need this to be a defaultdict or can we just have a `{}` instead?

infrahub_sdk/store.py
```diff
@@ -20,130 +23,336 @@ def get_schema_name(schema: str | type[SchemaType] | None = None) -> str | None:
...
 
-        if hfid:
-            self._store_by_hfid[hfid] = node
+        if node.id:
+            self._uuids[node.id] = node._internal_id
+
+        if hfid := node.get_human_friendly_id():
+            self._hfids[node.get_kind()][tuple(hfid)] = node._internal_id
+
+    def get(  # type: ignore[no-untyped-def]
```
ogenstad: Looks like you've fixed the problem with `# type: ignore[no-untyped-def]` so you can remove that part.

infrahub_sdk/store.py
dgarros: we could do a dict but it means we'll need to add a check to ensure the sub level dict as been initialized
is there a concern using a `defaultdict` ?

infrahub_sdk/store.py
```diff
@@ -20,130 +23,336 @@ def get_schema_name(schema: str | type[SchemaType] | None = None) -> str | None:
...
+            pass
 
         if not raise_when_missing:
             return None
+
+        if kind and found_invalid:
+            raise NodeInvalidError(
+                node_type="n/a",
+                identifier={"key": [key]},
+                message=f"Found a node of a differentkind instead of {kind} for key {key!r} in the store ({self.branch_name})",
```
ogenstad: typo differentkind -> different kind

infrahub_sdk/store.py
```diff
@@ -20,130 +23,336 @@ def get_schema_name(schema: str | type[SchemaType] | None = None) -> str | None:
...
+        try:
+            return self._get_by_hfid(key, kind=kind_name)
+        except NodeNotFoundError:
+            pass
 
         if not raise_when_missing:
             return None
+
+        if kind and found_invalid:
+            raise NodeInvalidError(
```
ogenstad, line 99: This error is raised when running `infrahub_edge.py` would be good with a pipeline in the Infrahub repo for this PR.

BeArchiTek:
Tested as replacement of https://github.com/opsmill/infrahub-sdk-python/pull/321, all good for me

infrahub_sdk/store.py
ogenstad: To me they are just an even less strict data structure than a `dict`, I think the code is often clearer without them.

infrahub_sdk/store.py
```diff
@@ -20,130 +23,336 @@ def get_schema_name(schema: str | type[SchemaType] | None = None) -> str | None:
...
+        if not kind:
+            node_kind, node_hfid = parse_human_friendly_id(hfid)
+        elif kind and hfid.startswith(kind):
+            node_kind, node_hfid = parse_human_friendly_id(hfid)
+        else:
+            node_kind = kind
+            node_hfid = [hfid]
+
+        exception_to_raise_if_not_found = NodeNotFoundError(
+            node_type=node_kind or "unknown",
```
ogenstad: We're mixing how we use node_type in the NodeNotFoundErrors here it's "unknown" where as in other places it's "n/a". Should we perhaps fix this within the error itself so that node_type is optional?

infrahub_sdk/store.py
```diff
@@ -20,130 +23,336 @@ def get_schema_name(schema: str | type[SchemaType] | None = None) -> str | None:
...
+
+
+class NodeStoreBase:
+    """Internal Store for InfrahubNode objects.
+
+    Often while creating a lot of new objects,
+    we need to save them in order to reuse them later to associate them with another node for example.
+    """
+
+    def __init__(self, default_branch: str = "main") -> None:
```
ogenstad: I think it was like this before but it might be unexpected for users that have configured a different default branch when initializing the SDK and then just use these store objects where we default to "main".

ogenstad (commented):
Looks good overall to me, but there seems to be some issues that need to be resolved as highlighted by the infrastructure_edge script. Also I don't think that we should add that extra package.

infrahub_sdk/store.py
dgarros: I removed the default branch for now but it will introduce a "breaking change" so not 100% sure if we should go with it

ogenstad: No with that in mind I'd suggest to keep it as is and instead create an issue so that we cycle back and refactor this in a controlled manner.

ogenstad:
Based on the changes done in `infrastructure_edge.py` are we introducing any kind of breaking changes what would impact other scripts out there? Also does this need a news fragment or will that come in the follow up work?

dgarros:
> Based on the changes done in `infrastructure_edge.py` are we introducing any kind of breaking changes what would impact other scripts out there? Also does this need a news fragment or will that come in the follow up work?

I will add a fragment

If we add back the default branch I think it should work out of the box for existing users but I asked the CS team to do a quick check just to be sure

infrahub_sdk/store.py
dgarros: I added the default branch back along with a deprecation warning

ogenstad (approved):

PR #327, opened 2025-03-27
Refactor handling of Jinja2 templates and add filters

ogenstad:
Refactors the management of Jinja2 templates, some notes on this PR.

* The new Jinja2Template class is initialized by providing a `template` parameter as either a `str` or a `pathlib.Path` object, in the first case the template will be rendered from a string and the latter part will use file based templates
* The rendering is done in an async manner in order to support async Jinja2 filters
* The `.get_variables()` method will return a list of any variables that are required by the template, a current limitation for now is that for file based templates this is only supported for the first file and not if the template includes other files as imports (this part can be added later, I just want to have it in place for the string based computed attributes and it's not used anywhere today)
* I reimplemented the identify_faulty_jinja_code() function with some minor modifications
* In this PR I've added custom exceptions under the infrahub_sdk.template folder, I'm not completely sure we want to do this? They all have the base infrahub_sdk.exceptions.Error as an ancestor
* Some older exceptions in infrahub_sdk.exceptions and the identify_faulty_jinja_code() will need to be deprecated

Note that we talked about providing a prefix for the netutils filters, though I'm not sure this is required now that we've listed the source of each filter within the documentation. Any thoughts around that?


infrahub_sdk/template/__init__.py
```diff
@@ -0,0 +1,161 @@
...
+from .exceptions import (
+    JinjaTemplateError,
+    JinjaTemplateNotFoundError,
+    JinjaTemplateOperationViolationError,
+    JinjaTemplateSyntaxError,
+    JinjaTemplateUndefinedError,
+)
+from .models import UndefinedJinja2Error
+
+netutils_filters = jinja2_convenience_function()
```
dgarros, line 24: curious why we have this function at the root of the file, if possible I think it would make sense to move it inside Jinja2Template

infrahub_sdk/template/__init__.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+class Jinja2Template:
+    def __init__(self, template: str | Path, template_directory: Path | None = None) -> None:
```
dgarros: I think it would be good to have a way to provide additional filters when we initialized this class, it would also make it easier to test it

dgarros (approved):

infrahub_sdk/template/__init__.py
ogenstad: Just a slight performance thing since the function would check to see if the modules have been imported each time we render a template. But I can move it into the class instead, we'd have to have a pretty significant number of templates rendered for it to be an issue.

infrahub_sdk/template/__init__.py
dgarros: All good, I was just curious

infrahub_sdk/template/__init__.py
dgarros: all good I was just curious

infrahub_sdk/template/__init__.py
ogenstad: This has been added now.

docs/docs/python-sdk/reference/templating.mdx
```diff
@@ -0,0 +1,153 @@
+---
+title: Python SDK Templating
+---
+
+
+## Builtin Jinja2 filters
```
BaptisteGi, line 6: Maybe worth pointing to the doc on this one (https://jinja.palletsprojects.com/en/stable/templates/#list-of-builtin-filters) so users know what's behind each line

BaptisteGi (commented):
I think it would be great to point to that reference page from the main infrahub documentation, I'm thinking about (might be some other places):
- Jinja2 computed attributes
- Jinja2 transform (artifact)

docs/docs/python-sdk/reference/templating.mdx
ogenstad: I've added these links.

PR #328, opened 2025-03-30
Add `infrahubctl object` command

dgarros:
Add `infrahub object load` and `infrahub object validate` commands to load objects defined in Yaml format into Infrahub.

These commands are meant to be used in an idempotent way and as such they would work better for models with a Human Friendly ID (hfid) defined.

## File format

All object files must start with the following format, all other format will be automatically ignored
Each file is intended for one specific top level kind, but one file can include multiple nested objects of any kind.
The kind of the top level object must be defined in `spec/kind`

```yaml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: <NamespaceName>
  data:
    - [,,,]
```

Below is a more detailed example with nested objects and with related objects referenced by their HFIDs

```yaml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: TestingPerson
  data:
    - name: Alex Thompson
      height: 180
      animals:  # Relationship of cardinality many that will create/update nodes as needed
                      # The owner of the dog will be automatically inserted as long as their is a bidirectional relationship
                      # between TestingPerson and TestingDog
        kind: TestingDog         # The kind must be provided when the relationship is based on a generic
        data:
          - name: Max
            weight: 25
            breed: Golden Retriever
            color: "#FFD700"
      tags:           # The kind is optional here because there is only one option possible (not a generic)
        data:
          - name: Dog Lover    # New Tag will be created if it doesn't exist


    - name: Mike Johnson
      height: 175
      best_friends: # Relationship of cardinality many that referenced existing nodes based on their HFID
        - [Jane Smith, Max]
        - [Sarah Williams, Charlie]
      tags:
        - Veterinarian    # Existing Node referenced by its HFID in string format
        - [Breeder]       # Existing Node referenced by its HFID in list format

### DIFF
--- a/infrahub_sdk/ctl/object.py
+++ b/infrahub_sdk/ctl/object.py
@@ -7,9 +7,14 @@ from rich.console import Console
 from ..async_typer import AsyncTyper
 from ..ctl.client import initialize_client
 from ..ctl.utils import catch_exception, init_logging
+from ..exceptions import ObjectValidationError, ValidationError
 from ..spec.object import ObjectFile
 from .parameters import CONFIG_PARAM
-from .utils import load_yamlfile_from_disk_and_exit
+from .utils import (
+    display_object_validate_format_error,
+    display_object_validate_format_success,
+    load_yamlfile_from_disk_and_exit,
+)
 
 app = AsyncTyper()
 console = Console()
@@ -39,9 +44,54 @@ async def load(
     files = load_yamlfile_from_disk_and_exit(paths=paths, file_type=ObjectFile, console=console)
     client = initialize_client()
 
-    for file in files:
-        file.validate_content()
-        schema = await client.schema.get(kind=file.spec.kind, branch=branch)
+    has_errors = False
 
-        for item in file.spec.data:
-            await file.spec.create_node(client=client, schema=schema, data=item, branch=branch)
+    for file in files:
+        try:
+            await file.validate_format(client=client, branch=branch)
+        except ValidationError as exc:
+            has_errors = True
+            display_object_validate_format_error(file=file, error=exc, console=console)
+
+    if has_errors:
+        raise typer.Exit(1)
+
+    for file in files:
+        try:
+            await file.process(client=client, branch=branch)
+        except ObjectValidationError as exc:
+            has_errors = True
+            console.print(f"[red] {exc!s}")
+
+    if has_errors:
+        raise typer.Exit(1)
+
+
+@app.command()
+@catch_exception(console=console)
+async def validate(
+    paths: list[Path],
+    debug: bool = False,
+    branch: str = typer.Option(None, help="Branch on which to validate the objects."),
+    _: str = CONFIG_PARAM,
+) -> None:
+    """Validate one or multiple objects files."""
+
+    init_logging(debug=debug)
+
+    logging.getLogger("infrahub_sdk").setLevel(logging.INFO)
+
+    files = load_yamlfile_from_disk_and_exit(paths=paths, file_type=ObjectFile, console=console)
+    client = initialize_client()
+
+    has_errors = False
+    for file in files:
+        try:
+            await file.validate_format(client=client, branch=branch)
+            display_object_validate_format_success(file=file, console=console)
+        except ValidationError as exc:
+            has_errors = True
+            display_object_validate_format_error(file=file, error=exc, console=console)
+
+    if has_errors:
+        raise typer.Exit(1)
--- a/infrahub_sdk/spec/object.py
+++ b/infrahub_sdk/spec/object.py
@@ -1,19 +1,331 @@
 from __future__ import annotations
 
+from enum import Enum
 from typing import TYPE_CHECKING, Any
 
 from pydantic import BaseModel, Field
 
+from ..exceptions import ObjectValidationError, ValidationError
+from ..schema import RelationshipSchema
 from ..yaml import InfrahubFile, InfrahubFileKind
 
 if TYPE_CHECKING:
     from ..client import InfrahubClient
-    from ..schema import MainSchemaTypesAPI
+    from ..node import InfrahubNode
+    from ..schema import MainSchemaTypesAPI, RelationshipSchema
+
+
+def validate_list_of_scalars(value: list[Any]) -> bool:
+    return all(isinstance(item, (str, int, float, bool)) for item in value)
+
+
+def validate_list_of_hfids(value: list[Any]) -> bool:
+    return all(isinstance(item, (str, list)) for item in value)
+
+
+def validate_list_of_data_dicts(value: list[Any]) -> bool:
+    return all(isinstance(item, dict) and "data" in item for item in value)
+
+
+def validate_list_of_objects(value: list[Any]) -> bool:
+    return all(isinstance(item, dict) for item in value)
+
+
+class RelationshipDataFormat(str, Enum):
+    UNKNOWN = "unknown"
+
+    ONE_REF = "one_ref"
+    ONE_OBJ = "one_obj"
+
+    MANY_OBJ_DICT_LIST = "many_obj_dict_list"
+    MANY_OBJ_LIST_DICT = "many_obj_list_dict"
+    MANY_REF = "many_ref_list"
+
+
+class RelationshipInfo(BaseModel):
+    name: str
+    rel_schema: RelationshipSchema
+    peer_kind: str
+    peer_rel: RelationshipSchema | None = None
+    reason_relationship_not_valid: str | None = None
+    format: RelationshipDataFormat = RelationshipDataFormat.UNKNOWN
+
+    @property
+    def is_bidirectional(self) -> bool:
+        """Indicate if a relationship with the same identifier exists on the other side"""
+        return bool(self.peer_rel)
+
+    @property
+    def is_mandatory(self) -> bool:
+        if not self.peer_rel:
+            return False
+        return not self.peer_rel.optional
+
+    @property
+    def is_valid(self) -> bool:
+        return not self.reason_relationship_not_valid
+
+    @property
+    def is_reference(self) -> bool:
+        return self.format in [RelationshipDataFormat.ONE_REF, RelationshipDataFormat.MANY_REF]
+
+    def get_context(self, value: Any) -> dict:
+        """Return a dict to insert to the context if the relationship is mandatory"""
+        if self.peer_rel and self.is_mandatory and self.peer_rel.cardinality == "one":
+            return {self.peer_rel.name: value}
+        if self.peer_rel and self.is_mandatory and self.peer_rel.cardinality == "many":
+            return {self.peer_rel.name: [value]}
+        return {}
+
+    def find_matching_relationship(
+        self, peer_schema: MainSchemaTypesAPI, force: bool = False
+    ) -> RelationshipSchema | None:
+        """Find the matching relationship on the other side of the relationship"""
+        if self.peer_rel and not force:
+            return self.peer_rel
+
+        try:
+            self.peer_rel = peer_schema.get_matching_relationship(
+                id=self.rel_schema.identifier or "", direction=self.rel_schema.direction
+            )
+        except ValueError:
+            pass
+
+        return self.peer_rel
+
+
+async def get_relationship_info(
+    client: InfrahubClient, schema: MainSchemaTypesAPI, name: str, value: Any, branch: str | None = None
+) -> RelationshipInfo:
+    """
+    Get the relationship info for a given relationship name.
+    """
+    rel_schema = schema.get_relationship(name=name)
+
+    info = RelationshipInfo(name=name, peer_kind=rel_schema.peer, rel_schema=rel_schema)
+
+    if isinstance(value, dict) and "data" not in value:
+        info.reason_relationship_not_valid = f"Relationship {name} must be a dict with 'data'"
+        return info
+
+    if isinstance(value, dict) and "kind" in value:
+        info.peer_kind = value["kind"]
+
+    peer_schema = await client.schema.get(kind=info.peer_kind, branch=branch)
+
+    try:
+        info.peer_rel = peer_schema.get_matching_relationship(
+            id=rel_schema.identifier or "", direction=rel_schema.direction
+        )
+    except ValueError:
+        pass
+
+    if rel_schema.cardinality == "one" and isinstance(value, list):
+        # validate the list is composed of string
+        if validate_list_of_scalars(value):
+            info.format = RelationshipDataFormat.ONE_REF
+        else:
+            info.reason_relationship_not_valid = "Too many objects provided for a relationship of cardinality one"
+
+    elif rel_schema.cardinality == "one" and isinstance(value, str):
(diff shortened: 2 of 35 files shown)

### CODE REVIEW

File: poetry.lock [hunk 1]
```diff
@@ -1,11 +1,12 @@
-# This file is automatically @generated by Poetry 1.8.4 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.1.1 and should not be changed by hand.
```
ogenstad: Locally my environment is complaining about this:

```
❯ poetry install
Installing dependencies from lock file
The lock file might not be compatible with the current version of Poetry.
Upgrade Poetry to ensure the lock file is read properly or, alternatively, regenerate the lock file with the `poetry lock` command.
```

Within the project we currently pin poetry to 1.8.5, I'm not certain if or what the impact would be but I think we should either stick with using poetry 1.x or update the pinned version in the CI. Or look at `uv`..

infrahub_sdk/spec/object.py
```diff
@@ -4,17 +4,135 @@
...
+    name: str
+    rel_schema: RelationshipSchema
+    peer_kind: str
+    peer_rel: RelationshipSchema | None = None
+    is_reference: bool = True
+    reason_relationship_not_valid: str | None = None
+
+    @property
+    def is_bidirectional(self) -> bool:
+        return bool(self.peer_rel)
```
ogenstad, line 56: What does `is_bidirectional` mean in this context?

Is it tied to this relationship direction in any way? If so it seems like something else would be required here. If not I don't quite know what the property should return.

```python
class RelationshipDirection(InfrahubStringEnum):
    BIDIR = "bidirectional"
    OUTBOUND = "outbound"
    INBOUND = "inbound"
```

pyproject.toml
```diff
@@ -63,7 +63,12 @@ pytest-xdist = "^3.3.1"
 types-python-slugify = "^8.0.0.3"
 invoke = "^2.2.0"
 towncrier = "^24.8.0"
-infrahub-testcontainers = "^1.1.0b2"
+infrahub-testcontainers = "~1.1.9"
+
+#     { version = "^1.2.2", python = ">=3.11" },
+#     { version = "1.1.9", python = "<3.11" },
+# ]
```
ogenstad: Is this a leftover?

ogenstad (approved):
One thing around this that isn't part of this PR the code already looked like this before but if we look at the example from the PR description:

```yaml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: TestingPerson
  data:
    - name: Mike Johnson
      height: 175
      best_friends: # Relationship of cardinality many that referenced existing nodes based on their HFID
        - [Jane Smith, Max]
        - [Sarah Williams, Charlie]
```

If TestPerson "Mike Johnson" existed going into this and had other best_friends relationships defined I think that those earlier relationships would be replaced with the ones defined here instead of just ensuring that the ones above also exist.

I'm not certain how I'd want it to work here, just highlighting to raise the issue. I think there are merits to both approaches, just not sure what users would expect to happen. Any thoughts?

poetry.lock
dgarros: The end goal is to migrate to `uv` but in the short term I think the best option is to upgrade to poetry `2.x`, not sure why we didn't initially but the upgrade is very simple (apart from the incompatible lock file)
I'll start a discussion on that internally

infrahub_sdk/spec/object.py
dgarros: `is_bidirectional` here means that a relationship with the same identifier exist on the remote peer, it's different than  `RelationshipDirection` but I understand it's confusing.

I'll try to find a better name

infrahub_sdk/ctl/object.py
```diff
@@ -40,8 +40,28 @@ async def load(
     client = initialize_client()
 
     for file in files:
-        file.validate_content()
-        schema = await client.schema.get(kind=file.spec.kind, branch=branch)
+        await file.validate_format(client=client, branch=branch)
```
ajtmccarty: not sure if this fits with how code is organized in the SDK, but in terms of SOLID, the `validate_format` method of `InfrahubObjectFileData` should be its own component: `ObjectFileDataValidator` or something
Something like this
```
class ObjectFileDataValidator:
    def __init__(self, client):...
    async def validate(self, branch, file): ...
```

infrahub_sdk/ctl/object.py
```diff
@@ -40,8 +40,28 @@ async def load(
...
 
-        for item in file.spec.data:
-            await file.spec.create_node(client=client, schema=schema, data=item, branch=branch)
+    for file in files:
+        await file.process(client=client, branch=branch)
```
ajtmccarty: same SOLID comment here as for `validate_format` a couple lines above
also, `process` seems more like a `create` method to me

infrahub_sdk/spec/object.py
```diff
@@ -4,17 +4,135 @@
...
+            return False
+        return not self.peer_rel.optional
+
+    @property
+    def is_valid(self) -> bool:
+        return not self.reason_relationship_not_valid
+
+
+async def get_relationship_info(
+    client: InfrahubClient, schema: MainSchemaTypesAPI, key: str, value: Any, branch: str | None = None
```
ajtmccarty: `key` and `value` could have more descriptive names. `key` seems like `name` to me, but not exactly sure about `value`. maybe a more accurate type-hint would help for `value` too

infrahub_sdk/spec/object.py
```diff
@@ -4,17 +4,135 @@
...
+                rel_info = await get_relationship_info(
+                    client=client, schema=schema, key=key, value=value, branch=branch
+                )
+                if not rel_info.is_valid:
+                    errors.append(
+                        ValidationError(
+                            identifier=key, message=rel_info.reason_relationship_not_valid or "Invalid relationship"
+                        )
+                    )
+                # TODO: Validate the sub-object too
```
ajtmccarty: does this need a follow-up issue?

BaptisteGi:
```bash
(infrahub-sdk-py3.12) ➜  bug-hunting-space infrahubctl schema check oneoone.yml
 schema 'oneoone.yml' is Valid!
diff:
    added:
        RandomStuff:
            added: {}
            changed: {}
            removed: {}
    changed: {}
    removed: {}

(infrahub-sdk-py3.12) ➜  bug-hunting-space infrahubctl object validate objects/server-template.yml
(infrahub-sdk-py3.12) ➜  bug-hunting-space 
```
- Validate command doesn't return anything when file is valid => Could be great to add a `object 'server-template.yml' is Valid!`
- For schema the command is called `check` here it's `validate` => I believe we need to align one way or the other to avoid confusions
- Maybe for an upcoming version, to have the number of node created/deleted returned by validate command if it makes sense

BaptisteGi:
```yml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: TemplateComputePhysicalServer
  data:
    - template_name: Test_pve
      interfaces:
        kind: TemplateInterfacePhysical
        data:
          - template_name: eth0
            name: eth0
            enabled: false
            not: valid
            description: This interface comes from CLI
            l2_mode: tunnel

```

```bash
(infrahub-sdk-py3.12) ➜  bug-hunting-space infrahubctl object load objects/server-template.yml
[10:29:01] INFO     Node: Test_pve                                                                                                                                                                                                               object.py:217
Error: Object is not valid
  File "/Users/Baptiste/Workspaces/Infrahub/infrahub-sdk-python/infrahub_sdk/ctl/utils.py", line 89, in async_wrapper
    return await func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Baptiste/Workspaces/Infrahub/infrahub-sdk-python/infrahub_sdk/ctl/object.py", line 46, in load
    await file.process(client=client, branch=branch)
  File "/Users/Baptiste/Workspaces/Infrahub/infrahub-sdk-python/infrahub_sdk/spec/object.py", line 312, in process
    await self.spec.process(client=client, branch=branch)
  File "/Users/Baptiste/Workspaces/Infrahub/infrahub-sdk-python/infrahub_sdk/spec/object.py", line 88, in process
    await self.create_node(client=client, schema=schema, data=item, branch=branch)
  File "/Users/Baptiste/Workspaces/Infrahub/infrahub-sdk-python/infrahub_sdk/spec/object.py", line 232, in create_node
    await cls.create_related_nodes(
  File "/Users/Baptiste/Workspaces/Infrahub/infrahub-sdk-python/infrahub_sdk/spec/object.py", line 273, in create_related_nodes
    node = await cls.create_node(
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Baptiste/Workspaces/Infrahub/infrahub-sdk-python/infrahub_sdk/spec/object.py", line 154, in create_node
    raise ValidationError(identifier=schema.kind, message="Object is not valid")
infrahub_sdk.exceptions.ValidationError: Object is not valid

(infrahub-sdk-py3.12) ➜  bug-hunting-space 
```

When creating invalid files, it's hard to tell where the issue comes from just by looking at the command output.

If possible, it would be helpful to guide users by:
- Indicating which instance is invalid (e.g. eth0 of kind `TemplateInterfacePhysical` isn't valid)
- Explaining why it's invalid (e.g. key `not` doesn't exist on kind `TemplateInterfacePhysical`)

BaptisteGi:
In some situation we could have multiple kinds hidden behind a single relationship:

image

Defining the `kind` directly within the relationship might limit us in scenarios like this.

```yml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: TemplateComputePhysicalServer
  data:
    - template_name: Test_pve
      interfaces:
        kind: TemplateInterfacePhysical
        data:
          - template_name: eth0
            name: eth0
            enabled: false
            not: valid
            description: This interface comes from CLI
            l2_mode: tunnel
```

I believe there's still a workaround by splitting interfaces into multiple files...

dgarros:
> Defining the `kind` directly within the relationship might limit us in scenarios like this.

This one has been bothering me as well and this is one of the reason I haven't merged this PR yet

there are a few options possible but I'm not sure which one is the best

### Option 1: each item in the list includes `kind` and `data`

```yaml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: TemplateComputePhysicalServer
  data:
    - template_name: Test_pve
      interfaces:
        - kind: TemplateInterfacePhysical
           data:
               template_name: eth0
               name: eth0
               enabled: false
        - kind: TemplateInterfaceLogical
           data:
               template_name: eth1
               name: eth1
               enabled: false
```

if the relationship doesn't use a generic we could have a simpler format too

```yaml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: TemplateComputePhysicalServer
  data:
    - template_name: Test_pve
      interfaces:
        - template_name: eth0
           name: eth0
           enabled: false
        - template_name: eth1
           name: eth1
           enabled: false
```

### Option 2: same format as previously but with a list 

```yaml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: TemplateComputePhysicalServer
  data:
    - template_name: Test_pve
      interfaces:
        - kind: TemplateInterfacePhysical
          data:
            - template_name: eth0
               name: eth0
               enabled: false
        - kind: TemplateInterfaceLogical
          data:
            - template_name: eth1
               name: eth1
               enabled: false
```

BaptisteGi:
Hmm, this one's tricky. I'd say option 1 is easier for users to understand. Option 2 is a bit more complex (a list within a list), but it's consistent with the rest of the file and could save users some time by not having to repeat the kind each time.

So I'd lean toward option 1, but I don't have any strong objections to option 2.

wvandeun:
Option 1 has my preference. It is more verbose, but it is very clear and explicit on what is going to happen.

> I'm not certain how I'd want it to work here, just highlighting to raise the issue. I think there are merits to both approaches, just not sure what users would expect to happen. Any thoughts?

I think it should update the relationship with the related nodes that are declared in the file. If not, we will need to find a way for a user to declare that he wants to remove a related node from an object. AKA we will have to do something very similar to how we allow people to remove things in the schema files. Most of our users don't seem to like that behavior.

BaptisteGi:
I must be missing something because I can't make it work with the new syntax:

```yaml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: TemplateComputePhysicalServer
  data:
    - template_name: Test123
      interfaces:
        - kind: TemplateInterfacePhysical
          data:
            template_name: eth0
            name: eth0
            enabled: false
            l2_mode: tunnel
```

```console
infrahub_sdk.exceptions.ValidationError: Object is not valid
```

I updated SDK to the latest version on this branch...

infrahub_sdk/spec/object.py
dgarros: Good point, i changed it to name

infrahub_sdk/spec/object.py
dgarros: I added the validation for nested objects

dgarros:
@BaptisteGi 

- The error messages for `object validate` have been updated to include more information
- The error ` Object is not valid` when using the new format has been fixed
- `object validate` now display something when the files are valid

Regarding `check` vs `validate` 
- `validate` is meant to be used only to validate the format of the local file(s), we are already using it in a few places. There is even a `infrahubctl validate` command that is meant to centralize all validations tasks
- `check` is used only for the schema and it's meant to show the diff calculated by the backend before updating a schema

having said that , I understand your point but not sure what is the best way proceed here

BaptisteGi:
When I try to add more than one component object I have an error message:

```yml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: ComputePhysicalServer
  data:
    - name: Server1234
      status: active
      location: rack00
      interfaces:
        - kind: InterfacePhysical
          data:
            name: eth0
            enabled: false
            l2_mode: access
        - kind: InterfacePhysical
          data:
            name: eth1
            enabled: false
            l2_mode: access
```

```shell
(infrahub-sdk-py3.12) ➜  bug-hunting-space infrahubctl object load objects/server.yml
[10:59:55] INFO     Node: Server1234                                                                                                                                                                                                                                                                          object.py:425
           INFO     Node: InterfacePhysical : 183623bb-e28a-9ea8-32b3-c51e3dd3f375                                                                                                                                                                                                                            object.py:425
['InterfacePhysicalUpsert'] InterfacePhysical - 183623bb-e28a-9ea8-32b3-c51e3dd3f375 cannot be added to relationship, must be of type: ['ComputePhysicalServer', 'DcimDevice', 'VirtualizationVirtualMachine'] at device
```

This works perfectly fine if I only have one row in the `interfaces` section:

```shell
(infrahub-sdk-py3.12) ➜  bug-hunting-space infrahubctl object load objects/server.yml         
[11:05:35] INFO     Node: Server1234                                                                                                                                                                                                                                                                          object.py:425
           INFO     Node: InterfacePhysical : 183623bb-e28a-9ea8-32b3-c51e3dd3f375                                                                                                                                                                                                                            object.py:425
```

dgarros:
Thanks @BaptisteGi, this one was also reported by the integration tests, it has been fixed

PR #344, opened 2025-04-04
fix(testing): remove version fixture

fatih-acar:
Remove this hardcoded fixture, otherwise using the `TestInfrahubDockerClient` class will use the `latest` Infrahub tag unless the environment variable is set. We want to use the version from the infrahub-testcontainers in that case and not fallback to `latest`.

Related to https://github.com/opsmill/infrahub/pull/6235


.github/workflows/ci.yml
```diff
@@ -288,9 +288,9 @@ jobs:
           pip install invoke toml codecov
       - name: "Install Package"
         run: "poetry install --all-extras"
-      - name: "Set environment variables for python_testcontainers"
-        run: |
-          echo INFRAHUB_TESTING_IMAGE_VER=latest >> $GITHUB_ENV
+      # - name: "Set environment variables for python_testcontainers"
```
dgarros: delete comments

pyproject.toml
```diff
@@ -64,7 +64,8 @@ pytest-xdist = "^3.3.1"
 types-python-slugify = "^8.0.0.3"
 invoke = "^2.2.0"
 towncrier = "^24.8.0"
-infrahub-testcontainers = "^1.1.0b2"
+#infrahub-testcontainers = "^1.2.4"
```
dgarros: Fix version

tests/unit/sdk/checks/conftest.py
```diff
@@ -17,5 +17,6 @@ async def mock_gql_query_my_query(httpx_mock: HTTPXMock) -> HTTPXMock:
         method="POST",
         json=response,
         url="http://localhost:8000/api/query/my_query?branch=main&update_group=false",
+        is_reusable=True,
```
dgarros: How is that related to this change ?

dgarros (approved):
LGTM, would be good to merge it soon

tests/unit/sdk/checks/conftest.py
dgarros: I created a dedicated PR to upgrade `pytest-httpx` #354

fatih-acar: Yes it was actually related to pytest-httpx... I've rebased this branch on top of yours

fatih-acar:
@dgarros I had to drop python 3.9 support, not sure if this was expected...

changelog/+5e549d33.removed.md
```diff
@@ -0,0 +1 @@
+Remove Python 3.9 support.
```
dgarros: Why do we need to remove support for Python 3.9 ?

fatih-acar: `infrahub_testcontainers` 1.2 does not support python 3.9.
I have reverted the change and made `infrahub_testcontainers` a dependency only for python > 3.9. The drawback is that integration tests cannot run with 3.9 anymore (which is still better than full SDK depreciation for 3.9).

PR #348, opened 2025-04-07
Update protocols to support Template and fix support for Sync

### LINKED ISSUE #329: task: Create Protocols for Template and Profile
### Component

Python SDK

### Task Description

Currently when Template or Profile is enable on an object in the schema, `infrahubctl protocols` will not generate protocols for the Profiles and Templates nodes.

As a result, it's not possible to achieve full typing for these objects.

To ensure that we can provide a full coverage and to improve the user experience, we need to extend the current command to generate protocols for Profiles and Templates nodes/generics in the schema.

dgarros:
Fixes #329 

This PR includes a few enhancements and fixes around Python Protocols
- Move the code and the template into a dedicated directory
- Add unit tests
- Fix support for Sync
- Add support for object Template
- Add guide to use Protocols

This PR also adds some invoke commands to help with the docs
- `invoke lint-doc`
- `invoke lint-vale`
- `invoke lint-markdownlint`

If `vale` and/or `markdownlint-cli2` are not installed, these commands will be skipped


docs/docs/python-sdk/guides/python-typing.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+# Type checker can now verify all attributes
+blue_tag = client.get(BuiltinTag, name__value="blue")  # blue_tag is of type BuiltinTag
+blue_tag.description.value = "The blue tag"  # No type errors
+blue_tag.save()
+```
+
+:::note Python Protocols
+
+Python Protocols, introduced in PEP 544, define a set of method and property signatures that a class must implement to be considered a match, enabling structural subtyping (also known as "duck typing" with static checks). They allow you to specify behavior without requiring inheritance, making code more flexible and type-safe.
```
ogenstad, line 60: Link to https://typing.python.org/en/latest/spec/protocol.html?

docs/docs/python-sdk/guides/python-typing.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+:::
+
+## Generating custom protocols based on your schema
+
+You can generate Python Protocols for your own models using the `infrahubctl protocols` command. This supports both synchronous and asynchronous Python code.
+
+```shell
+# Generate protocols for your schema
+infrahubctl protocols --schemas schemas/tag.schema.yml --out lib/protocols.py
```
ogenstad: We should mention that you can also target the server directly instead of reading from a local file.

infrahub_sdk/protocols_generator/generator.py
```diff
@@ -71,29 +56,44 @@ def __init__(self, schema: dict[str, MainSchemaTypesAll]):
...
 
     def render(self, sync: bool = True) -> str:
         jinja2_env = jinja2.Environment(loader=jinja2.BaseLoader(), trim_blocks=True, lstrip_blocks=True)
-        jinja2_env.filters["inheritance"] = self._jinja2_filter_inheritance
         jinja2_env.filters["render_attribute"] = self._jinja2_filter_render_attribute
         jinja2_env.filters["render_relationship"] = self._jinja2_filter_render_relationship
+        jinja2_env.filters["syncify"] = self._jinja2_filter_syncify
 
-        template = jinja2_env.from_string(PROTOCOLS_TEMPLATE)
+        template = jinja2_env.from_string(load_template())
```
ogenstad, line 77: I don't think we need to change this in this PR but I think it would make sense to the templating logic we have elsewhere within the SDK to render these templates. Possibly we should come to a final decision regarding #337 first if we should leave it as async by default or not.

tasks.py
```diff
@@ -165,12 +171,46 @@ def lint_ruff(context: Context) -> None:
...
+    print(" - Check documentation with markdownlint-cli2")
+    exec_cmd = "markdownlint-cli2 **/*.{md,mdx} --config .markdownlint.yaml"
+    with context.cd(MAIN_DIRECTORY_PATH):
+        context.run(exec_cmd)
+
+
+@task
+def lint_vale(context: Context) -> None:
+    """Run vale to check all documentation files."""
+    if not is_tool_installed("vale"):
```
ogenstad, line 190: I'm not sure about these is_tool_installed, it feels like we at some point could get into a situation where we're just running the lint_all command and we move around something so that these tools aren't installed. Then we wouldn't notice it in the pipeline.

tests/unit/sdk/test_protocols_generator.py
```diff
@@ -0,0 +1,58 @@
...
+        (
+            False,
+            ["LineageSource", "CoreNode", "CoreObjectTemplate"],
+            ["LineageSource", "CoreNode", "CoreObjectTemplate"],
+        ),
+    ],
+)
+async def test_filter_syncify(sync, input, output):
+    assert CodeGenerator._jinja2_filter_syncify(value=input, sync=sync) == output
+    assert CodeGenerator._jinja2_filter_syncify(value=input, sync=sync) == output
```
ogenstad: I liked this pattern of setting up dataclasses for the test cases, I think it adds clarity with regards to what the expected input fields are along with the possibility to give each test a relevant name.

https://github.com/opsmill/infrahub-sdk-python/blob/v1.10.2/tests/unit/sdk/test_template.py#L45-L81

ogenstad (approved):

tasks.py
dgarros: If someone doesn't have the right tools installed, the pipeline will still catch the issue right now.

These methods are mainly for us to make it easier to validate a branch locally but since not everyone will need them I would prefer to not require everyone to install `vale` & `markdownlint-cli2`

tests/unit/sdk/test_protocols_generator.py
dgarros: Good point I've updated the test

PR #350, opened 2025-04-09
set the default branch correctly on the client store in a generator

wvandeun:
When a generator runs on a task worker, the client that gets passed when we instantiate a generator, does not have the default branch set to the current active branch.

This leads to a situation where the default branch on the store never gets properly set. Due to this objects that get stored end-up in the correct NodeStoreBranch, but when we retrieve them we always try to fetch them from the `main` NodeStoreBranch.

### DIFF
--- a/changelog/+store_default_branch_generator.fixed.md
+++ b/changelog/+store_default_branch_generator.fixed.md
@@ -1,0 +1,1 @@
+fixes an issue where the default branch of the client store was not properly set in a generator
--- a/infrahub_sdk/generator.py
+++ b/infrahub_sdk/generator.py
@@ -40,6 +40,7 @@ class InfrahubGenerator:
         self.generator_instance = generator_instance
         self._init_client = client.clone()
         self._init_client.config.default_branch = self._init_client.default_branch = self.branch_name
+        self._init_client.store._default_branch = self.branch_name
         self._client: InfrahubClient | None = None
         self._nodes: list[InfrahubNode] = []
         self._related_nodes: list[InfrahubNode] = []


dgarros (commented):
this change is probably alright but it feels like we should also fix the issue in Infrahub and instantiate the SDK with the right default branch

ogenstad:
> this change is probably alright but it feels like we should also fix the issue in Infrahub and instantiate the SDK with the right default branch

I think this part will be refactored when we look at adding support for using this store and converting GraphQL objects into InfrahubNode objects for transforms as well.

To highlight the issue what happens in the generator looks like this:

```python
        self._init_client = client.clone()
        self._init_client.config.default_branch = self._init_client.default_branch = self.branch_name
```

With this we initialize a new client using the old config and then override the previous default branch so this part works. The problem is that within `client.clone()` we call the `._initialize()` method which in turn configures the store with the default branch which in most cases will be `main`.

As the SDK client running the generator will have the correct branch for the proposed change it will create objects in the desired branch.

The problem then comes when we start to use the store as the store has a .set() method which will default to `node.get_branch()` if a branch wasn't specified for the set command. but the .get() won't have a node to refer to so it will use the `main` branch.

I think this fixes the immediate issue and what we should change moving forward is that `client.clone()` should take an optional "branch" parameter, I've also been thinking of providing a parameter to indicate if you want a sync or async client back and potentially an option to pre populate the schema (so that the generator client doesn't have to start by downloading the schema.

ogenstad (approved):

PR #351, opened 2025-04-10
Fix attribute updates with original data

### LINKED ISSUE #349: bug: Updating node attribute to its original value
### Component

Python SDK

### Infrahub SDK version

1.10.1

### Current Behavior

If we create a node, fetch it from the server, then modifies an attribute value, and re-modify this value with the original value set during creation, the value is not updated:

This is because we consider blue to be the original data within the `_strip_unmodified` logic supposed to not send non-updated fields within the GraphQL query. We may need a similar fix to what has been done [for relationships,](https://github.com/opsmill/infrahub/commit/24b96d018ba1ba942e0e854c52c14d3765c4d799) ie adding a flag to indicate whether the attribute comes from initialization.

### Expected Behavior

Attribute value should be correctly updated

### Steps to Reproduce

```python
owner = await client.create(kind="TestPerson", name="owner", item="blue")
await owner.save()
owner = await client.get(kind="TestPerson", id=owner.id)

owner.item.value = "red"
await owner.save()

owner.item.value = "blue"
await owner.save()
node = await client.get(kind="TestPerson", id=owner.id)
assert node.item.value == "blue"  # --> fails, it's red 
```

### Additional Information

The issue still happens with `populate_store=False` within `get` calls.

LucasG0:
Fixes https://github.com/opsmill/infrahub-sdk-python/issues/349

Corresponding infrahub pipeline: https://github.com/opsmill/infrahub/pull/6266

### DIFF
--- a/infrahub_sdk/node.py
+++ b/infrahub_sdk/node.py
@@ -82,17 +82,18 @@ class Attribute:
 
         self.id: str | None = data.get("id", None)
 
-        self.value: Any | None = data.get("value", None)
+        self._value: Any | None = data.get("value", None)
+        self.value_has_been_mutated = False
         self.is_default: bool | None = data.get("is_default", None)
         self.is_from_profile: bool | None = data.get("is_from_profile", None)
 
-        if self.value:
+        if self._value:
             value_mapper: dict[str, Callable] = {
                 "IPHost": ipaddress.ip_interface,
                 "IPNetwork": ipaddress.ip_network,
             }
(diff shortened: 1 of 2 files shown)


ogenstad (approved):
LGTM

A potential future consideration would be if we should update the same field if we call save additional times. But it should only be a very corner case optimisation. And even then it could very well be that the attribute has been updated on the server already and we want to overwrite it if it differs.

PR #354, opened 2025-04-13
Various housekeeping tasks 

dgarros:
- Update `pytest-httpx` and set all responses as reusable (required by the latest version of pytest-httpx
- Add a fixture to always reset some environment variables before running tests
- Fix type for `metadata` in InfrahubFileData

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,10 +1,13 @@
 import asyncio
+import os
 
 import pytest
 
 from infrahub_sdk.ctl import config
 
 pytest_plugins = ["pytester"]
+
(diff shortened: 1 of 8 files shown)


tests/conftest.py
```diff
@@ -20,3 +23,19 @@ def event_loop():
...
+    for name in ENV_VARS_TO_CLEAN:
+        original_values[name] = os.environ.get(name)
+        if original_values[name] is not None:
+            del os.environ[name]
+
+    yield
+
+    for name in ENV_VARS_TO_CLEAN:
+        if original_values[name] is not None:
+            os.environ[name] = original_values[name]
```
ogenstad, line 41: Given that this will run until the session has completed I don't know if there's a point of resetting the environment variables after everything has completed. There should be nothing left that would use them at that point.

ogenstad (approved):

PR #360, opened 2025-04-16
Add method to populate the cache of the schema manually

### LINKED ISSUE #342: feature: Add an option to manually populate the cache of the schema
### Component

Python SDK

While doing unit tests it would be very useful if we could manually populate the cache of the schema directly. Currently in order to write unit tests we need to mock the query to `/api/schema`.

The proposal is to provide a method to manually provide a `SchemaRoot` to populate the cache :
- `client.schema.set_cache`

it should be possible to populate the cache of each branch individually

### Describe the Use Case

Improve the developer experience while writing unit tests

### Additional Information

_No response_

dgarros:
Fixes #342 

This PR mainly add a method to manually populate the cache for the schema manager, the primary use case for this method is to simplify the initialization of the client during unit tests

I also deprecated the timeout parameter while fetching the schema, it was not consistently used within the code and overall I think using the default timeout should be enough for the schema.
I'll open another issue shortly to track its removal

### DIFF
--- a/changelog/+schema-set-cache.added.md
+++ b/changelog/+schema-set-cache.added.md
@@ -1,0 +1,1 @@
+Add method schema.set_cache() to populate the cache manually (primarily for unit testing)
--- a/changelog/+schema-set-cache.added.md
+++ b/changelog/+schema-set-cache.added.md
@@ -1,1 +1,1 @@
-Add method schema.set_cache() to populate the cache manually (primarily for unit testing)
+Add method `client.schema.set_cache()` to populate the cache manually (primarily for unit testing)
--- a/infrahub_sdk/schema/__init__.py
+++ b/infrahub_sdk/schema/__init__.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import warnings
 from collections.abc import MutableMapping
 from enum import Enum
 from time import sleep
@@ -90,6 +91,13 @@ MainSchemaTypesAll: TypeAlias = Union[
 
 
 class InfrahubSchemaBase:
+    client: InfrahubClient | InfrahubClientSync
+    cache: dict[str, BranchSchema]
+
+    def __init__(self, client: InfrahubClient | InfrahubClientSync):
+        self.client = client
+        self.cache = {}
+
     def validate(self, data: dict[str, Any]) -> None:
         SchemaRoot(**data)
 
@@ -101,6 +109,14 @@ class InfrahubSchemaBase:
                     identifier=identifier,
                     message=f"{key} is not a valid value for {identifier}",
                 )
+
+    def set_cache(self, schema: dict[str, Any] | BranchSchema, branch: str | None = None) -> None:
+        branch = branch or self.client.default_branch
+
+        if isinstance(schema, dict):
+            schema = BranchSchema.from_api_response(data=schema)
+
+        self.cache[branch] = schema
 
     def generate_payload_create(
         self,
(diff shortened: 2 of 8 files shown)


infrahub_sdk/schema/__init__.py
```diff
@@ -639,49 +666,39 @@ def add_dropdown_option(
...
+            branch: Name of the branch to fetch the schema for.
+            timeout: Overrides default timeout used when querying the GraphQL API. Specified in seconds (deprecated).
+            populate_cache: Whether to populate the cache with the fetched schema. Defaults to True.
 
         Returns:
             dict[str, MainSchemaTypes]: Dictionary of all schema organized by kind
         """
-        branch_schema = self._fetch(branch=branch, namespaces=namespaces, timeout=timeout)
+        if timeout:
+            self._deprecated_schema_timeout()
```
ogenstad, line 691: We should also deprecate timeout in the `client.create()` methods it's only used to pass the argument to the schema function.

changelog/+schema-set-cache.added.md
```diff
@@ -0,0 +1 @@
+Add method schema.set_cache() to populate the cache manually (primarily for unit testing)
```
ogenstad: I've typically written this as `client.schema` instead of just `schema`. Don't know which is best or easiest to understand.

infrahub_sdk/schema/__init__.py
```diff
@@ -187,11 +203,18 @@ def _parse_schema_response(response: httpx.Response, branch: str) -> MutableMapp
 
         return data
 
+    @staticmethod
+    def _deprecated_schema_timeout() -> None:
+        warnings.warn(
+            "The 'timeout' parameter is deprecated while fetching the schema and will be removed version 2.0.0 of the Infrahub Python SDK. "
+            "Use client.default_timeout instead.",
```
ogenstad, line 219: Will you also create an issue in the 2.0 milestone to remove this.

infrahub_sdk/schema/__init__.py
```diff
@@ -102,6 +110,14 @@ def validate_data_against_schema(self, schema: MainSchemaTypesAPI, data: dict) -
                     message=f"{key} is not a valid value for {identifier}",
                 )
 
+    def set_cache(self, schema: dict[str, Any] | BranchSchema, branch: str | None = None) -> None:
```
ogenstad: I'd add a short docstring here to explain that you can use this to prepopulate a schema.

infrahub_sdk/schema/__init__.py
```diff
@@ -102,6 +110,14 @@ def validate_data_against_schema(self, schema: MainSchemaTypesAPI, data: dict) -
...
+        branch = branch or self.client.default_branch
+
+        if isinstance(schema, dict):
+            schema = BranchSchema.from_api_response(data=schema)
```
ogenstad, line 126: In a way I'd prefer if .set_cache took a BranchSchema object directly. That way there's no confusion regarding what the `dict[str, Any]` should look like. It would mostly be relevant if there are any errors in the schema, then those errors would be raised in the correct location. On the other hand it might be slightly easier for some users if they don't have to convert it into a dict.

ogenstad (approved):
LGTM. For the comments regarding client.timeout I think it's fine to deprecate those in an upcoming PR to as long as we do it a while before 2.0.

I think it could be useful to have some method to get the schema from the cache so that the users are returned a BranchSchema object that they'd then can store locally as ysml or yaml.

infrahub_sdk/schema/__init__.py
dgarros: It's already covered because if someone use `timeout` while creating a node, it will generate a deprecation warning.
Having said that I've updated the changelog to reflect that properly

infrahub_sdk/schema/__init__.py
dgarros: Will do

infrahub_sdk/schema/__init__.py
dgarros: `BranchSchema` is primarily an internal object so in most cases I would expect users to use the `dict` most of the time.
To clarify, I've added support for `SchemaRootAPI` as well which is the format returned by the `/api/schema` endpoint

infrahub_sdk/schema/__init__.py
```diff
@@ -102,6 +110,23 @@ def validate_data_against_schema(self, schema: MainSchemaTypesAPI, data: dict) -
                     message=f"{key} is not a valid value for {identifier}",
                 )
 
+    def set_cache(self, schema: dict[str, Any] | SchemaRootAPI | BranchSchema, branch: str | None = None) -> None:
```
LucasG0, line 113: I think `set_cache` should be a private method as it's meant for internal use.

dgarros: it's not meant for internal use, it's meant for user to use during unit tests

PR #363, opened 2025-04-16
Add `menu validate` command to validate the format of menu files. 

dgarros:
and a few fixes and cleanup following the changes on `object load` command

### DIFF
--- a/infrahub_sdk/ctl/menu.py
+++ b/infrahub_sdk/ctl/menu.py
@@ -7,10 +7,14 @@ from rich.console import Console
 from ..async_typer import AsyncTyper
 from ..ctl.client import initialize_client
 from ..ctl.utils import catch_exception, init_logging
-from ..exceptions import ObjectValidationError
+from ..exceptions import ObjectValidationError, ValidationError
 from ..spec.menu import MenuFile
 from .parameters import CONFIG_PARAM
-from .utils import load_yamlfile_from_disk_and_exit
+from .utils import (
+    display_object_validate_format_error,
+    display_object_validate_format_success,
+    load_yamlfile_from_disk_and_exit,
+)
 
 app = AsyncTyper()
 console = Console()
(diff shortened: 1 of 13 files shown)


infrahub_sdk/spec/menu.py
```diff
@@ -28,7 +28,7 @@ def spec(self) -> InfrahubMenuFileData:
         return self._spec
 
     def validate_content(self) -> None:
-        super().validate_content()
+        InfrahubFile.validate_content(self)
```
ogenstad, line 31: Why are we making this change? Swapping from inheriting from `InfrahubFile` and moving to `ObjectFile` instead but we override the spec method from ObjectFile completely and then call the .validate_content on InfrahubFile I assume to avoid using the one from `ObjectFile`. Am I missing something here?

infrahub_sdk/ctl/menu.py
```diff
@@ -40,21 +44,54 @@ async def load(
...
+            has_errors = True
+            console.print(f"[red] {exc!s}")
+
+    if has_errors:
+        raise typer.Exit(1)
+
+
+@app.command()
+@catch_exception(console=console)
+async def validate(
```
ogenstad, line 72: Would be nice with a unit test even if it's just for the failure scenario

ogenstad (approved):
LGTM, though I'd need some clafification regarding the MenuFile class.

infrahub_sdk/spec/menu.py
dgarros: I agree this isn't ideal, and it calls for a larger redesign / cleaned of the `ObjectFile` / `YamlFile` / `MenuFile` but I think this is a larger effort and I would prefer to manage that separately
I will open a dedicated issue to track that

ogenstad: Sounds good to me

infrahub_sdk/ctl/menu.py
dgarros: Good point, I added some unit tests

infrahub_sdk/spec/menu.py
dgarros: https://github.com/opsmill/infrahub-sdk-python/issues/367

PR #371, opened 2025-04-22
Infrahubctl repo Init command

minitriga:
`infrahubctl repository init /tmp/test` will create a new repository with some basic structure to get started with Infrahub. It will prompt the user what they would like to include.


docs/docs/infrahubctl/infrahubctl-repository.mdx
```diff
@@ -47,6 +48,26 @@ $ infrahubctl repository add [OPTIONS] NAME LOCATION
...
+
+**Usage**:
+
+```console
+$ infrahubctl repository init [OPTIONS] DST
+```
+
+**Arguments**:
+
+* `DST`: [required]
```
wvandeun: It's not clear what DST is, or what it is that we are supposed to provide.

docs/docs/infrahubctl/infrahubctl-repository.mdx
```diff
@@ -47,6 +48,26 @@ $ infrahubctl repository add [OPTIONS] NAME LOCATION
...
+
+**Options**:
+
+* `--data PATH`
```
wvandeun: Same, what is data in this case? A Path to a file? To what file? To a directory?

docs/docs/infrahubctl/infrahubctl-repository.mdx
wvandeun: Should DST be optional? If not provided, we could make a directory from the project name that we prompt for.

infrahub_sdk/ctl/example_repo/copier.yml
wvandeun, line 1: Can we make this options for the CLI command?

minitriga: This is what `--data` does allows user to pass in arguments via a file. Unfortunately defining options in two places I.E in the typer CLI and then in Copier prevents the copier prompts from appearing and will only use the CLI argument defaults. The idea of copier is that it remembers what the user selected previously so that repos can be updated in future where as the typer CLI does not.

docs/docs/infrahubctl/infrahubctl-repository.mdx
minitriga: I have added helpers to the arguments for the docs to be better.

docs/docs/infrahubctl/infrahubctl-repository.mdx
minitriga: Unfortunately from what I read copier requires the destination to be known upfront before the prompt appears using the `run_copy` function. By default copier will use `cwd` if nothing is passed in. which would require this workflow. We are using package name more for the `pyproject.toml` package name. 

```
mkdir infrahub-repo`
cd infrahub-repo
infrahubctl repository init
```

Rather than just `infrahubctl repository init infrahub-repo`

minitriga: DST now also has help to explain what it is.

ogenstad:
While I can see that this could be helpful for some people I think it would also be nice if we didn't have to add extra dependencies to the ctl. :)

Other solutions could potentially be to have some example repos that are easy to clone and use as a starting point.

Anyway some observations about this PR:

* The tool can generate a "lib" folder, there's also an example of how this is used in combination with an "infrahubctl run" script. I don't think the same type of imports would work with a Transform, Check or Generator. I could be wrong on with that  `from lib.example import print_nodes`, but if that's the case I think it might be confusing to include an example that would work with run but not the others
* We're talking about migrating to uv, perhaps we'll want do do the same for these packages. It might be simple enough to change that after a migration though
* I'd have "invoke" as a dev dependency instead of a main one
* The use of Transforms or Generators would imply that Queries would be required, don't know if this can be done with copier?

wvandeun:
I would consider not loading the schema file. The recommendation is that when you manage the schema through the repository that you don't use `infrahubctl schema` for that purpose. 

Some people find it very confusing that schema changes loaded using `infrahubctl` don't get synced into the schema files, when they use a normal external repository.

PR #376, opened 2025-04-24
Add ability to use convert_query_response with Python Transforms

### LINKED ISSUE #281: feature: Add `convert_query_response` to Python Transforms
### Component

infrahubctl CLI, Python SDK

Allow the same functionality of converting the GraphQL response into `InfrahubNode` objects for Python transformations.

### Describe the Use Case

We fetch device IDs in our GraphQL query for the transform and would like to fetch those devices later on in the transform by using the Infrahub store that holds the objects from the query in the store to be retrieved later. This allows us to not have to run further queries to get the device information within the Python transform.

### Additional Information

_No response_

ogenstad:
This PR introduced support for "convert_query_response" for Python Transforms.

Changes:
* The client method to clone the config has been moved to the config object itself and modified to not only rely on deepcopy. The reason for this comes from the tests within the Infrahub directory where we've overridden the client transport to a method that doesn't allow for pickling. As the transport itself shouldn't be of interest for the config we reuse some of the elements in the cloned config
* As some of the methods and logic is shared between Generators and Transforms a new "InfrahubOperation" class has been introduced and logic has been transfered and consolidated from both Generators and Transforms to this new base class.
* Some new protocols were added
* I added some new integration tests for both Transforms and Generators, previously these weren't tested within the SDK but and instead relied on tests within the Infrahub repository, but added to the SDK tests to have better coverage
* I ran into some issues with the tests for the "infrahubctl generator" typer app as it's an async app, after some troubleshooting I noticed that the `typer.testing.CliRunner` doesn't seem to support tests for async commands like this. Instead of spending more time on that I opted to call the `generator` function directly within the code for now. It would be good with a permanent solution for that but I didn't want to spend more time on it now.

Fixes #281

Backend PR: https://github.com/opsmill/infrahub/pull/6363

### DIFF
--- a/infrahub_sdk/ctl/cli_commands.py
+++ b/infrahub_sdk/ctl/cli_commands.py
@@ -41,6 +41,7 @@ from ..ctl.utils import (
 )
 from ..ctl.validate import app as validate_app
 from ..exceptions import GraphQLError, ModuleImportError
+from ..node import InfrahubNode
 from ..protocols_generator.generator import CodeGenerator
 from ..schema import MainSchemaTypesAll, SchemaRoot
 from ..template import Jinja2Template
@@ -330,7 +331,12 @@ def transform(
         console.print(f"[red]{exc.message}")
         raise typer.Exit(1) from exc
 
-    transform = transform_class(client=client, branch=branch)
+    transform = transform_class(
+        client=client,
+        branch=branch,
+        infrahub_node=InfrahubNode,
+        convert_query_response=transform_config.convert_query_response,
+    )
     # Get data
     query_str = repository_config.get_query(name=transform.query).load_query()
     data = asyncio.run(
--- a/infrahub_sdk/operation.py
+++ b/infrahub_sdk/operation.py
@@ -1,0 +1,80 @@
+from __future__ import annotations
+
+import os
+from typing import TYPE_CHECKING
+
+from .repository import GitRepoManager
+
+if TYPE_CHECKING:
+    from . import InfrahubClient
+    from .node import InfrahubNode
+    from .store import NodeStore
+
+
+class InfrahubOperation:
+    def __init__(
+        self,
+        client: InfrahubClient,
+        infrahub_node: type[InfrahubNode],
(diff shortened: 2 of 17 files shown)


infrahub_sdk/ctl/cli_commands.py
```diff
@@ -330,7 +331,12 @@ def transform(
         console.print(f"[red]{exc.message}")
         raise typer.Exit(1) from exc
 
-    transform = transform_class(client=client, branch=branch)
+    transform = transform_class(
+        client=client,
+        branch=branch,
+        infrahub_node=InfrahubNode,
+        convert_query_response=transform_config.convert_query_response,
+    )
```
ajtmccarty, line 339: do you think it is worth passing the whole `transform_config` here instead of just `convert_query_response`? might make it easier to add other parameters like this in the future

infrahub_sdk/operation.py
```diff
@@ -0,0 +1,80 @@
...
+        return self.branch
+
+    @property
+    def store(self) -> NodeStore:
+        """The store will be populated with nodes based on the query during the collection of data if activated"""
+        return self._init_client.store
+
+    @property
+    def nodes(self) -> list[InfrahubNode]:
+        """Returns nodes collected and parsed during the data collection process if this feature is enables"""
```
ajtmccarty: `enables` -> `enabled`

ajtmccarty (approved):

infrahub_sdk/operation.py
ogenstad: 🤦 thanks

infrahub_sdk/ctl/cli_commands.py
ogenstad: Yes I do, but it would require more changes in the backend when we use this outside of the CTL in situations where we haven't stored the config anywhere. I think for now we should leave it like this to not change too many things. But could be that we should do a larger refactor and consider what parts should be kept in the database and what options should only exist in Git.

PR #403, opened 2025-05-09
fix bug in Timestamp.add

ajtmccarty:
instantiate the new timestamp as `self.__class__` instead of `Timestamp` so that we keep the correct class

for example, infrahub has its own `Timestamp` class that inherites from the sdk's `Timestamp` and calling `.add` on an infrahub `Timestamp` yields an sdk `Timestamp`


dgarros (approved):
Looks good,
I think we should do the same change for `subtract` as well just below `add`

PR #409, opened 2025-05-12
Split node.py into multiple files

dgarros:
This PR is splitting the `node.py` into multiple files, it's currently close to 800 lines long.

To reduce the risk of side effect, I haven't changed the content of the file, with the exception that in some places we are now using Protocols to avoid some circular dependencies.

The change is expected to be backward compatible because everything that was part of `node.py` is exposed from `infrahub_sdk.node`


infrahub_sdk/node/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+    "RelatedNode",
+    "RelatedNodeBase",
+    "RelatedNodeSync",
+    "RelationshipManager",
+    "RelationshipManagerBase",
+    "RelationshipManagerSync",
+]
+
+
+def parse_human_friendly_id(hfid: str | list[str]) -> tuple[str | None, list[str]]:
```
ajtmccarty: it looks like this is only used in `store.py`, maybe it can just move over there so that there is no actual code in this `__init__.py`

ajtmccarty (approved):

PR #412, opened 2025-05-17
Refactor InfrahubNode to avoid dynamic class creation

ogenstad:
The main goal of this PR is to remove two lines:

```python
self.__class__ = type(f"{schema.kind}InfrahubNode", (self.__class__,), {})
```

and:

```python
self.__class__ = type(f"{schema.kind}InfrahubNodeSync", (self.__class__,), {})
```

What these lines do are to create a dynamic Python class for each object we define with the SDK when we create an InfrahubNode or InfrahubNodeSync object.

The main need for the creation of a new object came from the `generate_relationship_property()` function that created a setter/getter for cardinality=one relationships on the new dynamically created class.

This allowed you to override the data directly on a relationship of cardinality one at the attribute level. In pseudo code if we had a node that looked like this:

```python
class InfraInterface:
    name: Attribute
    device: Relationship
```

Then if we had an instance of this as an InfrahubNode object:

```python
interface.name.value = "Ethernet1" # Correctly sets the name attribute
interface.name = "Ethernet1" # Previously this would overwrite the name attribute and instead set "name" as a simple `str`, which wouldn't work
interface.device = "453a7da5-2c71-4cac-b96f-e00b787e0436" # this uses the attribute setter and assigns a value under the relationship
```

With the new implementation both attributes and cardinality=one relationships have both a setter and a getter, so the example above with `interface.name = "Ethernet1"`. Using the same type of logic we'd be able to add a setter to cardinality=many relationships as well, we'd only have to decide how it would work if it should add a relationship or replace all previous ones.

This PR changes the way we would have to manage typing, there are a few variations we could do for this an example can be seen in the two current commits of this PR within the file query_groups.py on the line `self.previous_members = group.members.peers` the first commit only updates the rules, the second commit removes the ignore rules but could make the code a bit harder to read.

Test pipeline within the Infrahub project: https://github.com/opsmill/infrahub/pull/6492

One change with this is that when using a Python REPL and using `dir(my_infrahub_node_object)` the name of the attributes and relationships won't be seen as attribute objects.

I'm also thinking if we should have `infrahubnode_` as a reserved prefix so that we can add public functions using that name without having to fear future conflicts.

### DIFF
--- a/infrahub_sdk/query_groups.py
+++ b/infrahub_sdk/query_groups.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any
 
 from .constants import InfrahubClientMode
@@ -19,7 +20,7 @@ class InfrahubGroupContextBase:
         self.related_node_ids: list[str] = []
         self.related_group_ids: list[str] = []
         self.unused_member_ids: list[str] | None = None
-        self.previous_members: list[RelatedNodeBase] | None = None
+        self.previous_members: Sequence[RelatedNodeBase] | None = None
         self.previous_children: list[RelatedNodeBase] | None = None
         self.identifier: str | None = None
         self.params: dict[str, str] = {}
@@ -101,7 +102,7 @@ class InfrahubGroupContext(InfrahubGroupContextBase):
         if not store_peers:
             return group
 
-        self.previous_members = group.members.peers  # type: ignore[attr-defined]
+        self.previous_members = group._get_relationship_many(name="members").peers
         return group
 
     async def delete_unused(self) -> None:
@@ -195,7 +196,7 @@ class InfrahubGroupContextSync(InfrahubGroupContextBase):
         if not store_peers:
             return group
 
-        self.previous_members = group.members.peers  # type: ignore[attr-defined]
+        self.previous_members = group._get_relationship_many(name="members").peers
         return group
 
     def delete_unused(self) -> None:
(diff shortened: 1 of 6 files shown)


infrahub_sdk/query_groups.py
```diff
@@ -94,7 +94,7 @@ async def get_group(self, store_peers: bool = False) -> InfrahubNode | None:
         if not store_peers:
             return group
 
-        self.previous_members = group.members.peers  # type: ignore[attr-defined]
+        self.previous_members = group._relationship_cardinality_many_data["members"].peers
```
dgarros: this one feels wrong to me, I think we should either use the Protocol or we should add a public method to access an attribute or a relationship by its name

```python 
group.get_relationship_many(name="members").peers
```

dgarros (approved):

infrahub_sdk/query_groups.py
ogenstad: Yes I agree, this was mainly to highlight the typing issue. I'm a bit reluctant to add new public methods, at least until we create some reservations for what we use. I like the way pydantic has done this with `model_`. Still even the code above could be a bit cleaner with a private method, alternatively we could use the option from the earlier commit where this was left as `group.members.peers` but with another mypy ignore code.

ogenstad:
This is ready for review now. I've added some private methods to get attributes or relationships as well as the `__dir__()` methods along with some changes for the typing as a result of this change.

ajtmccarty (approved):

ogenstad:
Rebased and squashed.

PR #422, opened 2025-05-23
Add objects field within InfrahubConfig

LucasG0:


infrahub_sdk/testing/schemas/car_person.py
```diff
@@ -48,6 +48,7 @@ def schema_person_base(self) -> NodeSchema:
             namespace=NAMESPACE,
             include_in_menu=True,
             label="Person",
+            default_filter="name__value",
```
LucasG0, line 51: We need this because `TestProposeChangeRepository` imports the git repository which loads the data where peers are referenced using notation `"John Doe"`, which does not work when default_filter is not set on the schema. We use this notation instead of the hfid notation [John Doe] because backend tests also import the repository but with schemas only defining `default_filter`. If above schema is meant to be used by users, it would probably be cleaner to define the hfid on backend tests schemas and use hfid notation, but these schemas are used by a lot of tests so there is a risk that some tests break.

gmazoyer (approved):

ogenstad (approved):

PR #427, opened 2025-06-04
update version spec for whenever dependency

wvandeun:
infrahub-testcontainers v1.3.0b3 defines a new dependency on prefect 3.4.1
prefect v3.4.1 has a dependency specification ">=0.7.3,0.9.0" for Python 3.13 which is incompatible with the current version specification we had.


wvandeun:
not needed anymore

PR #446, opened 2025-06-14
`infrahubctl check` to find and filter the same as PC checks

### LINKED ISSUE #445: bug: Check attempts to use the parameter key, but runs into `AttributeError` if not a valid attribute.
### Component

infrahubctl

### Infrahub SDK version

1.13.0

### Current Behavior

A PC check run will properly grab the correct attribute on an `InfrahubNode` based on the value of the parameter, but `infrahubctl` attempts to use the parameter key for the attribute fetch.

```
query TagsQuery($tag_name: String!) {
  BuiltinTag(name__value: $tag_name) {
    edges {
      node {
        name {
          value
        }
        description {
          value
        }
      }
    }
  }
}
```

```
check_definitions:
  - name: "check_color_tags_name"
    class_name: "ColorTagsCheck"
    file_path: "checks/tags_check/tags_check.py"
    targets: "ColorTags"
    parameters:
      tag_name: "name__value"
```

This would result in the following error when running the check via `infrahubctl`.

```shell
Error: 'InfrahubNode' object has no attribute 'tag_name'
  File "/Users/Mikhail/cloned/ih-guide/.venv/lib/python3.11/site-packages/infrahub_sdk/ctl/utils.py", line 100, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Mikhail/cloned/ih-guide/.venv/lib/python3.11/site-packages/infrahub_sdk/ctl/check.py", line 73, in run
    aiorun(
  File "/Users/Mikhail/.pyenv/versions/3.11.10/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/Users/Mikhail/.pyenv/versions/3.11.10/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Mikhail/.pyenv/versions/3.11.10/lib/python3.11/asyncio/base_events.py", line 654, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/Mikhail/cloned/ih-guide/.venv/lib/python3.11/site-packages/infrahub_sdk/ctl/check.py", line 198, in run_checks
    result = await run_targeted_check(
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Mikhail/cloned/ih-guide/.venv/lib/python3.11/site-packages/infrahub_sdk/ctl/check.py", line 168, in run_targeted_check
    attribute = getattr(member.peer, identifier)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Mikhail/cloned/ih-guide/.venv/lib/python3.11/site-packages/infrahub_sdk/node/node.py", line 530, in __getattr__
    raise AttributeError(f"'{self.__class__.__name__}' object has no attribute '{name}'")
AttributeError: 'InfrahubNode' object has no attribute 'tag_name'
```

This check runs successfully in a PC check.

### Expected Behavior

A check will work the same when executed the same way via a PC or `infrahubctl check` command.

### Steps to Reproduce

1. Start fresh instance
2. Create gql
```
query TagsQuery($tag_name: String!) {
  BuiltinTag(name__value: $tag_name) {
    edges {
      node {
        name {
          value
        }
        description {
          value
        }
      }
    }
  }
}
```
3. Create Python check file
```python
import re

from infrahub_sdk.checks import InfrahubCheck


RE_TAG = re.compile(r"^color-[a-z]+")


class ColorTagsCheck(InfrahubCheck):
    query = "tags_check"

    def validate(self, data):
        for tag in data["BuiltinTag"]["edges"]:
            if not RE_TAG.match(tag["node"]["name"]["value"]):
                self.log_error(
                    message=f"Invalid tag name: {tag['node']['name']['value']}",
                    object_id=tag["node"]["name"]["value"],
                    object_type="BuiltinTag",
                )
```
4. Add a `.infrahub.yml`
```yaml
check_definitions:
  - name: "check_color_tags_name"
    class_name: "ColorTagsCheck"
    file_path: "checks/tags_check/tags_check.py"
    targets: "ColorTags"
    parameters:
      tag_name: "name__value"

queries:
  - name: tags_check
    file_path: "checks/tags_check/tags_check.gql"
```
4. Add two tag objects as part of the `ColorTags` group:
  1. `color-red`
  2. `blue`
5. Attempt to execute `infrahubctl check check_color_tags_name`

### Additional Information

_No response_

FragmentedPacket:
Fixes #445 
- Iterate over `parameters` defined in a `check_definition` and set the parameters into the GraphQL query properly

1. This code is still somewhat rudimentary and I think we may want to iterate over parameters and change `identifier` to a dictionary with the to be properly passed into the GraphQL to support multiple parameters.

### DIFF
--- a/infrahub_sdk/ctl/check.py
+++ b/infrahub_sdk/ctl/check.py
@@ -157,9 +157,7 @@ async def run_targeted_check(
         )
         check_summary.append(result)
     else:
-        targets = await client.get(
-            kind="CoreGroup", include=["members"], **{"name__value": check_module.definition.targets}
-        )
+        targets = await client.get(kind="CoreGroup", include=["members"], name__value=check_module.definition.targets)
         await targets.members.fetch()
         for member in targets.members.peers:
             check_parameter = {}
--- a/infrahub_sdk/ctl/check.py
+++ b/infrahub_sdk/ctl/check.py
@@ -137,12 +137,10 @@ async def run_targeted_check(
     variables: dict[str, str],
     branch: str | None = None,
 ) -> bool:
-    filters = {}
     identifier = None
     identifier_attribute = None
     # TODO: Does this support multiple parameters and we need make identifiers a dict to iterate over on L166-168?
     for param_key, param_value in check_module.definition.parameters.items():
-        filters[param_value] = check_module.definition.targets
         identifier = param_key
         identifier_attribute = param_value.split("__")[0]
 
@@ -159,7 +157,9 @@ async def run_targeted_check(
         )
         check_summary.append(result)
     else:
-        targets = await client.get(kind="CoreGroup", include=["members"], **filters)
+        targets = await client.get(
+            kind="CoreGroup", include=["members"], **{"name__value": check_module.definition.targets}
+        )
         await targets.members.fetch()
         for member in targets.members.peers:
             check_parameter = {}
--- a/infrahub_sdk/ctl/check.py
+++ b/infrahub_sdk/ctl/check.py
@@ -137,6 +137,7 @@ async def run_targeted_check(
     variables: dict[str, str],
     branch: str | None = None,
 ) -> bool:
+    filters = {"name__value": check_module.definition.targets}
     identifier = None
     identifier_attribute = None
     # TODO: Does this support multiple parameters and we need make identifiers a dict to iterate over on L166-168?
@@ -157,7 +158,7 @@ async def run_targeted_check(
         )
         check_summary.append(result)
     else:
-        targets = await client.get(kind="CoreGroup", include=["members"], name__value=check_module.definition.targets)
+        targets = await client.get(kind="CoreGroup", include=["members"], **filters)
         await targets.members.fetch()
         for member in targets.members.peers:
             check_parameter = {}
--- a/infrahub_sdk/ctl/check.py
+++ b/infrahub_sdk/ctl/check.py
@@ -138,14 +138,13 @@ async def run_targeted_check(
     branch: str | None = None,
 ) -> bool:
     filters = {}
-    param_value = list(check_module.definition.parameters.values())
-    if param_value:
-        filters[param_value[0]] = check_module.definition.targets
-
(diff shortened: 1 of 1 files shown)


infrahub_sdk/ctl/check.py
```diff
@@ -137,15 +137,13 @@ async def run_targeted_check(
...
-    param_key = list(check_module.definition.parameters.keys())
+    filters = {"name__value": check_module.definition.targets}
     identifier = None
-    if param_key:
-        identifier = param_key[0]
+    identifier_attribute = None
+    # TODO: Does this support multiple parameters and we need make identifiers a dict to iterate over on L166-168?
+    for param_key, param_value in check_module.definition.parameters.items():
+        identifier = param_key
+        identifier_attribute = param_value.split("__")[0]
```
ogenstad, line 146: I think this might be fine as it resolves a previous error. But I'm not sure these params really work today if you have more than one. This is something that we should address. It's also problematic with regards to a query that might look like this instead:

```graphql
query TagsQuery($tag_name: String!) {
  BuiltinTag(name__value: $tag_name) {
    edges {
      node {
        description {
          value
        }
      }
    }
  }
}
```

I.e. it's a valid check but the name is missing and as such will be reported as null.

Have you tried to run this same check and code within the pipeline in Infrahub as well? I think we might need to fix something there too.

FragmentedPacket: Hey @ogenstad, I ran the same check from the issue in the pipeline and it succeeds with no issue.

If I update the GraphQL query to above, the check fails but that is due to the code within the check.

FragmentedPacket:
Closing due to a larger effort required: https://github.com/opsmill/infrahub/issues/7188

PR #448, opened 2025-06-18
Fixes the import path for the Attribute class

wvandeun:
In #409 we have split up the `node.py` file into multiple files.
This created an issue with the import path of the Attribute class.


wvandeun:
Related to https://github.com/opsmill/infrahub-ansible/pull/228

ogenstad (approved):

PR #456, opened 2025-06-27
Add setting typename, display_label, and kind if it exists when calling _init_relationships

### LINKED ISSUE #455: bug: Error when accessing RelatedNode on a created device missing fields to query it
### Component

Python SDK

### Infrahub SDK version

1.13.2

### Current Behavior

We get the following error when attempting to access the peer of a `RelatedNode` that was used in the `create` function of an object and we attempt to get the peer later on.
```
ValueError: Node must have at least one identifier (ID or HFID) to query it.
```

In this case, we create a device using a model that we fetched on a different node, and then when we attempt to get the peer of `device.model` we encounter the error above.

### Expected Behavior

This was a breaking change that was introduced recently, but the typename was not being set when we had it on the related node.

### Steps to Reproduce

Example above.

### Additional Information

_No response_

FragmentedPacket:
Fixes #455 

- We set the `__typename` when relationships are already a RelatedNode as we were omitting it previously so it could not do the lookup of the object in the store.
- also sets `kind` and `display_label` for the peer if it exists on the input `RelatedNode`

### DIFF
--- a/changelog/455.fixed.md
+++ b/changelog/455.fixed.md
@@ -1,0 +1,1 @@
+Update InfrahubNode creation to include __typename, display_label, and kind from a RelatedNode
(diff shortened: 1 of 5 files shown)


dgarros (approved):
Please can either of you add a changelog before merging this PR

Issue #462, opened 2025-07-03
bug: Can't load folder that contains other type of file


BaptisteGi:
### Component

infrahubctl

### Infrahub SDK version

1.13.3

### Current Behavior

Assuming I have the following hierarchy:

```bash
└── vlan
    ├── README.md
    └── vlan.yml
```

Essentially a folder that contains my schema yaml file and a small md piece of documentation.

If I try to load this schema pointing to the folder I have the following output:

```bash
infrahubctl schema load vlan        

vlan/README.md does not exist!
```

### Expected Behavior

I expect the command to ignore `.md` files (and everything which is not `.yml`)

### Steps to Reproduce

- Create a folder that contains both a `.yml` and a `.md` file
- Try to schema load that folder

### Additional Information

https://github.com/opsmill/infrahub-sdk-python/blob/02b4f12f3dd686d26e2dd69fc1943582d4f6d742/infrahub_sdk/yaml.py#L121-L134

- here having .json in a class that is called YAML tickles something
-  I believe we need to add an extra statement to differentiate `user passes an invalid path` from `that path contains a file which is not a yaml file`

BaptisteGi:
Fixed in #463 and #472

PR #463, opened 2025-07-03
Fix `load_from_disk` method to support folder with multiple file extensions

BaptisteGi:
This pull request refines the `load_from_disk` method in `infrahub_sdk/yaml.py` to make sure we can load folders that contain various file types.

Relates to #462

### DIFF
--- a/infrahub_sdk/yaml.py
+++ b/infrahub_sdk/yaml.py
@@ -120,16 +120,23 @@ class YamlFile(LocalFile):
     @classmethod
     def load_from_disk(cls, paths: list[Path]) -> list[Self]:
         yaml_files: list[Self] = []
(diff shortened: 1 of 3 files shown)


infrahub_sdk/yaml.py
```diff
@@ -120,16 +120,23 @@ def load_file_from_disk(cls, path: Path) -> list[Self]:
     @classmethod
     def load_from_disk(cls, paths: list[Path]) -> list[Self]:
         yaml_files: list[Self] = []
+        file_extensions = {".yaml", ".yml", ".json"}  # FIXME: .json is not a YAML file, should be removed
```
BaptisteGi, line 123: @LucasG0 do you know why we have .json here? too afraid about side effect of removing it from here

dgarros (approved):
Please can you add a changelog for the release note

PR #467, opened 2025-07-11
Infrahub repostory init

minitriga:


BaptisteGi (approved):
Just ran the command, everything works perfectly. Nice job! 👍 

Few details:
- Some could argue that if you have generator/transform you'll have by consequence queries. Don't know if we want to override the query flag based on generator/transform one.
- In my opinion, this should be the default way for a developer to kick the tires. Then it should be included in the infrahub getting started doc https://docs.infrahub.app/getting-started/next-steps.

minitriga:
@BaptisteGi Templates fixed queries will now be included by default if generators or transforms are used. 

https://github.com/opsmill/infrahub-template/pull/4

PR #468, opened 2025-07-14
Fix sync parallel filters

BeArchiTek:
When using the InfrahubClientSync.filters() with `parallel=True`, the pagination was unstable, with some delta returned now and then.

When dumping the queries, every time this issue happens, it seems like we are doing 2 times the 2nd page and not the first one. Forcing offset=0 fixes it.

I am not sure why this issue was not present with the Async Client, considering both of them are using the same `generate_query_data_init()` ...

### DIFF
--- a/infrahub_sdk/client.py
+++ b/infrahub_sdk/client.py
@@ -830,7 +830,7 @@ class InfrahubClient(BaseClient):
 
             for page_number in range(1, total_pages + 1):
                 page_offset = (page_number - 1) * pagination_size
-                batch_process.add(task=process_page, page_offset=page_offset, page_number=page_number)
+                batch_process.add(task=process_page, node=node, page_offset=page_offset, page_number=page_number)
 
             async for _, response in batch_process.execute():
                 nodes.extend(response[1]["nodes"])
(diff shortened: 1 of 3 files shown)


infrahub_sdk/client.py
```diff
@@ -1995,7 +1996,7 @@ def process_batch() -> tuple[list[InfrahubNodeSync], list[InfrahubNodeSync]]:
 
             for page_number in range(1, total_pages + 1):
                 page_offset = (page_number - 1) * pagination_size
-                batch_process.add(task=process_page, node=node, page_offset=page_offset, page_number=page_number)
+                batch_process.add(task=process_page, page_offset=page_offset, page_number=page_number)
```
ogenstad: The removal of the node parameter here as well as in the async version is that something that's related to this change?

Looking at your change it doesn't seem like `node` provided any value previously. However we're still setting node with:

```python
node = InfrahubNodeSync(client=self, schema=schema, branch=branch)
```

I wonder if it makes sense to keep this, it's also a bit problematic as we're shadowing the `node` variable a bit further down in the method:

```python
        if populate_store:
            for node in nodes:
                if node.id:
                    self.store.set(node=node)
            related_nodes = list(set(related_nodes))
            for node in related_nodes:
                if node.id:
                    self.store.set(node=node)
```

BeArchiTek: I could let `node` or do it in another PR, while debugging, I saw that we have node set twice out of 4 times when using the function, so from my point of view the function signature was not matching the usage.

ogenstad: We can perhaps create a follow up issue to remove the `node =` parts in these two methods.

ogenstad (approved):
Can you also add a newsfragment to indicate what was fixed?

PR #471, opened 2025-07-15
Allow more versions of Typer

dgarros:
Currently Typer is pinned to `"^0.12.3"` which is a bit agressive
This is especially problematic if we want to install the SDK along with another package that is using Typer

This PR relax the constraint to allow any version of Typer from 0.12.3 to 0.16.x


dgarros:
Will reopen it

PR #478, opened 2025-07-22
respect ordering of files when loading

ajtmccarty:
always sorting the files alphabetically will cause a problem if a file `a_objects.yml` is dependent on `z_objects.yml`
this update is to only sort files alphabetically if they are in a directory, in which case I think we can assume that order does not matter.
I'm not exactly sure why we need to order files alphabetically at all actually, so maybe the added line is unecessary

## Summary by CodeRabbit

* **Refactor**
  * Updated the order in which files are processed when loading from directories, ensuring subdirectory contents are sorted by filename while leaving the initial input order unchanged. This may affect the sequence in which files are loaded from disk.


ajtmccarty:
please feel free to update and/or merge without me if necessary

dgarros (approved):
Good catch

PR #481, opened 2025-07-23
ci: add api doc generation

a-delannoy:
Provide the basics for automatic Infrahub SDK API documentation generation:

* Add the `mdxify` package to the dependencies
* Generate a first version of the Infrahub SDK API documentation
* Fix docusaurus interpretation of mdx files with Icon element in it
* Update docusaurus sidebar configuration to include the SDK API documentation reference
* Add an invoke task to regenerate the documentation: `poetry run invoke generate-sdk-api-docs`
* Add a CI rule for checking the SDK API obsolescence and fail if it is not up to date

## Summary by CodeRabbit

* **New Features**
  * Added automated generation of Python SDK API documentation from code docstrings, including comprehensive reference pages for client and node modules.
  * Introduced a new documentation sidebar section for Python SDK API reference.
  * Enabled use of the Iconify Icon component in documentation MDX files.

* **Chores**
  * Added a continuous integration job to check for outdated SDK API documentation.
  * Added development dependencies for API doc generation and linting.

* **Documentation**
  * Improved and expanded Python SDK documentation with detailed API references.
  * Fixed minor typos and improved docstring clarity in several modules.


docs/src/theme/MDXComponents.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
```
dgarros, line 1: @petercrocker do we need to add this component to Infrahub-docs as well ?

dgarros (approved):
Looks great
Many things to improve but at least we have the foundation 👍🏼

dgarros:
please don't forget to add a fragment for the changelog

a-delannoy:
The only failing CI check is about documentation style using Vale. The thing is, it does not like some of the tokens that are used in Google style docstrings (like `Args` for example), among other more technical spellings.

I am not sure if it is relevant to have it run on doc files generated from the code. I would suggest excluding the newly generated doc path from Vale surveillance. Otherwise, we would have to add a lot of spelling exceptions over time that I don't find very relevant.
What do you think about this ?

dgarros:
> I would suggest excluding the newly generated doc path from Vale surveillance.
Agreed

a-delannoy:
> > I would suggest excluding the newly generated doc path from Vale surveillance.
> > Agreed

Ok, that's done !

a-delannoy:
> please don't forget to add a fragment for the changelog

Done as well !

.github/workflows/ci.yml
```diff
@@ -180,6 +178,50 @@ jobs:
...
+          node-version: 20
+          cache: "npm"
+          cache-dependency-path: "**/package-lock.json"
+      - name: Install markdown linter
+        run: npm install -g markdownlint-cli2
+      - name: "Generate SDK API documentation"
+        run: ${{ env.DOCS_COMMAND }}
+      - name: "Check if SDK API documentation needs to be refreshed"
+        run: |
+          git diff --quiet ${SDK_API_DOCS_DIR}
```
ogenstad, line 223: I'd add this as an invoke command too so that it's consistent with how we do it in other repos.

docs/docs/python-sdk/sdk_ref/infrahub_sdk/analyzer.mdx
```diff
@@ -0,0 +1,68 @@
+---
+title: analyzer
+sidebarTitle: analyzer
+---
+
+# `infrahub_sdk.analyzer`
+
+## Classes
+
+### `GraphQLQueryVariable` <sup><a href="https://github.com/opsmill/infrahub-sdk-python/blob/main/infrahub_sdk/analyzer.py#L20" target="_blank"><Icon icon="github" style="width: 14px; height: 14px;" /></a></sup>
```
ogenstad: There are a lot of these links to the `main` branch along with a line number. Looking at the published docs https://5390d612.infrahub-sdk-python.pages.dev/python-sdk/sdk_ref/infrahub_sdk/analyzer#graphqlqueryvariable- I can't actually see the these links on the page. But they are available in the source. I think we'd need to get rid of them to avoid conflicts as well as having a link that points to an incorrect branch.

If possible it could make sense to have this point to the current version at build time so that we have a link to a tagged release instead of a branch where the line numbers as well as overall structure of the file might change.

ogenstad (commented):
Nice work! I'd suggest that we start with a smaller scale and document the InfrahubClient classes as well as InfrahubNode and then curate the classes that we want to, later it could perhaps be based on file locations.

I think that we have too many public methods exposed today which we probably don't want to have documented like this.

Some of them could be confusing for a user such as: https://5390d612.infrahub-sdk-python.pages.dev/python-sdk/sdk_ref/infrahub_sdk/ctl/branch The functions under the CTL shouldn't really be used by end users directly but consumed by `infrahubctl`. I'm afraid that AI agents will make assumptions based of function names like this and generate faulty code as a result. Alternatively that "internal" functions are used in ways that might break when we change the code in future SDK versions. Also with things like https://5390d612.infrahub-sdk-python.pages.dev/python-sdk/sdk_ref/infrahub_sdk/constants that doesn't provide as much value as other pages it just adds noise.

Some of the output might be a bit confusing for humans when looking at for example the client.get method: https://5390d612.infrahub-sdk-python.pages.dev/python-sdk/sdk_ref/infrahub_sdk/client#get- most of these listings would be for `@overload` methods but it's not quite clear by looking at the docs. If possible it would be nice if we could mark the ones that are overload methods.

docs/src/theme/MDXComponents.js
petercrocker: I think it's already there: https://github.com/opsmill/infrahub-docs/blob/894076746c6fa744dd1eca674d6eb25ccb1d95f0/docs/package.json#L27

dgarros: Sorry I commented the wrong line in this file
I meant `"@iconify/react"` and the Icon component

.github/workflows/ci.yml
a-delannoy: Can you elaborate a bit about which part here you'd add as an invoke command ?
For now, I added an invoke command to generate the API docs.

docs/docs/python-sdk/sdk_ref/infrahub_sdk/analyzer.mdx
a-delannoy: I guess it is about docusaurus configuration because the [FastMCP doc](https://gofastmcp.com/python-sdk/fastmcp-exceptions) partake of those links while they don't appear in our documentation.

Mdxify lib provides us with a CLI arg to properly point at the desired git branch (might also work with any git ref but not sure about this) in generated code links. Do you want to leverage this ?

a-delannoy:
> Nice work! I'd suggest that we start with a smaller scale and document the InfrahubClient classes as well as InfrahubNode and then curate the classes that we want to, later it could perhaps be based on file locations.
> 
> I think that we have too many public methods exposed today which we probably don't want to have documented like this.
> 
> Some of them could be confusing for a user such as: https://5390d612.infrahub-sdk-python.pages.dev/python-sdk/sdk_ref/infrahub_sdk/ctl/branch The functions under the CTL shouldn't really be used by end users directly but consumed by `infrahubctl`. I'm afraid that AI agents will make assumptions based of function names like this and generate faulty code as a result. Alternatively that "internal" functions are used in ways that might break when we change the code in future SDK versions. Also with things like https://5390d612.infrahub-sdk-python.pages.dev/python-sdk/sdk_ref/infrahub_sdk/constants that doesn't provide as much value as other pages it just adds noise.
> 
> Some of the output might be a bit confusing for humans when looking at for example the client.get method: https://5390d612.infrahub-sdk-python.pages.dev/python-sdk/sdk_ref/infrahub_sdk/client#get- most of these listings would be for `@overload` methods but it's not quite clear by looking at the docs. If possible it would be nice if we could mark the ones that are overload methods.

I agree with this. Do you have a list of modules to include/exclude ?

a-delannoy:
Also, regarding decorators, this is a thing we already started discussing with Damien. We opened an issue on mdxify to start a discussion around this.

.github/workflows/ci.yml
ogenstad: I mean that we can run `poetry run invoke generate-sdk-api-docs` to generate the docs but then have another command to validate. It could simplify to also have a shortcut for the git diff like we have here:
https://github.com/opsmill/infrahub/blob/infrahub-v1.3.4/tasks/backend.py#L176

This part isn't critical though just a suggestion.

docs/docs/python-sdk/sdk_ref/infrahub_sdk/analyzer.mdx
ogenstad: I'd say there are two issues the first is that the links are currently broken due to the `main` branch reference. The second is that as I can't even see the links on the website I don't think they fill a need. It might be because the icon for Github is missing so there's nothing to click on?

ogenstad:
> I agree with this. Do you have a list of modules to include/exclude ?

For the first iteration perhaps InfrahubClient, InfrahubClientSync, InfrahubNode and InfrahubNodeSync. If we have an easy way to just update it later that should be enough to add more as a follow up step.

a-delannoy:
> > I agree with this. Do you have a list of modules to include/exclude ?
> 
> For the first iteration perhaps InfrahubClient, InfrahubClientSync, InfrahubNode and InfrahubNodeSync. If we have an easy way to just update it later that should be enough to add more as a follow up step.

Just a precision, mdxify works on a module basis. So I found the proper modules, but the full generation is generated for everything that's inside it.

polmichel:
This PR has been taken over: https://github.com/opsmill/infrahub-sdk-python/pull/822.

PR #487, opened 2025-07-29
Rework `get_flat_value` to fix node relationships

gmazoyer:
This change moves the `get_flat_value` function to methods of the `InfrahubNode` and `InfrahubNodeSync` classes. This allows to handle traversing relationships of cardinality one using a flat notation in addition to attributes.

The `extract` method has also been moved as it needs to be `async` for regular nodes now.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of value lookups using flat notation (e.g., `foo__bar__value`) for relationships with cardinality one, ensuring correct resolution of nested attributes.

* **New Features**
  * Introduced enhanced methods for retrieving and extracting nested values using flat key notation in both asynchronous and synchronous node interfaces.

* **Tests**
  * Updated and expanded tests to separately verify flat value retrieval and extraction, including error handling for invalid lookups.
  * Removed outdated tests related to the previous utility function for flat value extraction.

### DIFF
--- a/infrahub_sdk/node/node.py
+++ b/infrahub_sdk/node/node.py
@@ -1053,7 +1053,7 @@ class InfrahubNode(InfrahubNodeBase):
 
         if rel.cardinality != RelationshipCardinality.ONE:
             raise ValueError(
-                f"Can only look up flat value for relationships of cardinality {RelationshipCardinality.ONE}"
+                f"Can only look up flat value for relationships of cardinality {RelationshipCardinality.ONE.value}"
             )
 
         related_node: RelatedNode = getattr(self, first)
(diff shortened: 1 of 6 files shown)


infrahub_sdk/node/node.py
```diff
@@ -1036,6 +1028,46 @@ def _get_relationship_one(self, name: str) -> RelatedNode:
...
+            return attr
+
+        try:
+            rel = self._schema.get_relationship(name=first)
+        except ValueError as exc:
+            raise ValueError(f"No attribute or relationship named '{first}' for '{self._schema.kind}'") from exc
+
+        if rel.cardinality != RelationshipCardinality.ONE:
+            raise ValueError(
+                f"Unable to lookup flat value for relationship of cardinality {RelationshipCardinality.MANY}"
```
ajtmccarty: it might be more clear to use the positive case instead of the negative here. something like `Can only look up flat value for relationships of cardinality one`

infrahub_sdk/node/node.py
```diff
@@ -1036,6 +1028,46 @@ def _get_relationship_one(self, name: str) -> RelatedNode:
 
         raise ResourceNotDefinedError(message=f"The node doesn't have a cardinality=one relationship for {name}")
 
+    async def get_flat_value(self, key: str, separator: str = "__") -> Any:
+        """Query recursively an value defined in a flat notation (string), on a hierarchy of objects
```
ajtmccarty: `an value` -> `a value`

ajtmccarty (approved):

PR #494, opened 2025-08-12
handle None comparison in deep_merge_dict

ajtmccarty:
part of fixing https://github.com/opsmill/infrahub/issues/7022

handles the situation when a `None` is compared to a non-`None`. for example `{"name": None}` and `{"name": {"value": None}}` should be merged as `{"name": {"value": None}}`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deep-merge behavior when combining values with None, ensuring non-None values are preserved and equal values are respected.
  * Clarified conflict handling for mismatched non-None values to prevent ambiguous merges.
  * Ensured list merging remains consistent while avoiding duplicates.

* **Tests**
  * Added unit tests covering None-vs-non-None merge scenarios to validate expected outcomes and prevent regressions.

### DIFF
--- a/infrahub_sdk/utils.py
+++ b/infrahub_sdk/utils.py
@@ -142,14 +142,18 @@ def deep_merge_dict(dicta: dict, dictb: dict, path: list | None = None) -> dict:
     if path is None:
         path = []
     for key in dictb:
+        b_val = dictb[key]
         if key in dicta:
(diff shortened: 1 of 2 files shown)


ogenstad (approved):
LGTM. I'm wondering if there's a point to have `deep_merge_dict()` in the SDK, to me that feels wrong as it's only used within Infrahub. Not really related to this PR though.

gmazoyer (approved):

PR #508, opened 2025-08-25
IHS-154 Deprecate using `raise_for_error = False`

### LINKED ISSUE #493: bug: Using client `execute_graphql` with `raise_for_error = False` still raises
### Component

Python SDK

### Infrahub SDK version

1.13.5

### Current Behavior

When using the `execute_graphql` method, of both the sync and async clients, with the keyword arg `raise_for_error` set to `False`, if the GraphQL query produces an error, that error will still be raised as a `GraphQLError` exception.

### Expected Behavior

The `GraphQLError` exception should not be raised, and the return value of the method should contain the error.

### Steps to Reproduce

* Build a SDK client (sync or async)
* Use a call like `await client.execute_graphql(query=some_query, variables=some_vars, raise_for_error=False)`

### Additional Information

_No response_

gmazoyer:
This is a simple change that adds a warning when using a client's method with the named parameter `raise_for_error` set. This parameter must be removed later.

## Summary by CodeRabbit

- Documentation
  - Documented deprecation of explicitly setting raise_for_error (now optional); guidance to use try/except for custom HTTP error handling and note that GraphQL errors still raise regardless.

- Chores
  - Emit deprecation warnings when raise_for_error is provided (non-null) in both sync and async client calls.
  - Made raise_for_error optional: unset or true → HTTP errors raised; false → HTTP errors not raised.

### DIFF
--- a/infrahub_sdk/client.py
+++ b/infrahub_sdk/client.py
@@ -1649,14 +1649,17 @@ class InfrahubClientSync(BaseClient):
             branch_name (str, optional): Name of the branch on which the query will be executed. Defaults to None.
             at (str, optional): Time when the query should be executed. Defaults to None.
             timeout (int, optional): Timeout in second for the query. Defaults to None.
-            raise_for_error (bool, optional): Deprecated, flag to indicate that we need to raise an exception if the response has some errors.
-            Defaults to True.
+            raise_for_error (bool | None, optional): Deprecated. Controls only HTTP status handling.
+                - None (default) or True: HTTP errors raise via `resp.raise_for_status()`.
+                - False: HTTP errors are not automatically raised.
+              GraphQL errors always raise `GraphQLError`.
+            Defaults to None.
+
         Raises:
-            GraphQLError: When an error occurs during the execution of the GraphQL query or mutation.
+            GraphQLError: When the GraphQL response contains errors.
 
         Returns:
-            dict: The result of the GraphQL query or mutation.
-        """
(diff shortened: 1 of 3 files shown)


ogenstad:
We'd want a news fragment for this as well to indicate the bug that was addressed.

infrahub_sdk/client.py
```diff
@@ -893,13 +894,20 @@ async def execute_graphql(
...
 
         Returns:
             _type_: _description_
         """
+        if not raise_for_error:
+            warnings.warn(
+                "Setting `raise_for_error=False` is deprecated, use `try/except` to handle errors.",
+                DeprecationWarning,
+                stacklevel=1,
+            )
```
ogenstad: I'm wondering if we should instead change the option to be `raise_for_error: bool | None` and then have:

```python
        if raise_for_error is not None:
            warnings.warn(
                "Setting `raise_for_error` is deprecated, use `try/except` to handle errors.",
                DeprecationWarning,
                stacklevel=1,
            )
```

I.e. we also want to deprecate setting raise_for_error=True. Does that make sense?

ogenstad:
When looking through the SDK for other reasons I realise we have a number of methods that's using this raise_for_error for example .allocate_next_ip_prefix(), I think we need to deprecate those as well and avoid calling this deprecated parameter from those other methods as well.

ogenstad (approved):
Thank you for the updates!

PR #513, opened 2025-08-26
Fix filters `include` parameter

### LINKED ISSUE #6976: bug: Artifact generation fails when device is removed from the artifact definition group
h3.  Component

[| #component]

API Server / GraphQL

h3.  Infrahub version

[| #infrahub-version]

1.3.2/3/4

h3.  Current Behavior

[| #current-behavior]

Artifact generation failure within a Proposed Change if a device is removed from the artifact definition's target group.

[!https://github.com/user-attachments/assets/7218bdc0-dcb6-4e62-aeed-95e4d000673f|width=1518,height=256!| https://github.com/user-attachments/assets/7218bdc0-dcb6-4e62-aeed-95e4d000673f]

h3.  Expected Behavior

[| #expected-behavior]

Artifact generation to succeed and not run against the device that was removed from the target group

h3.  Steps to Reproduce

[| #steps-to-reproduce]

# Fresh instance of Infrahub on 1.3.4
# invoke load-schema load-data ([https://github.com/opsmill/infrahub-demo-dc-fabric| https://github.com/opsmill/infrahub-demo-dc-fabric])
# poetry run infrahubctl run bootstrap/generate_topology.py topology=fra05-pod1
# Add repository to Infrahub (artifacts get generated)
# Create new branch
# Remove `fra05-pod1-leaf1` from `arista_devices` group
# Create a PC with the artifact still present
# it should fail on artifact generation

h3.  Additional Information_No response_

[| #additional-information_no-response_]

LucasG0:
Help to fix https://github.com/opsmill/infrahub/issues/6976

Fix `client.filters(include=[...])` that was not fetching relationships

## Summary by CodeRabbit

- New Features
  - Specifying include now loads selected relationships without enabling prefetch in both async and sync clients.
  - Include-driven relationship loading expands nested relationship fetches appropriately across modes.

- Bug Fixes
  - Ensures peers are expanded when relationships are explicitly included, preventing empty/missing data when prefetch is off.

- Tests
  - Added integration test validating include behavior for owner and tags.
  - Extended unit tests and fixtures to cover both async and sync clients and schema caching.

### DIFF
--- a/infrahub_sdk/client.py
+++ b/infrahub_sdk/client.py
@@ -539,7 +539,6 @@ class InfrahubClient(BaseClient):
         response: dict[str, Any],
         schema_kind: str,
         branch: str,
-        prefetch_relationships: bool,
         timeout: int | None = None,
     ) -> ProcessRelationsNode:
         """Processes InfrahubNode and their Relationships from the GraphQL query response.
@@ -548,7 +547,6 @@ class InfrahubClient(BaseClient):
             response (dict[str, Any]): The response from the GraphQL query.
             schema_kind (str): The kind of schema being queried.
             branch (str): The branch name.
-            prefetch_relationships (bool): Flag to indicate whether to prefetch relationship data.
             timeout (int, optional): Overrides default timeout used when querying the graphql API. Specified in seconds.
 
         Returns:
@@ -564,10 +562,9 @@ class InfrahubClient(BaseClient):
             node = await InfrahubNode.from_graphql(client=self, branch=branch, data=item, timeout=timeout)
             nodes.append(node)
 
-            if prefetch_relationships:
-                await node._process_relationships(
-                    node_data=item, branch=branch, related_nodes=related_nodes, timeout=timeout
-                )
+            await node._process_relationships(
+                node_data=item, branch=branch, related_nodes=related_nodes, timeout=timeout
+            )
 
         return ProcessRelationsNode(nodes=nodes, related_nodes=related_nodes)
 
@@ -814,7 +811,6 @@ class InfrahubClient(BaseClient):
                 response=response,
                 schema_kind=schema.kind,
                 branch=branch,
-                prefetch_relationships=prefetch_relationships,
                 timeout=timeout,
             )
             return response, process_result
--- a/infrahub_sdk/client.py
+++ b/infrahub_sdk/client.py
@@ -558,9 +558,6 @@ class InfrahubClient(BaseClient):
                 - 'related_nodes': A list of InfrahubNode objects representing the related nodes
         """
 
-        # Ideally, include and relationships wouldn't be parameters of this method, they should only
-        # be used to build the request for the server, and this method would build node according to the response.
-
         nodes: list[InfrahubNode] = []
         related_nodes: list[InfrahubNode] = []
 
@@ -568,7 +565,7 @@ class InfrahubClient(BaseClient):
             node = await InfrahubNode.from_graphql(client=self, branch=branch, data=item, timeout=timeout)
             nodes.append(node)
 
-            if prefetch_relationships or include is not None:
+            if prefetch_relationships or (include and any(rel in include for rel in node._relationships)):
                 await node._process_relationships(
                     node_data=item,
                     branch=branch,
@@ -1864,7 +1861,7 @@ class InfrahubClientSync(BaseClient):
             node = InfrahubNodeSync.from_graphql(client=self, branch=branch, data=item, timeout=timeout)
             nodes.append(node)
 
-            if prefetch_relationships or include is not None:
+            if prefetch_relationships or (include and any(rel in include for rel in node._relationships)):
                 node._process_relationships(node_data=item, branch=branch, related_nodes=related_nodes, timeout=timeout)
 
         return ProcessRelationsNodeSync(nodes=nodes, related_nodes=related_nodes)
--- a/infrahub_sdk/client.py
+++ b/infrahub_sdk/client.py
@@ -539,6 +539,8 @@ class InfrahubClient(BaseClient):
         response: dict[str, Any],
         schema_kind: str,
         branch: str,
+        prefetch_relationships: bool,
+        include: list[str] | None,
         timeout: int | None = None,
     ) -> ProcessRelationsNode:
         """Processes InfrahubNode and their Relationships from the GraphQL query response.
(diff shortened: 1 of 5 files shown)


tests/unit/sdk/conftest.py
```diff
@@ -2641,3 +2646,20 @@ async def mock_query_tasks_05(httpx_mock: HTTPXMock) -> HTTPXMock:
         is_reusable=True,
     )
     return httpx_mock
+
+
+async def set_builtin_tag_schema_cache(client) -> None:
```
ajtmccarty: seems like this file in mainly fixtures, so maybe this function should move to `test_node.py`

infrahub_sdk/client.py
```diff
@@ -557,16 +558,22 @@ async def _process_nodes_and_relationships(
                 - 'related_nodes': A list of InfrahubNode objects representing the related nodes
         """
 
+        # Ideally, include and relationships wouldn't be parameters of this method, they should only
+        # be used to build the request for the server, and this method would build node according to the response.
```
ajtmccarty: I agree with this comment and I think you can do this differently and then remove the comment

what about instead of passing `include` in here, you update the `prefetch_relationships` parameter to be a `do_process_relationships` Boolean flag and then update the two places where `_process_nodes_and_relationships` is called to set `do_process_relationships` based on `prefetch_relationships` and if `include` is not empty (or maybe if `include` actually includes the name of a relationship)

LucasG0: This is about the fact that we should rely on these parameters only to fetch the data from the server and not while building the nodes client side.

The current logic is:

```
if prefetch_relationships / include has a rel_name:
   fetch_rel_from_server

if prefetch_relationships / include has a rel_name:
    process_relationships() # build peers nodes
```

The comment means that we should get rid of the second if condition, ie, while building the nodes we should just build it ouf of the payload that we previously fetched. Does this make sense? I tried to do so but it led to some complicated bugs that should be treated out of this PR if we want to clean that.

I still made a minor change following up your comment so that include checks on whether it contains a relationship name.

Using a `do_prefetch_relationships` as you suggested might not fit well here because we have access to node relationships once it is built inside the `_process_nodes_and_relationships` method, and we need relationships names to check whether `include` contains some.

ajtmccarty: I was just suggesting making the below small changes. I didn't realize that this comment was related to existing functionality. This sounds like something that deserves its own issue in GH/Jira instead of a comment that will be ignored / forgotten
```
diff --git a/infrahub_sdk/client.py b/infrahub_sdk/client.py
index f9a380b..cae970a 100644
--- a/infrahub_sdk/client.py
+++ b/infrahub_sdk/client.py
@@ -539,8 +539,7 @@ class InfrahubClient(BaseClient):
         response: dict[str, Any],
         schema_kind: str,
         branch: str,
-        prefetch_relationships: bool,
-        include: list[str] | None,
+        do_process_relationships: bool,
         timeout: int | None = None,
     ) -> ProcessRelationsNode:
         """Processes InfrahubNode and their Relationships from the GraphQL query response.
@@ -549,7 +548,7 @@ class InfrahubClient(BaseClient):
             response (dict[str, Any]): The response from the GraphQL query.
             schema_kind (str): The kind of schema being queried.
             branch (str): The branch name.
-            prefetch_relationships (bool): Flag to indicate whether to prefetch relationship data.
+            do_process_relationships (bool): Flag to indicate whether to fetch relationship data.
             timeout (int, optional): Overrides default timeout used when querying the graphql API. Specified in seconds.
 
         Returns:
@@ -568,7 +567,7 @@ class InfrahubClient(BaseClient):
             node = await InfrahubNode.from_graphql(client=self, branch=branch, data=item, timeout=timeout)
             nodes.append(node)
 
-            if prefetch_relationships or include is not None:
+            if do_process_relationships:
                 await node._process_relationships(
                     node_data=item,
                     branch=branch,
@@ -821,9 +820,8 @@ class InfrahubClient(BaseClient):
                 response=response,
                 schema_kind=schema.kind,
                 branch=branch,
-                prefetch_relationships=prefetch_relationships,
+                do_process_relationships=prefetch_relationships or include is not None,
                 timeout=timeout,
-                include=include,
             )
             return response, process_result
```

LucasG0: I understood that this was the change you suggested, and I think we cannot do that because 
`do_process_relationships=prefetch_relationships or include is not None` is missing checking relationships names as you nicely pointed out, and we need the node being built to do that, cf my previous comment

> Using a do_prefetch_relationships as you suggested might not fit well here because we have access to node relationships once it is built inside the _process_nodes_and_relationships method, and we need relationships names to check whether include contains some.

LucasG0: I created IHS-161

infrahub_sdk/client.py
```diff
@@ -557,16 +558,22 @@ async def _process_nodes_and_relationships(
...
+
         nodes: list[InfrahubNode] = []
         related_nodes: list[InfrahubNode] = []
 
         for item in response.get(schema_kind, {}).get("edges", []):
             node = await InfrahubNode.from_graphql(client=self, branch=branch, data=item, timeout=timeout)
             nodes.append(node)
 
-            if prefetch_relationships:
+            if prefetch_relationships or include is not None:
```
ogenstad: I wonder if we should have an extra check here. I.e. if "include" just consists of a list of attribute names we don't really need to run the code for process_relationships?

infrahub_sdk/client.py
```diff
@@ -1855,7 +1864,7 @@ def _process_nodes_and_relationships(
             node = InfrahubNodeSync.from_graphql(client=self, branch=branch, data=item, timeout=timeout)
             nodes.append(node)
 
-            if prefetch_relationships:
+            if prefetch_relationships or include is not None:
```
ogenstad: Same comment as above.

infrahub_sdk/node/node.py
```diff
@@ -742,12 +742,11 @@ async def generate_query_data_node(
                 continue
 
             peer_data: dict[str, Any] = {}
-            if rel_schema and prefetch_relationships:
+            should_fetch_relationship = prefetch_relationships or (include is not None and rel_name in include)
```
ogenstad, line 745: It was something like this I was thinking about in the comments above.

ogenstad (approved):
LGTM, just added some comments about some potential additional restrictions before trying to fetch the relationships (I'm not completely sure it's applicable, i.f. if we'd end up not querying for those relationships regardless my comments could be ignored.

infrahub_sdk/node/node.py
LucasG0: I had actually addressed these changes following Aaron's comment but didn't push them

ajtmccarty (approved):

PR #518, opened 2025-08-28
Respect default branch for client.query_gql_query()

### LINKED ISSUE #236: bug: Function `query_gql_query` doesn't take into account client config
h3.  Component

[| #component]

Python SDK

h3.  Infrahub SDK version

[| #infrahub-sdk-version]

1.1.0

h3.  Current Behavior

[| #current-behavior]

When running `query_gql_query` function I need to explicitly specify the branch otherwise it runs the query on main. This is particularly impacting generator/transform part where the client is configured to run on a branch.

<pre>    fabric_data = await self.client.query_gql_query(
        name="fabric-informations",
        variables={"fabric_name": device_dict["fabric"]["node"]["name"]["value"]},
        branch_name=self.client.config.default_branch,
    )</pre>

h3.  Expected Behavior

[| #expected-behavior]

I expect infrahub to run the graphql query toward the branch setup in `self.client`

h3.  Steps to Reproduce

[| #steps-to-reproduce]

Create a Graphql query and run it from the SDK

h3.  Additional Information

[| #additional-information]

Maybe other client functions might be impacted ...

ogenstad:
Fixes #236.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured all GraphQL requests (async and sync) always include a branch parameter, defaulting to the configured default branch when none is provided. This delivers consistent branch-scoped behavior across requests and prevents unintended cross-branch queries; other request parameters remain unchanged.

* **Documentation**
  * Added changelog entry noting respect for the default branch in relevant client operations.

### DIFF
--- a/changelog/236.fixed.md
+++ b/changelog/236.fixed.md
@@ -1,0 +1,1 @@
+Respect default branch for client.query_gql_query() and client.set_context_properties()
--- a/infrahub_sdk/client.py
+++ b/infrahub_sdk/client.py
@@ -209,7 +209,7 @@ class BaseClient:
             delete_unused_nodes=delete_unused_nodes,
             group_type=group_type,
             group_params=group_params,
-            branch=branch,
+            branch=branch or self.default_branch,
         )
 
     def _graphql_url(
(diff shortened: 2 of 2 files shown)


ajtmccarty:
it looks like at least some of the client-related code expects the `default_branch` property of the client to actually be the default branch name (`main` in most cases) and not "the branch to use for this generator/transform/whatever"

for example
in `BaseClient.__init__` 
```
        self.default_branch = self.config.default_infrahub_branch
```

in `InfrahubClient._initialize`
```
    self.store = NodeStore(default_branch=self.default_branch)
```
where `NodeStore` seems to expect the `default_branch` parameter to actually be the name of the default branch

so maybe what we really need is a client config option like, `execution_branch` or something to cover this use case

ajtmccarty (approved):

PR #531, opened 2025-09-05
Add `create_diff` to the clients to manually create a diff between two times.

minitriga:
## Summary by CodeRabbit

* New Features
  * Create diff summaries between two timestamps, with optional background execution.
  * Retrieve diff summaries filtered by name and/or time range.
* Bug Fixes
  * Top-level node actions in diff summaries now show UPDATED when applicable instead of None.
* Documentation
  * Changelog entry added for the new diff creation API and time-range support.
* Tests
  * Unit tests updated to reflect corrected top-level node action behavior.

### DIFF
--- a/changelog/529.added.md
+++ b/changelog/529.added.md
@@ -1,0 +1,2 @@
+Add `create_diff` method to create a diff summary between two timestamps
+Update `get_diff_summary` to accept optional time range parameters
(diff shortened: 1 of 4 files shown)


ogenstad (approved):
LGTM, though it would be good with some integration tests as well so we can validate that the new methods are working as intended.

minitriga:
@ogenstad Yes I was unsure if they would be required as we dont have any integration tests for a similar API (branch create).

PR #536, opened 2025-09-09
Fix issue with `render` ctl command not using branch environment variables

### LINKED ISSUE #535: bug: CTL Render command does use local git branch
### Component

_No response_

### Infrahub SDK version

1.14.0

### Current Behavior

User notes that the command does not support the environment variables `INFRAHUB_DEFAULT_BRANCH` or `INFRAHUB_DEFAULT_BRANCH_FROM_GIT`. This is inconsistent with how other parts of the system might handle default branch settings.

### Expected Behavior

Expected Behavior (as proposed)
The expected behavior would be a consistent and hierarchical approach to determining the default branch, following this order of precedence:

Explicit CLI Argument: If the user provides a branch name using the --branch option (e.g., infrahubctl render --branch my-feature-branch), this should take the highest priority.

Environment Variables: If no branch is specified on the command line, infrahubctl render should check for and respect the INFRAHUB_DEFAULT_BRANCH and INFRAHUB_DEFAULT_BRANCH_FROM_GIT environment variables. This would align its behavior with other Infrahub tools.

Local Git Repository: As a fallback, if neither a CLI argument nor an environment variable is found, the command would default to reading the active branch from the local Git repository, which is the current behavior.

### Steps to Reproduce

- `git checkout -b feature/test-branch`
- `export INFRAHUB_DEFAULT_BRANCH=main`
- `infrahubctl render device`

The current behavior is for the command to render the template using the active local Git branch (feature/test-branch), ignoring the `INFRAHUB_DEFAULT_BRANCH` environment variable you just set.

### Additional Information

_No response_

minitriga:
Fixes: #535 

## Summary by CodeRabbit

- Bug Fixes
  - Fixed branch resolution so commands and GraphQL calls consistently use the branch from CLI, environment default, or repository-detected branch.
- Tests
  - Added tests to verify branch selection across CLI, environment, and Git sources; added supporting test fixtures.
- Chores
  - Added a changelog entry documenting the fix.

### DIFF
--- a/changelog/535.fixed.md
+++ b/changelog/535.fixed.md
@@ -1,0 +1,1 @@
+Fix branch handling in `_run_transform` and `execute_graphql_query` functions.
--- a/changelog/535.fixed.md
+++ b/changelog/535.fixed.md
@@ -1,1 +1,1 @@
-Fix branch handling in `_run_transform` and `execute_graphql_query` functions.
+Fix branch handling in `_run_transform` and `execute_graphql_query` functions in Infrahubctl to use environment variables for branch management. 
--- a/infrahub_sdk/ctl/cli_commands.py
+++ b/infrahub_sdk/ctl/cli_commands.py
@@ -208,7 +208,6 @@ async def _run_transform(
         debug: Prints debug info to the command line
         repository_config: Repository config object. This is used to load the graphql query from the repository.
     """
-    branch = get_branch(branch)
 
     try:
         response = execute_graphql_query(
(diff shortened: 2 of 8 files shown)


infrahub_sdk/ctl/cli_commands.py
```diff
@@ -208,7 +208,6 @@ async def _run_transform(
         debug: Prints debug info to the command line
         repository_config: Repository config object. This is used to load the graphql query from the repository.
     """
-    branch = get_branch(branch)
```
dgarros, line 211: Please can you check if other commands are using get_branch(), I believe `validate_graphql` might be using it as well

changelog/535.fixed.md
```diff
@@ -0,0 +1 @@
+Fix branch handling in `_run_transform` and `execute_graphql_query` functions.
```
dgarros: Would be good to update the message to indicate that it's only impacting `infrahubctl` not the SDK itself

infrahub_sdk/ctl/cli_commands.py
minitriga: Covered that one now.

dgarros (approved):

PR #546, opened 2025-09-14
Add command to generate Pydantic model based on a GraphQL query

### LINKED ISSUE #545: feature: Generate Pydantic models per GraphQL query
### Component

infrahubctl CLI

Implement a way to generate typed models (e.g. using Pydantic) from arbitrary GraphQL queries, so responses can be checked / used as strongly typed objects rather than raw dicts.

```
infrahubctl graphql generate-return-types myquery.qgl
```

### Describe the Use Case

Currently GraphQL query results are returned as raw dictionaries which makes it difficult to catch bugs if the query changes but not the code processing the query.

Ideally each query should have its own Pydantic models that will match exactly what the query is returning the user experience for the developer would be much better and it would be easier to catch errors in the IDE, overall it would make the code more maintainable and testable.

This would be especially useful for Python Transform, Checks and Generators that are heavily relying on a GraphQL query

### Additional Information

Ideally the methods that are currently returning the data from GraphQL should be updated to support this new approach and be able to return the model directly instead of returning a current dictionary when provided to them.
- Python transform
- Check
- Generator
- `client.execute_graphql`

```python
data = await client.execute_graphql(query=query, return_type=MyQueryModel)
```

dgarros:
Fixes #545 

This PR adds a new command to generate Pydantic model that matches the format of a given GraphQL query
For each graphql query, a python file with the same name will be created with the corresponding Pydantic models.
```
infrahubctl graphql generate-return-types
```

> The current implementation support fragments that are stored within the same query file 

In order to generate the return types, the graphql schema needs to exist locally so this PR also adds a command to fetch the graphql schema from the infrahub instance.

```
infrahubctl graphql export-schema
```

## Implementation 

Internally, the command is leveraging the library `ariadne-codegen` to generate the Pydantic models. `ariadne-codegen` provide a plugin system that makes it very easy to control the format of the generated file, currently we are doing some small adjustment but the system is very powerful.
The main drawback if using ``ariadne-codegen` is that by default the system has been designed to generate not just the return types but the full client so a wrapper was required to only generate the types.

## Summary by CodeRabbit

- **New Features**
  - Adds a graphql CLI group with commands to export a schema and generate Pydantic return types; SDK exposes GraphQL helpers and schema fetch APIs.

- **Documentation**
  - CLI docs for graphql commands and a Python guide on generating Pydantic models from GraphQL queries; minor typo fix.

- **Refactor**
  - Improved GraphQL rendering and query builder surface; new public constants and plugins to produce type-safe models.

- **Tests**
  - New unit tests and fixtures for renderers, plugins, and codegen flow.

- **Chores**
  - Adds optional codegen dependency and removes a branded-term mapping.


docs/docs/python-sdk/guides/python-typing.mdx
```diff
@@ -101,3 +101,51 @@ my_object = client.get(MyOwnObject, name__value="example")
 ```
 
 > if you don't have your own Python module, it's possible to use relative path by having the `procotols.py` in the same directory as your script/transform/generator
+
+## Generating Pydantic models from GraphQL queries
```
ogenstad, line 105: As you're working with this file can you also fix the typo with `procotols.py` just above that code rabbit is complaining about?

docs/docs/python-sdk/guides/python-typing.mdx
```diff
@@ -101,3 +101,51 @@ my_object = client.get(MyOwnObject, name__value="example")
...
+
+When working with GraphQL queries, you can generate type-safe Pydantic models that correspond to your query return types. This provides excellent type safety and IDE support for your GraphQL operations.
+
+### Why use generated return types?
+
+Generated Pydantic models from GraphQL queries offer several important benefits:
+
+- **Type Safety**: Catch type errors at development time instead of runtime
```
ogenstad: Suggested change:

```diff
- Catch type errors at development time instead of runtime
+ Catch type errors during development time instead of at runtime
```

infrahub_sdk/ctl/graphql.py
```diff
@@ -0,0 +1,174 @@
...
+    Find all files with .gql extension in the specified directory.
+
+    Args:
+        query_path: Path to the directory to search for .gql files
+
+    Returns:
+        List of Path objects for all .gql files found
+    """
+    if not query_path.exists():
+        raise FileNotFoundError(f"Directory not found: {query_path}")
```
ogenstad: This could look a bit weird if `query_path` was a file instead of a directory. I.e. that we'd say "Directory not found" and point to a file could be confusing.

infrahub_sdk/ctl/graphql.py
```diff
@@ -0,0 +1,174 @@
...
+    queries_str = queries_path.read_text(encoding="utf-8")
+    queries_ast = parse(queries_str)
+    validation_errors = validate(
+        schema=schema,
+        document_ast=queries_ast,
+        rules=[r for r in specified_rules if r is not NoUnusedFragmentsRule],
+    )
+    if validation_errors:
+        raise ValueError("\n\n".join(error.message for error in validation_errors))
+    return queries_ast.definitions
```
ogenstad, line 75: While we don't support it yet within the SDK this would be problematic with fragments spread out in multiple files. Not an issue for this PR, just highlighting if we want to do something different with regards to the API of the command because of this.

infrahub_sdk/ctl/graphql.py
```diff
@@ -0,0 +1,174 @@
...
+@app.command()
+@catch_exception(console=console)
+async def export_schema(
+    destination: Path = typer.Option("schema.graphql", help="Path to the GraphQL schema file."),
+    _: str = CONFIG_PARAM,
+) -> None:
+    """Export the GraphQL schema to a file."""
+
+    client = initialize_client()
+    response = await client._get(url=f"{client.address}/schema.graphql")
```
ogenstad: I'd suggest that we move this and avoid calling `client._get` from the CTL. Instead I's suggest that we add some `client.schema.get_graphql_schema()`

Do we need a branch parameter here?

infrahub_sdk/ctl/graphql.py
```diff
@@ -0,0 +1,174 @@
...
+    for gql_file in gql_files:
+        gql_per_directory[gql_file.parent].append(gql_file)
+
+    # Generate the Pydantic Models for the GraphQL queries
+    for directory, gql_files in gql_per_directory.items():
+        for gql_file in gql_files:
+            try:
+                definitions = get_graphql_query(queries_path=gql_file, schema=graphql_schema)
+            except ValueError as exc:
+                print(f"Error generating result types for {gql_file}: {exc}")
```
ogenstad: We should use the `console.print` method to show output to users.

infrahub_sdk/ctl/graphql.py
```diff
@@ -0,0 +1,174 @@
...
+                    include_comments=CommentsStrategy.NONE,
+                ),
+                plugin_manager=plugin_manager,
+            )
+
+            try:
+                for query_operation in queries:
+                    package_generator.add_operation(query_operation)
+            except ParsingError as exc:
+                console.print(f"[red]Unable to process {gql_file.name}: {exc}")
```
ogenstad, line 173: It looks like we ignore this error and the one above aside from displaying output. Do we want the command to fail here?

infrahub_sdk/graphql/plugin.py
```diff
@@ -0,0 +1,86 @@
...
+        for idx, item in enumerate(module.body):
+            if isinstance(item, ast.ImportFrom) and item.module == "base_model":
+                return idx
+        return -1
+
+    @classmethod
+    def replace_base_model_import(cls, module: ast.Module) -> ast.Module:
+        base_model_index = cls.find_base_model_index(module)
+        if base_model_index == -1:
+            raise ValueError("BaseModel not found in module")
```
ogenstad: Will we call `.find_base_model_index()` in other locations too? I think we should avoid logic like this where `.replace_base_model_import()` has to know what a `-1` index means. In other situations it could just as well be the last item in a sequence. I think it would b cleaner if it was `find_base_model_index` that raised the ValueError directly.

infrahub_sdk/graphql/query.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+        lines.append("}")
+
+        return "\n" + "\n".join(lines) + "\n"
+
+
+class Mutation(BaseGraphQLQuery):
+    query_type = "mutation"
+
+    def __init__(self, *args: Any, mutation: str, input_data: dict, **kwargs: Any):
```
ogenstad, line 49: Realise this is just moved code but I think we should remove`*args: Any` and also `**kwargs: Any)` to make this class easier to use..

infrahub_sdk/graphql/utils.py
```diff
@@ -0,0 +1,30 @@
...
+    return -1
+
+
+def insert_fragments_inline(module: ast.Module, fragment: ast.Module) -> ast.Module:
+    """Insert the Pydantic classes for the fragments inline into the module."""
+    module_class_def_index = get_class_def_index(module)
+
+    fragment_classes: list[ast.ClassDef] = [item for item in fragment.body if isinstance(item, ast.ClassDef)]
+    for idx, item in enumerate(fragment_classes):
+        module.body.insert(module_class_def_index + idx, item)
```
ogenstad: What would this look like if `get_class_def_index()` returns a -1. The first entry gets added to the end and then the other fragments gets added starting from the beginning?

ogenstad (commented):
Added some comments, but great addition!

infrahub_sdk/graphql/query.py
dgarros: I think it's best to do that in a separate PR to avoid adding unexpected issues

infrahub_sdk/ctl/graphql.py
dgarros: Good catch, update the logic to ignore failing files and continue

infrahub_sdk/ctl/graphql.py
dgarros: I agree we'll need to revisit this part of the code to improve the support for Fragments

infrahub_sdk/ctl/graphql.py
dgarros: Good call, done

infrahub_sdk/ctl/graphql.py
```diff
@@ -0,0 +1,184 @@
...
+        raise FileNotFoundError(f"File or directory not found: {query_path}")
+
+    if not query_path.is_dir() and query_path.is_file():
+        return [query_path]
+
+    return list(query_path.glob("**/*.gql"))
+
+
+def get_graphql_query(queries_path: Path, schema: GraphQLSchema) -> tuple[DefinitionNode, ...]:
+    """Get graphql queries definitions from a single GraphQL file."""
```
ogenstad: Nitpicking here but we should use the same case for GraphQL throughout, i.e. the first graphql should be changed to GraphQL.

ogenstad (approved):

PR #548, opened 2025-09-15
Add support for clearing optional attributes

minitriga:
## Summary by CodeRabbit

- New Features
  - You can clear optional dropdown attributes by setting them to None; the generated payload now explicitly includes null for that field.

- Bug Fixes
  - Input data generation no longer omits optional dropdown fields set to None after user mutation.

- Tests
  - Added tests validating dropdown attributes, including clearing values to None and checking the resulting payload.

- Chores
  - Updated changelog to document clearing optional dropdown attributes.


infrahub_sdk/node/attribute.py
```diff
@@ -76,6 +76,8 @@ def _generate_input_data(self) -> dict | None:
         variables: dict[str, Any] = {}
 
         if self.value is None:
+            if self._schema.kind == "Dropdown" and self._schema.optional:
```
dgarros: Looking at it, I'm wondering if the issue is specific to the Dropdown or if we have the same issue with other kind of attributes too ?

minitriga: I should have resolved this now by only clearing an optional attribute if it has been mutated.

ogenstad (approved):

minitriga:
Fixes: #549

Issue #528, opened 2025-09-04
task: Replace 'toml' package


ogenstad:
### Task Description

We currently have the 'toml' package listed as an optional requirement in the SDK:

https://github.com/opsmill/infrahub-sdk-python/blob/v1.14.0/pyproject.toml#L39

Since Python 3.11 there's a module called `tomllib` bundled with Python and according to https://realpython.com/python311-tomllib/ it looks like its API is the same as another package called `tomli`.

Meaning that when we import `toml` here:
https://github.com/opsmill/infrahub-sdk-python/blob/v1.14.0/infrahub_sdk/ctl/config.py#L7

We could instead have:
```python
try:
    import tomllib
except ModuleNotFoundError:
    import tomli as tomllib
```

The pyproject.toml file would need to be updated to optionally install `tomli` when using Python 3.9 and 3.10 and when using Python 3.11 or above we wouldn't need to install any additional toml package and as such we can remove one of the external dependencies from the SDK.

ogenstad:
Fixed in #551.

PR #551, opened 2025-09-17
IFC-1811: Replace toml package with tomllib and tomli optionally

solababs:
## Description
Replace toml package with tomllib and tomli optionally when project python version is `<3.11` as tomllib ships with `python 3.11`

## Test
[x] - Run test
[x] - Update docs
[x] - Add towncrier message

## Summary by CodeRabbit

- **New Features**
  - Configuration parsing now uses the platform-native TOML parser on newer Python releases with a compatible fallback for older versions.

- **Documentation**
  - Minor docs formatting adjustment; syntax highlighting unchanged.
  - Added changelog entry noting the TOML dependency change.

- **Chores**
  - CI reduces default dependency installs and conditionally enables TOML support for older Python versions.
  - Added pre-commit config with automated linting/formatting hooks.

- **Refactor**
  - Replaced legacy TOML dependency with a standards-aligned approach.


pyproject.toml
```diff
@@ -46,6 +45,7 @@ whenever = ">=0.7.2,<0.8.0"
 netutils = "^1.0.0"
 click = { version = "8.1.*", optional = true }
 copier = { version = "^9.8.0", optional = true }
+tomli = { version = ">=1.1.0", python = "<3.11" }
```
ogenstad: I think we still want `tomli` to be optional=True and included in the poetry.extras `ctl` section just like 'toml' was as we don't need to have it installed if someone is only installing the SDK and won't use the `infrahubctl` command.

ogenstad (commented):
LGTM. I just added a minor comment regarding the optional aspect of the tomli package.

One additional thing that would be nice is that we're using [towncrier](https://towncrier.readthedocs.io/) to generate release notes. As part of this PR it would be good to run `towncrier create` and indicate that the change is for issue #528 (without the pound character). Then just write it as a housekeeping issue that we've removed "toml" as a requirement and added "tomli" for Python versions under 3.11.

changelog/528.housekeeping.md
```diff
@@ -0,0 +1 @@
+Replace toml package with tomllib and tomli optionally for when python version is less than 3.11
```
ogenstad: We typically write Python with a capital "P".

ogenstad (approved):
LGTM, I just added a minor comment about the casing for Python.

PR #554, opened 2025-09-18
Add convert_object_type method

LucasG0:
## Summary by CodeRabbit

- New Features
  - Added ability to convert an existing object to a different type via the SDK, available in both async and sync clients.
  - Supports specifying a target type, optional branch, and flexible field mapping for attributes and relationships (scalars and single/multiple links).
- Documentation
  - Added changelog entry describing the new object-type conversion capability.
- Tests
  - Added integration tests covering end-to-end conversion, mapping, and assertions for both client types.


infrahub_sdk/convert_object_type.py
```diff
@@ -0,0 +1,60 @@
...
+                fields_mapping: $fields_mapping
+            }) {
+                ok
+                node
+        }
+    }
+"""
+
+
+class ConversionFieldValue(BaseModel):  # Only one of these fields can be not None
```
LucasG0, line 21: I'll make a follow up PR infrahub side to use these classes from the SDK instead of having them duplicated

infrahub_sdk/convert_object_type.py
```diff
@@ -0,0 +1,60 @@
...
+    Holds the new value of the destination field during an object conversion.
+    Use `attribute_value` to specify the new raw value of an attribute.
+    Use `peer_id` to specify new peer of a cardinality one relationship.
+    Use `peer_ids` to specify new peers of a cardinality many relationship.
+    Only one of `attribute_value`, `peer_id` and `peers_ids` can be specified.
+    """
+
+    attribute_value: Any | None = None
+    peer_id: str | None = None
+    peers_ids: list[str] | None = None
```
ajtmccarty, line 32: the docstring says `peer_ids`, but the code says `peers_ids`
I think `peer_ids` is better, but, either way, they should match

tests/integration/test_convert_object_type.py
```diff
@@ -0,0 +1,100 @@
...
+        person_1 = await client.create(kind="TestconvPerson1", name=f"person_{uuid.uuid4()}")
+        await person_1.save()
+        car_1 = await client.create(kind="TestconvCar", name=f"car_{uuid.uuid4()}")
+        await car_1.save()
+
+        new_age = 25
+        fields_mapping = {
+            "name": ConversionFieldInput(source_field="name"),
+            "age": ConversionFieldInput(data=ConversionFieldValue(attribute_value=new_age)),
+            "worst_car": ConversionFieldInput(data=ConversionFieldValue(peer_id=car_1.id)),
```
ajtmccarty, line 79: is this in here to test that it is ignored?

tests/integration/test_convert_object_type.py
```diff
@@ -0,0 +1,100 @@
...
+            person_2 = client_sync.convert_object_type(
+                node_id=person_1.id,
+                target_kind="TestconvPerson2",
+                branch=client.default_branch,
+                fields_mapping=fields_mapping,
+            )
+
+        assert person_2.get_kind() == "TestconvPerson2"
+        assert person_2.name.value == person_1.name.value
+        assert person_2.age.value == new_age
```
ajtmccarty, line 100: maybe worth testing that `fastest_cars` is correctly updated too

ajtmccarty (approved):

tests/integration/test_convert_object_type.py
LucasG0: No, I updated schema to replace `my_car` by `worst_car`

infrahub_sdk/convert_object_type.py
LucasG0: Going for `peers_ids` as it what the server expects

Issue #464, opened 2025-07-04
bug: Schema loading process crashes ungracefully with wrong schema file


a-delannoy:
### Component

infrahubctl

### Infrahub SDK version

1.12.1

### Current Behavior

When trying to load a specific wrongly structured schema (defined in the steps to reproduce the issue), the schema loading process fails ungracefully with the following error message:

```zsh
Unable to load the schema:
Error: list indices must be integers or slices, not str
Traceback (most recent call last):
  File "/Users/Antoine/projects/dogfooding/.venv/lib/python3.12/site-packages/infrahub_sdk/ctl/utils.py", line 91, in async_wrapper
    return await func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Antoine/projects/dogfooding/.venv/lib/python3.12/site-packages/infrahub_sdk/ctl/schema.py", line 129, in load
    display_schema_load_errors(response=response.errors, schemas_data=schemas_data)
  File "/Users/Antoine/projects/dogfooding/.venv/lib/python3.12/site-packages/infrahub_sdk/ctl/schema.py", line 80, in display_schema_load_errors
    input_label = node[loc_path[6]].get("name", None)
                  ~~~~~~~~~~~~~~^^^^^^^^^^^^^
TypeError: list indices must be integers or slices, not str
```

### Expected Behavior

According to the backend team, this error should be handled as a gracefull error handling.

#### Infrahub instance

The Infrahub instance is setup using the docker compose instantiation on the [infrahub repository](https://github.com/opsmill/infrahub/tree/35f6192207a577f9783d874618ae73147dbae308)

The command used to run the instance is: `inv demo.start`

#### Environement variables

Only two env vars are defined:

* INFRAHUB_API_TOKEN=<default api token>
* INFRAHUB_ADDRESS=http://localhost:8000

#### The schema file

To load the schema, I use the following command:

`infrahubctl schema load schema-library/base security-schema.yaml`

Where:

* `schema-library/base` is conform to [this state](https://github.com/opsmill/schema-library/tree/5ee4805668634bb476123c550bc2c3ebbc48f442) in the `schema-library` repository
* `security-schema.yaml` has the following content

```yaml
---
version: "1.0"
generics:
  - name: Rule
    namespace: Security
    icon: mdi:security
    order_by:
      - index__value
    display_labels:
      - name__value
    uniqueness_constraints:
      - ["rule_set", "index__value"]
[...]

PR #555, opened 2025-09-19
IHS-147: Fix schema load failure exception

solababs:
## Summary by CodeRabbit

- Bug Fixes
  - Improved handling of deeply nested schema load errors to correctly extract and display attribute labels for both list- and index-based structures; error lines are clearer and consistent.

- Tests
  - Added unit test covering detailed error reporting when errors occur inside attributes or relationships, asserting formatted output and node lookup behavior.

- Chores
  - Added a changelog entry noting schema load now reports errors gracefully instead of raising an exception.

### DIFF
--- a/changelog/464.housekeeping.md
+++ b/changelog/464.housekeeping.md
@@ -1,0 +1,1 @@
+Handle error gracefully when loading schema instead of failing with an exception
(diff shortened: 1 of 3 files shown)


ogenstad (approved):
LGTM, can you also add an entry to towncrier indicating that #464 was fixed?

PR #578, opened 2025-10-12
Add flags to `CoreGeneratorDefinition` to control when each generator should be executed

dgarros:
Add the flags `execute_in_proposed_change` and `execute_after_merge` to `CoreGeneratorDefinition` to allow the user to control when a generator should be executed.

Related to https://github.com/opsmill/infrahub/pull/7183

## Summary by CodeRabbit

- New Features
  - Added configurable options to control whether generators run within proposed changes and/or after merges (both enabled by default).
  - Generators now automatically adjust their grouping behavior based on post-merge execution settings for more accurate tracking and organization.
  - Repository views now expose related group objects, improving discoverability and navigation of generator-linked data.

- Documentation
  - Descriptions provided for the new execution options to clarify behavior and defaults.

### DIFF
--- a/infrahub_sdk/generator.py
+++ b/infrahub_sdk/generator.py
@@ -26,6 +26,8 @@ class InfrahubGenerator(InfrahubOperation):
         generator_instance: str = "",
         params: dict | None = None,
         convert_query_response: bool = False,
+        execute_in_proposed_change: bool = True,
+        execute_after_merge: bool = True,
         logger: logging.Logger | None = None,
         request_context: RequestContext | None = None,
     ) -> None:
@@ -44,6 +46,8 @@ class InfrahubGenerator(InfrahubOperation):
         self._client: InfrahubClient | None = None
         self.logger = logger if logger else logging.getLogger("infrahub.tasks")
         self.request_context = request_context
+        self.execute_in_proposed_change = execute_in_proposed_change
+        self.execute_after_merge = execute_after_merge
 
     @property
     def subscribers(self) -> list[str] | None:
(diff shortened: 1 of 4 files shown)


infrahub_sdk/generator.py
```diff
@@ -16,6 +16,14 @@
 class InfrahubGenerator(InfrahubOperation):
     """Infrahub Generator class"""
 
+    generator_instance: str
+    execute_in_proposed_change: bool
+    execute_after_merge: bool
+    logger: logging.Logger
+    request_context: RequestContext | None
+    params: dict
+    query: str
```
ogenstad: I'm not sure these values do anything aside from providing potential confusion if there's a mismatch as to what gets assigned in the `__init__()` method. Why would this be needed?

infrahub_sdk/operation.py
```diff
@@ -12,6 +12,10 @@
 
 
 class InfrahubOperation:
+    branch: str
+    root_directory: str
+    convert_query_response: bool
```
ogenstad: Same comment as above. Not sure if we're trying to solve a problem that I'm not aware of?

ogenstad (approved):
LGTM aside from the comments on the additional attribute definitions outside of `__init__`

infrahub_sdk/generator.py
dgarros: Agreed this is not mandatory
I find it easier to read what are the attributes available at the object level and it provide a more explicit typing 
I can remove it if you prefer

ogenstad: There are two things that I'd like to avoid. The first one is if the attribute itself doesn't actually exist and the second is if we mix up the types.

```python
class MyClass:
    query: dict
    logger: logging.Logger

    def __init__(self, query: str) -> None:
        self.query = query


m = MyClass(query="hello")

m.logger.info("Log something")
```

The first one would be if mypy or ruff complains about `query` being defined as a dict under the class but as an `str` under the init method. I think that this might get picked up. The other one is worse where it looks like the class has a logger attribute but in fact it doesn't and the last line will produce an `AttributeError` and I'm not sure if any tool would complain about this before we run into the problem.

Perhaps we should convert these classes to dataclasses and make it cleaner. I agree with you that it's easier to read if we have the attributes clearly defined but I think it could also be risky if we don't have a control that the attributes actually are set.

PR #580, opened 2025-10-14
Add `tls_ca_file` and `tls_insecure` configuration settings to infrahubctl

wvandeun:
Add the `tls_ca_file` and `tls_insecure` configuration settings for infrahubctl.
This allows you to define these configuration settings in the infrahubctl configuration file. 

It was already possible to configure these settings through environment variables.

## Summary by CodeRabbit

* **New Features**
  * Added TLS options to configure a custom CA certificate file and optionally disable certificate verification.
  * Available via environment variables (INFRAHUB_TLS_CA_FILE, INFRAHUB_TLS_INSECURE) and corresponding configuration file entries.
  * Client now respects these settings when establishing connections.

* **Documentation**
  * Updated environment variables table and configuration examples to include the new TLS options and defaults.

### DIFF
--- a/docs/docs/infrahubctl/infrahubctl.mdx
+++ b/docs/docs/infrahubctl/infrahubctl.mdx
@@ -31,6 +31,8 @@ The `infrahubctl` command line utility is installed as a part of the [Infrahub S
 | `INFRAHUB_ADDRESS`        | http://localhost:8000                  |
 | `INFRAHUB_API_TOKEN`      | `06438eb2-8019-4776-878c-0941b1f1d1ec` |
 | `INFRAHUB_DEFAULT_BRANCH` | main                                   |
+| `INFRAHUB_TLS_CA_FILE`    | /path/to/ca.crt                        |
+| `INFRAHUB_TLS_INSECURE`   | true                                   |
 
 > You can also provide the location of a configuration file via the environment variable `INFRAHUBCTL_CONFIG`.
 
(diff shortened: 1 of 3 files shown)


docs/docs/infrahubctl/infrahubctl.mdx
```diff
@@ -31,6 +31,8 @@ The `infrahubctl` command line utility is installed as a part of the [Infrahub S
 | `INFRAHUB_ADDRESS`        | http://localhost:8000                  |
 | `INFRAHUB_API_TOKEN`      | `06438eb2-8019-4776-878c-0941b1f1d1ec` |
 | `INFRAHUB_DEFAULT_BRANCH` | main                                   |
+| `INFRAHUB_TLS_CA_FILE`    | /path/to/ca.crt                        |
+| `INFRAHUB_TLS_INSECURE`   | true                                   |
```
ogenstad, line 35: This should be an exact replica of what we support in the SDK as we reuse the existing environment variables. As such I think that instead of adding options here we can have the INFRAHUB_ADDRESS one as an example and then link to the config page for all of the SDK environment variables.

ogenstad (approved):
LGTM, I'd change the stuff in the docs.

Does this mean that people are using `infrahubctl.toml` files?

I think we can merge this change for now but that we should get rid of the config class for Infrahubctl. If we want to provide a file the expected format should be a 1-1 match against what the SDK itself supports so that we don't have to add specific entries like this and include the if-statement logic for setting up the config of the client within the SDK.

PR #581, opened 2025-10-15
feat(client): add ssl context cache and config

fatih-acar:
This allows configuring a custom SSL context or cache a default one if not supplied instead of re-creating a default ssl context on each request (default httpx behavior).

## Summary by CodeRabbit

* **New Features**
  * Support supplying a custom SSL/TLS context via configuration.
  * Config object is now exposed in the public API for easier setup and customization.

* **Refactor**
  * TLS verification behavior unified across synchronous and asynchronous clients.

* **Bug Fixes**
  * TLS configuration errors are surfaced earlier (at client construction) for faster diagnosis.


infrahub_sdk/client.py
```diff
@@ -143,13 +144,38 @@ def __init__(
...
+        self._cached_ssl_context: SSLContext | None = None
+        self._cached_ssl_context_source: str | None = None
         self.group_context: InfrahubGroupContext | InfrahubGroupContextSync
         self._initialize()
         self._request_context: RequestContext | None = None
 
     def _initialize(self) -> None:
         """Sets the properties for each version of the client"""
 
+    def _get_verify_config(self) -> bool | SSLContext:
```
ogenstad: I'd suggest to move this to the config class instead and have it as a property there.

Then have something like:

```python
        async with httpx.AsyncClient(
            **proxy_config,  # type: ignore[arg-type]
            verify=self.config.tls_context,
        ) as client:
```

infrahub_sdk/client.py
```diff
@@ -143,13 +144,38 @@ def __init__(
...
+
+        if self.config.tls_insecure:
+            return False
+
+        if self.config.tls_ca_file:
+            if self._cached_ssl_context is None or self._cached_ssl_context_source != self.config.tls_ca_file:
+                self._cached_ssl_context = create_default_context(cafile=self.config.tls_ca_file)
+                self._cached_ssl_context_source = self.config.tls_ca_file
+            if self._cached_ssl_context is None:
+                raise RuntimeError("Failed to create SSL context using provided CA file")
```
ogenstad: Will `create_default_context` ever return a `None`? I'd expect it to fail if it's not able to create a context. Will we ever get to this runtime error?

infrahub_sdk/config.py
```diff
@@ -78,6 +79,7 @@ class ConfigBase(BaseSettings):
     Can be useful to test with self-signed certificates.""",
     )
     tls_ca_file: str | None = Field(default=None, description="File path to CA cert or bundle in PEM format")
+    tls_ssl_context: SSLContext | None = Field(default=None, description="Custom SSL context used for TLS connections")
```
ogenstad: Having a class like this makes it problematic from a documentation point of view and we'd be expected to be able to use `INFRAHUB_TLS_SSL_CONTEXT` as an environment variable to populate this option.

If we need this here I'd instead have it as a private attribute.

```python
_ssl_context: SSLContext | None = PrivateAttr(default=None)
```

Then the property could set this value when it caches the context. If we want a setter in order to transfer the SSL context from Infrahub we can have a specific method within the SDK config class that would set this private field.

ogenstad (commented):
Added some suggestions. With this change I'd also modify it so that we try to load and cache the SSL config when we intialize the config so that an eventual error happens early and not when we first try to use the SDK.

infrahub_sdk/client.py
fatih-acar: Right, that was LLM generated... checked manually if create_default_context ever returns `None` and it looks like not (it directly raises exceptions). (Also asked LLM about that and it agreed... :man_facepalming:)

infrahub_sdk/config.py
fatih-acar: Good point thanks, it was also an issue due to Pydantic serialization... switching to PrivateAttr also solves that.

fatih-acar:
@ogenstad thank you for the relevant suggestions, it's much better now.

infrahub_sdk/config.py
```diff
@@ -133,6 +135,25 @@ def default_infrahub_branch(self) -> str:
     def password_authentication(self) -> bool:
         return bool(self.username)
 
+    @property
+    def tls_context(self) -> bool | ssl.SSLContext:
+        if self._ssl_context:
+            return self._ssl_context
+
+        if self.tls_insecure:
+            return False
```
ogenstad: I thought you mentioned that there was some processing needed even if we just used HTTP to create the context each time? Is it that this doesn't happen if we use HTTP together with tls_insecure=True?

infrahub_sdk/client.py
```diff
@@ -146,6 +146,7 @@ def __init__(
         self.group_context: InfrahubGroupContext | InfrahubGroupContextSync
         self._initialize()
         self._request_context: RequestContext | None = None
+        _ = self.config.tls_context  # Early load of the TLS context to catch errors
```
ogenstad, line 149: I would prefer if this was cached by the config during the creation of the config object and we populated `config.__ssl_context` at that time. However if we do that I'm not sure that it it would be possible to use the `.set_ssl_context()` method within the Config.clone, or rather it would be possible but at that time we'd already have created a new SSL Context within the clone method. So I think this might be the best option for now. I guess the only thing to remember is that we'd need to initialize a Client with a config in order to populate the config for reuse elsewhere but I think we always do that.

In order to ensure this stays in place I think we'll want an additional unittest below where you also clone the config and verify that the context is the same object in both of the configs.

ogenstad (approved):
LGTM, I added a suggestion along with a question about when we're returning `False` for tls_insecure.

Also can you please add a newsfragment?

One change in behaviour could be that when we move to creating the SSL context like we do that the behaviour gets slightly modified with regards to situations when we use the CA file as an environment variable.

Within Infrahub we have this code, it's quite hacky but you see that the parameters to the SSL context is different if we're loading a file `cafile` vs when we are loading a string from an environment variable `cadata`.

```python
        tls_ca_path = Path(self.tls_ca_bundle)

        try:
            possibly_file = tls_ca_path.exists()
        except OSError:
            # Raised if the filename is too long which can indicate
            # that the value is a PEM certificate in string form.
            possibly_file = False

        if possibly_file and tls_ca_path.is_file():
            context = ssl.create_default_context(cafile=str(tls_ca_path))
        else:
            context = ssl.create_default_context()
            context.load_verify_locations(cadata=self.tls_ca_bundle)
```

It might be that HTTPX manages this part differently internally (or it was never tested).

Can you please verify with your branch the behaviour if you do a:

```sh
export INFRAHUB_TLS_CA_FILE=PEM-CERT-AS-STRING
```

I thought we had tests for that in the backend but now see that we don't seem to have that :(

infrahub_sdk/config.py
fatih-acar: Yes, thanks for the reminder, you are right (https://github.com/encode/httpx/blob/0.28.1/httpx/_config.py#L41)
I replaced this plain `return False` by a manually crafted SSLContext (similar to what's done within httpx).

infrahub_sdk/client.py
fatih-acar: Added a test for clone

fatih-acar:
> Can you please verify with your branch the behaviour if you do a:
export INFRAHUB_TLS_CA_FILE=PEM-CERT-AS-STRING

@ogenstad 

```
INTERNALERROR>   File "/home/ubuntu/.local/share/uv/python/cpython-3.12.11-linux-aarch64-gnu/lib/python3.12/ssl.py", line 707, in create_default_context
INTERNALERROR>     context.load_verify_locations(cafile, capath, cadata)
INTERNALERROR> OSError: [Errno 36] File name too long
...
```

I think this is expected. Checking the infrahub code, I see that `tls_ca_bundle` is only available for the HTTP adapter. For the other adapters, it's `tls_ca_file` (like we do within the SDK). So we only allow passing a filename/path, not a plaintext CA bundle. Do we want to support that?

ogenstad:
> > Can you please verify with your branch the behaviour if you do a:
> > export INFRAHUB_TLS_CA_FILE=PEM-CERT-AS-STRING
> 
> @ogenstad
> 
> ```
> INTERNALERROR>   File "/home/ubuntu/.local/share/uv/python/cpython-3.12.11-linux-aarch64-gnu/lib/python3.12/ssl.py", line 707, in create_default_context
> INTERNALERROR>     context.load_verify_locations(cafile, capath, cadata)
> INTERNALERROR> OSError: [Errno 36] File name too long
> ...
> ```
> 
> I think this is expected. Checking the infrahub code, I see that `tls_ca_bundle` is only available for the HTTP adapter. For the other adapters, it's `tls_ca_file` (like we do within the SDK). So we only allow passing a filename/path, not a plaintext CA bundle. Do we want to support that?

Ah yes the option is named differently. I think a problem here is the description of `tls_ca_file`:

```python
    tls_ca_file: str | None = Field(default=None, description="File path to CA cert or bundle in PEM format")
```

This is a case where the oxford comma would have been good as you can read this in a number of different ways. Let's ignore this for now since it's outside of the scope of your PR but I think we should clean it up a bit. For consistency we should perhaps start to replace `tls_ca_file` with `tls_ca_bundle` so it can work exactly like in Infrahub, but we'll consider that later.

PR #585, opened 2025-10-26
Replace strategy in object file with parameters section

dgarros:
This PR introduces a more flexible and extensible configuration system for Infrahub Object files. Instead of using a simple strategy field, Object files now support a comprehensive parameters section that allows for fine-grained control over data processing.

Object files now support a parameters section in the spec that controls how data is processed:

```yaml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: BuiltinLocation
  parameters:
    expand_range: true  # Enable range expansion
  data:
    - name: AMS[1-5]
      type: Country
```

## Future Extensibility
This new architecture makes it easy to add additional data processing parameters in the future, such as: Jinja2 processor 

## Summary by CodeRabbit

* **Documentation**
  * Renamed "Data Processing Strategies" to "Data Processing Parameters" and updated examples/narrative to use parameters.expand_range.

* **New Features / Updates**
  * Added parameters.expand_range (boolean, default false) to control when range expansion runs.
  * Range expansion produces multiple items when enabled and validates consistent lengths; mismatches yield a clear validation error.
  * Processing now uses a centralized parameters-driven pipeline.

* **Tests**
  * Tests updated to use parameters.expand_range and pre-seeded schema fixtures.

### DIFF
--- a/infrahub_sdk/spec/object.py
+++ b/infrahub_sdk/spec/object.py
@@ -168,7 +168,7 @@ async def get_relationship_info(
 
 class InfrahubObjectFileData(BaseModel):
     kind: str
-    parameters: InfrahubObjectParameters = InfrahubObjectParameters()
+    parameters: InfrahubObjectParameters = Field(default_factory=InfrahubObjectParameters)
     data: list[dict[str, Any]] = Field(default_factory=list)
 
     async def _get_processed_data(self, data: list[dict[str, Any]]) -> list[dict[str, Any]]:
--- a/infrahub_sdk/spec/object.py
+++ b/infrahub_sdk/spec/object.py
@@ -10,7 +10,6 @@ from ..schema import GenericSchemaAPI, RelationshipKind, RelationshipSchema
 from ..yaml import InfrahubFile, InfrahubFileKind
 from .models import InfrahubObjectParameters
 from .processors.factory import DataProcessorFactory
-from .processors.range_expand_processor import RangeExpandDataProcessor
 
 if TYPE_CHECKING:
     from ..client import InfrahubClient
@@ -558,7 +557,10 @@ class InfrahubObjectFileData(BaseModel):
                 rel_info.find_matching_relationship(peer_schema=peer_schema)
                 context.update(rel_info.get_context(value=parent_node.id))
 
-            expanded_data = RangeExpandDataProcessor.expand_data_with_ranges(data=data["data"])
+            expanded_data = await DataProcessorFactory.process_data(
+                kind=peer_kind, data=data["data"], parameters=parameters
+            )
+
             for idx, peer_data in enumerate(expanded_data):
                 context["list_index"] = idx
                 if isinstance(peer_data, dict):
--- a/infrahub_sdk/spec/object.py
+++ b/infrahub_sdk/spec/object.py
@@ -287,8 +287,9 @@ class InfrahubObjectFileData(BaseModel):
         context: dict | None = None,
         branch: str | None = None,
         default_schema_kind: str | None = None,
-        parameters: InfrahubObjectParameters = InfrahubObjectParameters(),
+        parameters: InfrahubObjectParameters | None = None,
     ) -> list[ObjectValidationError]:
+        parameters = parameters or InfrahubObjectParameters()
         context = context.copy() if context else {}
         errors: list[ObjectValidationError] = []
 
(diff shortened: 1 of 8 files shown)


infrahub_sdk/spec/processors/range_expand_processor.py
```diff
@@ -0,0 +1,51 @@
...
+                    new_item[key] = values[i]
+                expanded.append(new_item)
+        return expanded
+
+    @classmethod
+    async def process_data(
+        cls,
+        data: list[dict[str, Any]],
+    ) -> list[dict[str, Any]]:
+        return cls.expand_data_with_ranges(data)
```
ogenstad: As we're just calling another method here I'd suggest that we instead remove `expand_data_with_ranges` and implement that code directly here.

infrahub_sdk/spec/processors/data_processor.py
```diff
@@ -0,0 +1,10 @@
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+class DataProcessor(ABC):
+    """Abstract base class for data processing strategies"""
+
+    @abstractmethod
+    async def process_data(self, data: list[dict[str, Any]]) -> list[dict[str, Any]]:
```
ogenstad, line 9: Realising that when looking at the we're just sending opaque data back and forth without really knowing what to expect. While not tied to this PR it makes me wonder about the object specs that we have. We should be able to read the yaml files and convert them into a known structure instead, but that would be some follow up work.

infrahub_sdk/spec/object.py
```diff
@@ -173,97 +167,21 @@ async def get_relationship_info(
...
-    @classmethod
-    def register_processor(cls, strategy: ObjectStrategy, processor_class: type[DataProcessor]) -> None:
-        """Register a new processor for a strategy - useful for future extensions"""
-        cls._processors[strategy] = processor_class
-
-
 class InfrahubObjectFileData(BaseModel):
     kind: str
-    strategy: ObjectStrategy = ObjectStrategy.NORMAL
+    parameters: InfrahubObjectParameters = InfrahubObjectParameters()
```
ogenstad: Suggested change:

```diff
- parameters: InfrahubObjectParameters = InfrahubObjectParameters()
+ parameters: InfrahubObjectParameters = Field(default_factory=InfrahubObjectParameters)
```

infrahub_sdk/spec/object.py
```diff
@@ -370,7 +288,7 @@ async def validate_related_nodes(
         context: dict | None = None,
         branch: str | None = None,
         default_schema_kind: str | None = None,
-        strategy: ObjectStrategy = ObjectStrategy.NORMAL,
+        parameters: InfrahubObjectParameters = InfrahubObjectParameters(),
```
ogenstad: I'd suggest that we change this to:
```python
parameters: InfrahubObjectParameters | None= None
```
Then in the method code:

```python
parameters = parameters or InfrahubObjectParameters()
```

infrahub_sdk/spec/object.py
```diff
@@ -598,6 +523,7 @@ async def create_related_nodes(
         context: dict | None = None,
         branch: str | None = None,
         default_schema_kind: str | None = None,
+        parameters: InfrahubObjectParameters = InfrahubObjectParameters(),
```
ogenstad: Same comment as above.

infrahub_sdk/spec/object.py
```diff
@@ -631,7 +558,7 @@ async def create_related_nodes(
                 rel_info.find_matching_relationship(peer_schema=peer_schema)
                 context.update(rel_info.get_context(value=parent_node.id))
 
-            expanded_data = expand_data_with_ranges(data=data["data"])
+            expanded_data = RangeExpandDataProcessor.expand_data_with_ranges(data=data["data"])
```
ogenstad: I might be missing something here, but are we always calling this method to `expand_data_with_ranges()`?  Aren't we expected to consider the `parameters` for the setting first?

infrahub_sdk/spec/processors/range_expand_processor.py
minitriga: Resolved this

infrahub_sdk/spec/processors/data_processor.py
minitriga: Issue created for this: https://github.com/opsmill/infrahub-sdk-python/issues/589

infrahub_sdk/spec/object.py
minitriga: DataProcessorFactory.process_data does not always process the data it will first evaluate if the user has configured a parameter before it gets applied. It only applies the range expansion if parameters.expand_range is True. If parameters.expand_range is False (the default), the data is returned unchanged—no expansion happens. So, the parameters are always considered first before any processing is done.

Doing it this way means we dont have to check each time we evaluate the data which is in multiple places we can handle the check in a single place which is in that function.

ogenstad (approved):

Issue #380, opened 2025-04-29
bug: Name of protocol sync classes rendered incorrectly


ogenstad:
### Component

Python SDK

### Infrahub SDK version

v1.11.1

### Current Behavior

When trying to use the protocols together with the sync client we run into a problem due to the names of the classes.

### Expected Behavior

This should not be a problem.

### Steps to Reproduce

Run this snippet:

```python

from infrahub_sdk import InfrahubClientSync
from infrahub_sdk.protocols import BuiltinTagSync

client = InfrahubClientSync()
tags = client.all(kind=BuiltinTagSync)
```

The error will be:

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/patrick/Code/opsmill/infrahub-sdk-python/infrahub_sdk/client.py", line 1779, in all
    return self.filters(
           ^^^^^^^^^^^^^
  File "/Users/patrick/Code/opsmill/infrahub-sdk-python/infrahub_sdk/client.py", line 1915, in filters
    schema = self.schema.get(kind=kind, branch=branch)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/Code/opsmill/infrahub-sdk-python/infrahub_sdk/schema/__init__.py", line 563, in get
    raise SchemaNotFoundError(identifier=kind_str)
infrahub_sdk.exceptions.SchemaNotFoundError: Unable to find the schema 'BuiltinTagSync'.
```

The problematic call comes from this method in `client.schema``

```python
    @staticmethod
    def _get_schema_name(schema: type[SchemaType | SchemaTypeSync] | str) -> str:
        if hasattr(schema, "_is_runtime_protocol") and schema._is_runtime_protocol:  # type: ignore[union-attr]
            return schema.__name__  # type: ignore[union-attr]

        if isinstance(schema, str):
            return schema

        raise ValueError("schema must be a protocol or a string")
```

PR #586, opened 2025-10-28
IHS-128: Replace `Sync` in protocol sync classes schema name

solababs:
Fixes https://github.com/opsmill/infrahub-sdk-python/issues/380

This PR replaces the `Sync` word in the protocol schema name so that the correct kind can be gotten from the cache.

## Summary by CodeRabbit

* **Changelog**
  * Added an entry noting a protocol schema name change related to "Sync" handling.

* **Bug Fixes**
  * Fixed protocol schema name resolution so the correct schema kind is retrieved from cache (handles names ending with "Sync").

* **Tests**
  * Added unit tests validating schema name resolution for protocol-based schemas.

### DIFF
--- a/infrahub_sdk/schema/__init__.py
+++ b/infrahub_sdk/schema/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import inspect
 import json
 import warnings
 from collections.abc import MutableMapping
@@ -185,11 +186,15 @@ class InfrahubSchemaBase:
 
     @staticmethod
     def _get_schema_name(schema: type[SchemaType | SchemaTypeSync] | str) -> str:
-        if hasattr(schema, "_is_runtime_protocol") and schema._is_runtime_protocol:  # type: ignore[union-attr]
-            return schema.__name__.replace("Sync", "")  # type: ignore[union-attr]
-
         if isinstance(schema, str):
             return schema
+
+        if hasattr(schema, "_is_runtime_protocol") and getattr(schema, "_is_runtime_protocol", None):
+            if inspect.iscoroutinefunction(schema.save):
+                return schema.__name__
+            if schema.__name__[-4:] == "Sync":
+                return schema.__name__[:-4]
+            return schema.__name__
 
         raise ValueError("schema must be a protocol or a string")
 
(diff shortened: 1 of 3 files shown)


infrahub_sdk/schema/__init__.py
```diff
@@ -186,7 +186,7 @@ def _validate_load_schema_response(response: httpx.Response) -> SchemaLoadRespon
     @staticmethod
     def _get_schema_name(schema: type[SchemaType | SchemaTypeSync] | str) -> str:
         if hasattr(schema, "_is_runtime_protocol") and schema._is_runtime_protocol:  # type: ignore[union-attr]
-            return schema.__name__  # type: ignore[union-attr]
+            return schema.__name__.replace("Sync", "")  # type: ignore[union-attr]
```
ogenstad: Hey looking at this method I realize that the order and typing is a bit weird, we could get rid of the type ignores if we checked the string first and made a small change like this:

```python
        if isinstance(schema, str):
            return schema

        if hasattr(schema, "_is_runtime_protocol") and getattr(schema, "_is_runtime_protocol", None):
            return schema.__name__
```

For the fix I think this would be fine for most usecases, however it could be problematic depending on what users have called their schema nodes. I.e It's possible that someone has a node called `ProccessSyncBatch` then we'd rename it here so that it would be called `ProcessBatch` instead.

Even though it might be a bit more verbose perhaps we can have something like:

```python
import inspect

if inspect.iscoroutinefunction(schema.save):
    return schema.__name__
else:
    # Remove trailing 'Sync' suffix
    return schema.__name__[:-4]
```

I think it would be good with some unittests when selecting some of the classes from ./protocols.py to avoid future changes to this method if we break the behaviour.

tests/unit/sdk/test_schema.py
```diff
@@ -452,3 +453,8 @@ async def test_display_schema_load_errors_details_when_error_is_in_attribute_or_
   Node: SecurityTailscaleSSHRule | Attribute: check_period (10080) | Extra inputs are not permitted (extra_forbidden)
 """
         assert output == expected_console
+
+
+def test_schema_base__get_schema_name__returns_correct_schema_name_for_protocols():
+    assert InfrahubSchemaBase._get_schema_name(schema=BuiltinTagSync) == "BuiltinTag"
+    assert InfrahubSchemaBase._get_schema_name(schema=BuiltinIPAddressSync) == "BuiltinIPAddress"
```
ogenstad, line 462: Can you also add the same check for `BuiltinTag` and `BuiltinIPAddress` without the syncs, and additionally also add them as pure strings? I.e. so that we have six asserts?

ogenstad (approved):
Thanks!

PR #600, opened 2025-11-05
Add `graph_version` and `status` to branch

gmazoyer:
This PR simply adds the `graph_version` and `status` properties on `Branch` as they will be available in Infrahub 1.5.

## Summary by CodeRabbit

* **New Features**
  * Branch list/table now shows Graph Version and Status (Open, Needs Rebase, Needs Upgrade Rebase, Deleting).
* **Bug Fixes**
  * Non-default branches now display their own Graph Version and Status instead of inheriting the default branch value.
* **Tests**
  * Test fixtures updated to include Graph Version and Status for branch data.

### DIFF
--- a/infrahub_sdk/branch.py
+++ b/infrahub_sdk/branch.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import warnings
-from enum import StrEnum
+from enum import Enum
 from typing import TYPE_CHECKING, Any, Literal, overload
 from urllib.parse import urlencode
 
@@ -15,7 +15,7 @@ if TYPE_CHECKING:
     from .client import InfrahubClient, InfrahubClientSync
 
 
-class BranchStatus(StrEnum):
+class BranchStatus(str, Enum):
     OPEN = "OPEN"
     NEED_REBASE = "NEED_REBASE"
     NEED_UPGRADE_REBASE = "NEED_UPGRADE_REBASE"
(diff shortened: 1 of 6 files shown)


infrahub_sdk/branch.py
```diff
@@ -21,6 +21,7 @@ class BranchData(BaseModel):
     sync_with_git: bool
     is_default: bool
     has_schema_changes: bool
+    graph_version: int | None = None
```
ogenstad, line 32: We should probably also add `status`.

infrahub_sdk/branch.py
```diff
@@ -22,6 +30,7 @@ class BranchData(BaseModel):
     is_default: bool
     has_schema_changes: bool
     graph_version: int | None = None
+    status: BranchStatus
```
ogenstad: I wonder if we should have this default to `BranchStatus.OPEN` without something like that I think it could be problematic to use the SDK with earlier versions of Infrahub?

ogenstad (approved):

infrahub_sdk/branch.py
ogenstad: Ah ok, we get that regardless with the query as seen in the integration tests..

infrahub_sdk/ctl/branch.py
```diff
@@ -46,6 +46,8 @@ async def list_branch(_: str = CONFIG_PARAM) -> None:
     table.add_column("Sync with Git")
     table.add_column("Has Schema Changes")
     table.add_column("Is Default")
+    table.add_column("Graph Version")
```
dgarros: I don't think we should show the graph version by default to the user, it's an internal information a user won't need 99% of the time. 
If we really want to display that in the cli it should be required to specific the highest level of logs or similar

PR #603, opened 2025-11-07
IHS-176: Remove `is_visible` property from attribute metadata

solababs:
This PR removes `is_visible` property from attribute metadata.

## Summary by CodeRabbit

* **Refactor**
  * Removed the is_visible attribute across the SDK — it no longer appears on attribute instances, in recognized property flags, payload generation, or produced node/relationship outputs.
* **Tests**
  * Updated unit tests and fixtures to remove expectations and sample data containing is_visible.
* **Changelog**
  * Added entry documenting the removal of the is_visible property.

### DIFF
--- a/changelog/+08aa1a85.removed.md
+++ b/changelog/+08aa1a85.removed.md
@@ -1,0 +1,1 @@
+Remove `is_visible` property from infrahub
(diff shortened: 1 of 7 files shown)


gmazoyer:
This PR should target `infrahub-develop` instead of `develop`.

ogenstad:
> This PR should target `infrahub-develop` instead of `develop`.

I think either or would be fine in this case as IS_VISIBLE was never really used and even if you update the property it doesn't do anything sensible.

ajtmccarty (approved):

ogenstad:
@solababs, can we add a news fragment of the "removed" kind and just mention that the SDK will no longer query for the is_visible property as that's being phased out of Infrahub?

ajtmccarty (approved):

PR #625, opened 2025-11-12
Fix bug with nested object template range expansion

### LINKED ISSUE #624: bug: Object files broken
### Component

infrahubctl

### Infrahub SDK version

1.15.0

### Current Behavior

Currently with the following object file:

```

apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: TemplateDcimDevice
  data:
    # ---------------- Core Router Template ----------------
    - template_name: test-device-template
      device_type: Cisco ASR 1006-X
      role: core
      status: active
      interfaces:
        kind: TemplateInterfacePhysical
        parameters:
          expand_range: true
        data:
          - template_name: t-device-temp_TenGigE0/0/[1-20]
            name: TenGigE0/0/[1-20]
            role: upstream
            index: 1 
```

the range expansion is not working correctly and fails to expand nested objects. 

### Expected Behavior

I want to expand the nested objects and I would expect that this works. 

### Steps to Reproduce

```
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: TemplateDcimDevice
  data:
    # ---------------- Core Router Template ----------------
    - template_name: test-device-template
      device_type: Cisco ASR 1006-X
      role: core
      status: active
      interfaces:
        kind: TemplateInterfacePhysical
        parameters:
          expand_range: true
        data:
          - template_name: t-device-temp_TenGigE0/0/[1-20]
            name: TenGigE0/0/[1-20]
            role: upstream
            index: 1 
```

`infrahubctl object load test.yml`

### Additional Information

_No response_

minitriga:
Fixes: #624 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed nested object template range expansion so per-item embedded parameters are respected when creating related objects, ensuring parameter overrides apply correctly and validation/expansion behave consistently.

* **Tests**
  * Added tests for parameter handling in nested objects: top-level, missing, empty, and invalid (non-dictionary) parameter cases to prevent regressions.

### DIFF
--- a/infrahub_sdk/spec/object.py
+++ b/infrahub_sdk/spec/object.py
@@ -459,7 +459,6 @@ class InfrahubObjectFileData(BaseModel):
                             data=value,
                             branch=branch,
                             default_schema_kind=default_schema_kind,
-                            parameters=parameters,
                         )
                         clean_data[key] = nodes[0]
 
@@ -471,7 +470,9 @@ class InfrahubObjectFileData(BaseModel):
                             data=value,
                             branch=branch,
                             default_schema_kind=default_schema_kind,
-                            parameters=parameters,
+                            parameters=InfrahubObjectParameters(**value.get("parameters"))
+                            if "parameters" in value
+                            else None,
                         )
                         clean_data[key] = nodes
 
@@ -510,7 +511,9 @@ class InfrahubObjectFileData(BaseModel):
                 context=context,
                 branch=branch,
                 default_schema_kind=default_schema_kind,
-                parameters=InfrahubObjectParameters(**data[rel].get("parameters")) if "parameters" in data[rel] else None,
+                parameters=InfrahubObjectParameters(**data[rel].get("parameters"))
+                if "parameters" in data[rel]
+                else None,
             )
 
         return node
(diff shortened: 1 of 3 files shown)


dgarros (approved):

infrahub_sdk/spec/object.py
```diff
@@ -509,7 +510,7 @@ async def create_node(
                 context=context,
                 branch=branch,
                 default_schema_kind=default_schema_kind,
-                parameters=parameters,
+                parameters=InfrahubObjectParameters(**data[rel].get("parameters")) if "parameters" in data[rel] else None,
```
ajtmccarty: should the `None` be `parameters` here? I'm not sure how parameters passed into `create_node` are expected to behave in combination with parameters defined inside of `data`

ajtmccarty: should the other calls to `create_related_nodes` in this method do something similar with overriding parameters?

minitriga: If none is passed to the function it will create a new instance of `InfrahubObjectParameters` with `parameters = parameters or InfrahubObjectParameters()`

minitriga: You're correct. There is one other instance where range expansion is possible. I have added that but in the case where `RelationshipDataFormat.ONE_OBJ` is the relation there is no need for parameters as you cant expand into a cardinality of ONE.

PR #515, opened 2025-08-27
IHS-152 Allow unsetting one-to-one relationship

gmazoyer:
This change fixes the use of `my_object.my_relationship = None` by ensuring this is removing the link between an object and a related one tied together by an optional relationship of cardinality one.

## Summary by CodeRabbit

- New Features
  - Allow unsetting optional one-to-one relationships by assigning None.
- Bug Fixes
  - Correctly serialize None to GraphQL null, preventing incorrect “None” strings.
- Tests
  - Added integration coverage to verify nullifying a HFID-based relationship removes the link.
  - Updated unit tests to expect None for optional relationships in generated input data.
- Documentation
  - Added changelog entry describing the ability to unset optional one-to-one relationships via None.

PR #631, opened 2025-11-13
Prevent InfrahubNode from instantiating optional relationships of car…

### LINKED ISSUE #630: bug: Creating an object with an object template with optional attributes fails to set the optional attributes
### Component

Python SDK

### Infrahub SDK version

1.15.1

### Current Behavior

Currently if you have a optional relationship of cardinality one and try to use the object templates the SDK will provide an override value to the backend this causes the backend not to apply the value provided by object template.

The UI provides a mutation like so:

```graphql
mutation {
  DcimDeviceCreate(
    data: {object_template: {id: "18779f71-9564-3911-3bc2-1065c8cd8363"}, name: {value: "123123"}, location: {id: "18779f71-e8aa-55c1-3bc0-1065ebd40353"}}
  ) {
    object {
      id
      display_label
      hfid
      __typename
    }
    __typename
  }
}
```

However the SDK would provide a graphql mutation which looks like the following:

```graphql
mutation {
    DcimDeviceUpsert(
        data: {
            status: {
                value: "active"
            }
            name: {
                value: "123123"
            }
            index: {
                value: 1
            }
            mlag_domain: null
            object_template: {
                id: "18779f71-55f1-390f-3bcf-10652e6ca9d3"
            }
            platform: {
                id: "test"
            }
            primary_address: null
            location: {
                id: "18779f72-1586-2058-3bc3-106527497af8"
            }
            device_type: null
        }
    ){
        ok
        object {
            id
        }
    }
}
```

You can see there are many null fields being passed in. This also happens when using object files to load data. 

### Expected Behavior

I would expect optional fields to be ignored and not instantiated by the client when creating a node the backend does not require them so no need to include them.

### Steps to Reproduce

1. Load infrahub with the base schema from the Schema library https://github.com/opsmill/schema-library/blob/main/base/dcim.yml#L87
2. Create some manufactures and device_types
3. Create a device type using an object file

```
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: DcimDevice
  data:
    # Paris HQ Building - Cisco Core Routers
    - name: CoreRouter1
      object_template: CISCO-ASR1006-CORE
      location: hq-01
      status: active
      index: 1
    - name: CoreRouter2
      object_template: CISCO-ASR1006-CORE
      location: hq-01
      status: active
      index: 2
```

### Additional Information

_No response_

minitriga:
…dinality one with None data.

Fixes: #630 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of optional one-to-one relationships. Uninitialized optional relationships are now properly excluded from mutation requests instead of being set to None.

### DIFF
--- a/changelog/630.fixed.md
+++ b/changelog/630.fixed.md
@@ -1,0 +1,1 @@
+Prevent InfrahubNode from instantiating optional relationships of cardinality one with None data.
--- a/infrahub_sdk/node/node.py
+++ b/infrahub_sdk/node/node.py
@@ -235,7 +235,6 @@ class InfrahubNodeBase:
             rel: RelatedNodeBase | RelationshipManagerBase = getattr(self, item_name)
 
             if rel_schema.cardinality == RelationshipCardinality.ONE and rel_schema.optional and not rel.initialized:
-                data[item_name] = None
                 continue
 
             if rel is None or not rel.initialized:
--- a/tests/unit/sdk/test_node.py
+++ b/tests/unit/sdk/test_node.py
@@ -1365,7 +1365,6 @@ async def test_create_input_data(client, location_schema: NodeSchemaAPI, client_
             "name": {"value": "JFK1"},
             "description": {"value": "JFK Airport"},
             "type": {"value": "SITE"},
-            "primary_tag": None,
         }
     }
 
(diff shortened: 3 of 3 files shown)


ogenstad (changes requested):
We need to consider this a bit more, it looks like we're reverting the things done in #515 with this PR. So we need to take a closer look at that's going on here.

minitriga:
Closing for now as I have a work around.

PR #637, opened 2025-11-17
Add `infrahubctl branch report` command

wvandeun:
## Summary by CodeRabbit

* **New Features**
  * Public API to fetch full diff trees with metadata
  * Branch report CLI command with detailed, table-based cleanup/status reports
  * Human-readable timestamp formatting and git-change detection in reports

* **Bug Fixes**
  * Always include properties and updated timestamps in node/relationship outputs
  * Expanded GraphQL variable handling to support nullable types and DateTime
  * Variable rendering now accepts date/time values

* **Tests**
  * Extended coverage for diff trees, branch reporting, and deep relationship processing

* **Documentation**
  * Added docs for the branch report command

### DIFF
--- a/infrahub_sdk/client.py
+++ b/infrahub_sdk/client.py
@@ -34,7 +34,7 @@ from .config import Config
 from .constants import InfrahubClientMode
 from .convert_object_type import CONVERT_OBJECT_MUTATION, ConversionFieldInput
 from .data import RepositoryBranchInfo, RepositoryData
-from .diff import NodeDiff, diff_tree_node_to_node_diff, get_diff_summary_query
+from .diff import DiffTreeData, NodeDiff, diff_tree_node_to_node_diff, get_diff_summary_query, get_diff_tree_query
 from .exceptions import (
     AuthenticationError,
     Error,
@@ -1282,6 +1282,64 @@ class InfrahubClient(BaseClient):
 
         return node_diffs
 
+    async def get_diff_tree(
+        self,
+        branch: str,
+        name: str | None = None,
+        from_time: datetime | None = None,
+        to_time: datetime | None = None,
+        timeout: int | None = None,
+        tracker: str | None = None,
+        raise_for_error: bool | None = None,
+    ) -> DiffTreeData | None:
+        """Get complete diff tree with metadata and nodes.
+
+        Returns None if no diff exists.
+        """
+        query = get_diff_tree_query()
+        input_data = {"branch_name": branch}
+        if name:
+            input_data["name"] = name
+        if from_time and to_time and from_time > to_time:
+            raise ValueError("from_time must be <= to_time")
+        if from_time:
+            input_data["from_time"] = from_time.isoformat()
+        if to_time:
+            input_data["to_time"] = to_time.isoformat()
+
+        response = await self.execute_graphql(
+            query=query.render(),
+            branch_name=branch,
+            timeout=timeout,
+            tracker=tracker,
+            raise_for_error=raise_for_error,
+            variables=input_data,
+        )
+
+        diff_tree = response["DiffTree"]
+        if diff_tree is None:
+            return None
+
+        # Convert nodes to NodeDiff objects
+        node_diffs: list[NodeDiff] = []
+        if "nodes" in diff_tree:
+            for node_dict in diff_tree["nodes"]:
+                node_diff = diff_tree_node_to_node_diff(node_dict=node_dict, branch_name=branch)
+                node_diffs.append(node_diff)
+
+        return DiffTreeData(
+            num_added=diff_tree.get("num_added") or 0,
+            num_updated=diff_tree.get("num_updated") or 0,
+            num_removed=diff_tree.get("num_removed") or 0,
+            num_conflicts=diff_tree.get("num_conflicts") or 0,
+            to_time=diff_tree["to_time"],
+            from_time=diff_tree["from_time"],
+            base_branch=diff_tree["base_branch"],
+            diff_branch=diff_tree["diff_branch"],
+            name=diff_tree.get("name"),
+            nodes=node_diffs,
+        )
+
     @overload
     async def allocate_next_ip_address(
         self,
@@ -2519,6 +2577,64 @@ class InfrahubClientSync(BaseClient):
             node_diffs.append(node_diff)
 
         return node_diffs
+
+    def get_diff_tree(
+        self,
+        branch: str,
+        name: str | None = None,
+        from_time: datetime | None = None,
+        to_time: datetime | None = None,
+        timeout: int | None = None,
+        tracker: str | None = None,
+        raise_for_error: bool | None = None,
+    ) -> DiffTreeData | None:
+        """Get complete diff tree with metadata and nodes.
+
+        Returns None if no diff exists.
+        """
+        query = get_diff_tree_query()
+        input_data = {"branch_name": branch}
+        if name:
+            input_data["name"] = name
+        if from_time and to_time and from_time > to_time:
+            raise ValueError("from_time must be <= to_time")
+        if from_time:
+            input_data["from_time"] = from_time.isoformat()
+        if to_time:
+            input_data["to_time"] = to_time.isoformat()
+
+        response = self.execute_graphql(
+            query=query.render(),
+            branch_name=branch,
+            timeout=timeout,
+            tracker=tracker,
+            raise_for_error=raise_for_error,
+            variables=input_data,
+        )
+
+        diff_tree = response["DiffTree"]
+        if diff_tree is None:
+            return None
+
+        # Convert nodes to NodeDiff objects
+        node_diffs: list[NodeDiff] = []
+        if "nodes" in diff_tree:
+            for node_dict in diff_tree["nodes"]:
+                node_diff = diff_tree_node_to_node_diff(node_dict=node_dict, branch_name=branch)
+                node_diffs.append(node_diff)
+
+        return DiffTreeData(
+            num_added=diff_tree.get("num_added") or 0,
+            num_updated=diff_tree.get("num_updated") or 0,
+            num_removed=diff_tree.get("num_removed") or 0,
+            num_conflicts=diff_tree.get("num_conflicts") or 0,
+            to_time=diff_tree["to_time"],
+            from_time=diff_tree["from_time"],
+            base_branch=diff_tree["base_branch"],
+            diff_branch=diff_tree["diff_branch"],
+            name=diff_tree.get("name"),
+            nodes=node_diffs,
+        )
 
     @overload
     def allocate_next_ip_address(
--- a/infrahub_sdk/client.py
+++ b/infrahub_sdk/client.py
@@ -1290,7 +1290,6 @@ class InfrahubClient(BaseClient):
         to_time: datetime | None = None,
         timeout: int | None = None,
         tracker: str | None = None,
-        raise_for_error: bool | None = None,
     ) -> DiffTreeData | None:
         """Get complete diff tree with metadata and nodes.
 
@@ -1312,7 +1311,6 @@ class InfrahubClient(BaseClient):
             branch_name=branch,
             timeout=timeout,
             tracker=tracker,
-            raise_for_error=raise_for_error,
             variables=input_data,
         )
 
@@ -2586,7 +2584,6 @@ class InfrahubClientSync(BaseClient):
         to_time: datetime | None = None,
         timeout: int | None = None,
         tracker: str | None = None,
-        raise_for_error: bool | None = None,
     ) -> DiffTreeData | None:
         """Get complete diff tree with metadata and nodes.
 
@@ -2608,7 +2605,6 @@ class InfrahubClientSync(BaseClient):
             branch_name=branch,
             timeout=timeout,
             tracker=tracker,
-            raise_for_error=raise_for_error,
             variables=input_data,
         )
 
--- a/infrahub_sdk/ctl/branch.py
+++ b/infrahub_sdk/ctl/branch.py
@@ -1,4 +1,5 @@
 import logging
+import sys
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING
 
@@ -194,6 +195,10 @@ async def report(  # noqa: PLR0915
     _: str = CONFIG_PARAM,
 ) -> None:
     """Generate branch cleanup status report."""
+
+    if branch_name == "main":
+        console.print("[red]Cannot create a report for the main branch!")
+        sys.exit(1)
 
     client = initialize_client()
 
--- a/infrahub_sdk/ctl/branch.py
+++ b/infrahub_sdk/ctl/branch.py
@@ -11,6 +11,7 @@ from ..async_typer import AsyncTyper
 from ..utils import calculate_time_diff, decode_json
 from .client import initialize_client
 from .parameters import CONFIG_PARAM
+from ..protocols import CoreProposedChange
 from .utils import catch_exception
 
 if TYPE_CHECKING:
@@ -277,7 +278,10 @@ async def report(  # noqa: PLR0915
 
     # Fetch proposed changes for the branch
     proposed_changes = await client.filters(
-        kind="CoreProposedChange", source_branch__value=branch_name, include=["created_by"], prefetch_relationships=True
+        kind=CoreProposedChange,  # type: ignore[type-abstract]
+        source_branch__value=branch_name,
+        include=["created_by"],
+        prefetch_relationships=True
     )
 
     # Print proposed changes section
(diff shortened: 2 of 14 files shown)


wvandeun:
@coderabbitai review

wvandeun:
@coderabbitai review

infrahub_sdk/graphql/constants.py
```diff
@@ -1 +1,14 @@
...
+    (str | None, "String"),
+    (int, "Int!"),
+    (int | None, "Int"),
+    (float, "Float!"),
+    (float | None, "Float"),
+    (bool, "Boolean!"),
+    (bool | None, "Boolean"),
+    (datetime, "DateTime!"),
+    (datetime | None, "DateTime"),
+)
```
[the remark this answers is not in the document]
ogenstad: If this PR targets `infrahub-develop` instead we've removed Python 3.9 support in that branch so it would be better to retarget this to avoid linting violations if we reintroduce `Union`.

infrahub_sdk/graphql/constants.py
```diff
@@ -1,14 +1,15 @@
...
+    (Union[int, None], "Int"),
     (float, "Float!"),
-    (float | None, "Float"),
+    (Union[float, None], "Float"),
     (bool, "Boolean!"),
-    (bool | None, "Boolean"),
+    (Union[bool, None], "Boolean"),
     (datetime, "DateTime!"),
-    (datetime | None, "DateTime"),
+    (Union[datetime, None], "DateTime"),
```
ogenstad: Hey Wim, I commented on the comment that CodeRabbit added. The better approach here would be to target the `infrahub-develop` branch where Python 3.9 is removed. So we don't want the old style `Union` here as it would cause linting issues in the other branch.

.github/workflows/publish-pypi.yml
```diff
@@ -41,39 +41,36 @@ jobs:
...
-      - name: "show output"
-        run: "ls -la dist/"
+      - name: Show output
+        run: ls -la dist/
 
-      - name: "Poetry push PyPI"
+      - name: Publish to PyPI
         if: ${{ inputs.publish }}
-        run: "poetry publish"
+        run: uv publish
```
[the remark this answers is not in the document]
fatih-acar: Fixed in https://github.com/opsmill/infrahub-sdk-python/pull/639

infrahub_sdk/ctl/branch.py
```diff
@@ -143,3 +185,115 @@ async def validate(branch_name: str, _: str = CONFIG_PARAM) -> None:
...
+            proposal_table = Table(show_header=False, box=None)
+            proposal_table.add_column(justify="left")
+            proposal_table.add_column(justify="right")
+
+            # Extract data from node
+            proposal_table.add_row("Name", pc.name.value)  # type: ignore[union-attr]
+            proposal_table.add_row("State", str(pc.state.value))  # type: ignore[union-attr]
+            proposal_table.add_row("Is draft", "Yes" if pc.is_draft.value else "No")  # type: ignore[union-attr]
+            proposal_table.add_row("Created by", pc.created_by.peer.name.value)  # type: ignore[union-attr]
+            proposal_table.add_row("Created at", format_timestamp(str(pc.created_by.updated_at)))  # type: ignore[union-attr]
```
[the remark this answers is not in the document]
wvandeun: This is a workaround that was implemented because we don't have easy access to the creation time of the proposed change in the SDK. This will change as soon as we implemented object level meta data in the 1.7 release of Infrahub.

infrahub_sdk/client.py
```diff
@@ -1282,6 +1282,64 @@ async def get_diff_summary(
...
 
+    async def get_diff_tree(
+        self,
+        branch: str,
+        name: str | None = None,
+        from_time: datetime | None = None,
+        to_time: datetime | None = None,
+        timeout: int | None = None,
+        tracker: str | None = None,
+        raise_for_error: bool | None = None,
```
ogenstad: Will try to take a closer look tomorrow but `raise_for_error` is deprecated in the execute_graphql method. So we should not add that anywhere new.

infrahub_sdk/client.py
```diff
@@ -1282,6 +1282,62 @@ async def get_diff_summary(
 
         return node_diffs
 
+    async def get_diff_tree(
```
ogenstad, line 1285: Do you see a value from an SDK point of view to have this as a method directly under the client? I don't have a strong objection to it but the client file is already massive as it is and at some point I'd like to break it apart into smaller components. Don't know if this command would remain in the same place at that point. Not really ready for that refactoring just now. But if we don't see a clear use case to have this within the SDK itself I'd suggest that we start out by just having it in the CTL.

infrahub_sdk/client.py
```diff
@@ -1282,6 +1282,62 @@ async def get_diff_summary(
...
+        if "nodes" in diff_tree:
+            for node_dict in diff_tree["nodes"]:
+                node_diff = diff_tree_node_to_node_diff(node_dict=node_dict, branch_name=branch)
+                node_diffs.append(node_diff)
+
+        return DiffTreeData(
+            num_added=diff_tree.get("num_added") or 0,
+            num_updated=diff_tree.get("num_updated") or 0,
+            num_removed=diff_tree.get("num_removed") or 0,
+            num_conflicts=diff_tree.get("num_conflicts") or 0,
```
ogenstad, line 1332: Looking at this I think it comes from the fact that the GraphQL query itself has an incorrect return format. I.e. all of these should be required:

https://github.com/opsmill/infrahub/blob/infrahub-v1.5.3/backend/infrahub/graphql/queries/diff/tree.py#L62-L66

I.e. it should never return a `null` value like it looks like the query can do now. Based on the issue in Infrahub I think your approach would be the way to go. But we can create a followup task for it later to clean this up.

infrahub_sdk/ctl/branch.py
```diff
@@ -18,6 +24,42 @@
 ENVVAR_CONFIG_FILE = "INFRAHUBCTL_CONFIG"
 
 
+def format_timestamp(timestamp: str) -> str:
+    """Format ISO timestamp to 'YYYY-MM-DD HH:MM:SS'."""
+    try:
+        dt = datetime.fromisoformat(timestamp.replace("Z", "+00:00"))
+        return dt.strftime("%Y-%m-%d %H:%M:%S")
+    except (ValueError, AttributeError):
+        return timestamp
```
ogenstad: This looks like a safeguard incase the response from Infrahub is invalid. With the docstring it's not clear that an incorrect value could be returned.

Is the issue here that we ignore typing somewhere and `timestamp` can be `None`?

infrahub_sdk/ctl/branch.py
```diff
@@ -143,3 +185,115 @@ async def validate(branch_name: str, _: str = CONFIG_PARAM) -> None:
...
+            proposal_table.add_column(justify="right")
+
+            # Extract data from node
+            proposal_table.add_row("Name", pc.name.value)  # type: ignore[union-attr]
+            proposal_table.add_row("State", str(pc.state.value))  # type: ignore[union-attr]
+            proposal_table.add_row("Is draft", "Yes" if pc.is_draft.value else "No")  # type: ignore[union-attr]
+            proposal_table.add_row("Created by", pc.created_by.peer.name.value)  # type: ignore[union-attr]
+            proposal_table.add_row("Created at", format_timestamp(str(pc.created_by.updated_at)))  # type: ignore[union-attr]
+            proposal_table.add_row("Approvals", str(len(pc.approved_by.peers)))  # type: ignore[union-attr]
+            proposal_table.add_row("Rejections", str(len(pc.rejected_by.peers)))  # type: ignore[union-attr]
```
ogenstad: We should avoid adding new ignore statements for typing. I'm guessing that these could be removed if we instead used the protocol definition and not the string `"CoreProposedChange"` when querying for these objects.

infrahub_sdk/ctl/branch.py
```diff
@@ -143,3 +185,115 @@ async def validate(branch_name: str, _: str = CONFIG_PARAM) -> None:
...
+            # Create proposal table
+            proposal_table = Table(show_header=False, box=None)
+            proposal_table.add_column(justify="left")
+            proposal_table.add_column(justify="right")
+
+            # Extract data from node
+            proposal_table.add_row("Name", pc.name.value)  # type: ignore[union-attr]
+            proposal_table.add_row("State", str(pc.state.value))  # type: ignore[union-attr]
+            proposal_table.add_row("Is draft", "Yes" if pc.is_draft.value else "No")  # type: ignore[union-attr]
+            proposal_table.add_row("Created by", pc.created_by.peer.name.value)  # type: ignore[union-attr]
```
ogenstad: Based on the changes for immutable metadata in Infrahub 1.7, I'd consider waiting to include the creator of the PC.

infrahub_sdk/ctl/branch.py
```diff
@@ -143,3 +185,115 @@ async def validate(branch_name: str, _: str = CONFIG_PARAM) -> None:
...
+async def report(  # noqa: PLR0915
+    branch_name: str = typer.Argument(..., help="Branch name to generate report for"),
+    update_diff: bool = typer.Option(False, "--update-diff", help="Update diff before generating report"),
+    _: str = CONFIG_PARAM,
+) -> None:
+    """Generate branch cleanup status report."""
+
+    if branch_name == "main":
+        console.print("[red]Cannot create a report for the main branch!")
+        sys.exit(1)
```
ogenstad: While the main branch should be standard everywhere, the approach seems wrong here. I'd instead let the SDK query for the branch and then check the object on line 206 and if the returned object is the default branch raise an error there.

infrahub_sdk/ctl/branch.py
```diff
@@ -143,3 +185,115 @@ async def validate(branch_name: str, _: str = CONFIG_PARAM) -> None:
     client = initialize_client()
     await client.branch.validate(branch_name=branch_name)
     console.print(f"Branch '{branch_name}' is valid.")
+
+
+@app.command()
+@catch_exception(console=console)
+async def report(  # noqa: PLR0915
```
ogenstad: I'd prefer to now add ignore statements for complexity like this, we already have lots of them that we should clean up at some point. :( 

If we look at a snippet like this:
```python

    # Fetch diff tree (with metadata)
    diff_tree = await client.get_diff_tree(branch=branch_name)

    if diff_tree:
        branch_table.add_row("Diff last updated", format_timestamp(diff_tree["to_time"]))
        branch_table.add_row("Amount of additions", str(diff_tree["num_added"]))
        branch_table.add_row("Amount of deletions", str(diff_tree["num_removed"]))
        branch_table.add_row("Amount of updates", str(diff_tree["num_updated"]))
        branch_table.add_row("Amount of conflicts", str(diff_tree["num_conflicts"]))
    else:
        branch_table.add_row("Diff last updated", "No diff available")
        branch_table.add_row("Amount of additions", "-")
        branch_table.add_row("Amount of deletions", "-")
        branch_table.add_row("Amount of updates", "-")
        branch_table.add_row("Amount of conflicts", "-")
```

This looke like it could live in a function that would return a list of the rows that can them be added to the table.

infrahub_sdk/node/related_node.py
```diff
@@ -64,7 +64,7 @@ def __init__(self, branch: str, schema: RelationshipSchemaAPI, data: Any | dict,
                 self._display_label = node_data.get("display_label", None)
                 self._typename = node_data.get("__typename", None)
 
-            self.updated_at: str | None = data.get("updated_at", data.get("_relation__updated_at"))
+            self.updated_at: str | None = data.get("updated_at", properties_data.get("updated_at", None))
```
ogenstad, line 67: Was this intentional and a bug that needed to be fixed or do we have a response that looks like this from the backend?

infrahub_sdk/node/relationship.py
```diff
@@ -87,9 +87,8 @@ def _generate_query_data(cls, peer_data: dict[str, Any] | None = None, property:
                 properties[prop_name] = None
             for prop_name in PROPERTIES_OBJECT:
                 properties[prop_name] = {"id": None, "display_label": None, "__typename": None}
-
-        if properties:
```
ogenstad, line 91: Was this something that was required?

tests/unit/ctl/test_branch_report.py
```diff
@@ -0,0 +1,436 @@
...
+    assert result is False
+
+
+async def test_check_git_files_changed_http_error(httpx_mock: HTTPXMock) -> None:
+    """Test check_git_files_changed raises exception on HTTP error."""
+    httpx_mock.add_response(
+        method="GET",
+        url="http://mock/api/diff/files?branch=test-branch",
+        status_code=404,
+    )
```
ogenstad, line 99: I think this test is easier to read than the ones above, if we look at a fixture like `mock_git_files_empty_response` which is only used once and is so short I'd recommend having it inline within the test body as it's easier to see at a glance what's going on. If we know that we're going to use the fixture in multiple locations I think that's fine, but if they only are used once we also run the risk of leaving the fixture in place if the test is ever deleted.

tests/unit/ctl/test_branch_report.py
```diff
@@ -0,0 +1,436 @@
...
+        method="POST",
+        json={"data": {"CoreProposedChange": {"count": 0, "edges": []}}},
+    )
+
+    return httpx_mock
+
+
+def test_branch_report_command_without_proposed_change(
+    mock_branch_report_command: HTTPXMock,
+    mock_schema_query_05,  # type: ignore[misc]
```
ogenstad: Looks like this is a `HTTPXMock` class to, so it should be fine to add that and remove the ignore statement?

tests/unit/sdk/test_diff_summary.py
```diff
@@ -158,3 +161,239 @@ async def test_diffsummary(clients: BothClients, mock_diff_tree_query, client_ty
...
     } in node_diffs
+
+
+def test_get_diff_tree_query_structure() -> None:
+    """Test that get_diff_tree_query returns proper Query object."""
+    query = get_diff_tree_query()
+
+    # Verify it's a Query object
+    assert hasattr(query, "render")
+    assert hasattr(query, "name")
```
ogenstad: Typing should catch this and if the name or render function doesn't exist the asserts just below would fail. So we can remove the hasattr checks.

infrahub_sdk/ctl/branch.py
```diff
@@ -143,3 +185,115 @@ async def validate(branch_name: str, _: str = CONFIG_PARAM) -> None:
...
+
+    branch_table.add_row("Synced with Git", "Yes" if branch.sync_with_git else "No")
+
+    # Add Git files changed
+    if git_files_changed is not None:
+        branch_table.add_row("Git files changed", "Yes" if git_files_changed else "No")
+    else:
+        branch_table.add_row("Git files changed", "N/A")
+
+    branch_table.add_row("Has schema changes", "Yes" if branch.has_schema_changes else "No")
```
ogenstad: I think this part could potentially be a bit misleading for stale branches. The only thing this flag does is to indicate if the schema differs compared to the default branch. This field is probably incorrectly named within Infrahub and should be called something like "schema in sync with default branch". The problem here would be if the default branch has moved on with other changes.

Could be that this naming would remain consistent with what we have within Infrahub, but I think the proper way to do this would be to do a diff of the schema as well. This might also be enough for an initial round.

infrahub_sdk/client.py
wvandeun: Opened this task for this:
https://github.com/opsmill/infrahub/issues/7778

infrahub_sdk/client.py
wvandeun: Yes I do, today there is no way to get this kind of diff for a single branch in the UI. The only way you do it today would be to craft a GraphQL query yourself. IE I do think this is a good intermediate to be able to expose this easier.

infrahub_sdk/ctl/branch.py
wvandeun: Ok, let's revisit this later.

infrahub_sdk/node/related_node.py
wvandeun: Yes, this is intentional. There was an issue where the `updated_at` metadata for an attribute and relationship was not properly populated.

wvandeun:
```python
import infrahub_sdk
print(infrahub_sdk.__version__)
1.15.1

from infrahub_sdk import InfrahubClientSync, Config
config = Config(api_token="06438eb2-8019-4776-878c-0941b1f1d1ec")
client = InfrahubClientSync(config=config,address="https://sandbox.infrahub.app")

device = client.get("InfraDevice", name__value="atl1-edge1", include=["interfaces"], prefetch_relationships=True, property=True)
print(device.name.updated_at) # None
print(device.interfaces.peers[0].updated_at) # None
```

infrahub_sdk/node/relationship.py
wvandeun: Not necessary required. I just think this is cleaner. The previous approach was checking if we have value in the properties dict, whereas we are already checking if the properties should be queried using the property argument. This reads a bit cleaner, imo

PR #663, opened 2025-11-28
Python 3.14

ogenstad:
## Summary by CodeRabbit

* **Chores**
  * Added official Python 3.14 support and updated CI to test it.
  * Bumped a runtime dependency to a newer compatible release.

* **Tests**
  * Updated unit tests to align with revised datetime parsing/formatting behavior.


ogenstad:
Thanks @fatih-acar, we'd want to have a pipeline against the Infrahub repo as a test that includes this PR. The `whenever` package is also pinned in Infrahub and used directly in a small number of places. Could be that we don't need to update the code there but would be nice to verify.

fatih-acar:
I've done it in https://github.com/opsmill/infrahub/pull/7746
But I can do a separate PR containing only the SDK update?

ogenstad:
> I've done it in [opsmill/infrahub#7746](https://github.com/opsmill/infrahub/pull/7746) But I can do a separate PR containing only the SDK update?

I don't think a separate one is required. As long as we see that it doesn't break anything. Then once this PR is merged we should update the infrahub develop branch to use the latest `infrahub-develop` from the SDK.

ajtmccarty (approved):

PR #686, opened 2025-12-09
IHS-173 Detect if a relationship is set from a profile

gmazoyer:
This change adds the `is_from_profile` property to related nodes and relationship manager in order to check if a relationship has been set from a profile.

The check is performed based on the kind of the relationship source. If the kind of the source is a profile, we consider that a related node comes from a profile. For a relationship of cardinality many, we consider that it is set by a profile if all of its related nodes are set by a profile.

## Summary by CodeRabbit

* **New Features**
  * Related nodes gain an is_from_profile property to indicate if they originate from a profile.
  * Relationships gain an is_from_profile property that reports True only when all associated peers originate from profiles.

* **Tests**
  * Added unit tests covering is_from_profile behavior across related nodes and relationship scenarios.


ajtmccarty (approved):

infrahub_sdk/node/related_node.py
```diff
@@ -125,6 +126,13 @@ def kind(self) -> str | None:
             return self._peer.get_kind()
         return self._kind
 
+    @property
+    def is_from_profile(self) -> bool:
+        """Return whether this relationship was set from a profile. Done by checking if the source is of a profile kind."""
+        if not self._source_typename:
+            return False
+        return self._source_typename.startswith(PROFILE_KIND_PREFIX)
```
ogenstad: Small comment here, we could potentially make this a bit more safe where we'd have a regular expression instead of `PROFILE_KIND_PREFIX`. What we want to look for is any kind that starts with "Profile" and then has a capital letter following it as that would indicate the start of the real namespace.

I.e. if someone were to create a nodekind with:

* namespace: Profilebob
* name: Device

And this got used in a source location we'd mark it as coming from a profile even though the source was set to a real node type. I wouldn't imagine this coming up a lot but it's something that could happen.

gmazoyer: Changed to a regex pattern

ogenstad (approved):

PR #701, opened 2025-12-18
fix(analyzer): allowing passing a parsed query

fatih-acar:
This will be used within Infrahub to pass cached documents.

## Summary by CodeRabbit

* **Enhancements**
  * Query analyzer now supports accepting pre-parsed documents to improve processing efficiency and reduce redundant parsing operations.

### DIFF
--- a/infrahub_sdk/analyzer.py
+++ b/infrahub_sdk/analyzer.py
@@ -30,10 +30,10 @@ class GraphQLOperation(BaseModel):
 
 
 class GraphQLQueryAnalyzer:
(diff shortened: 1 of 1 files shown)


ogenstad (approved):
Good catch! I think it gets a bit weird that we still require the query string to be set when we pass in a document. But this is probably fine for now. In the near future I think we'll want to move some parts of the Infrahub Query Analyzer into the SDK to be able to analyze queries within the SDK itself to allow for that work to be done from infrahubctl. At that time we'll probably refactor this anyway.

PR #714, opened 2025-12-29
Fix issue #713 and #712

### LINKED ISSUE #712: bug: Provided name  cannot be used as python identifier.
### Component

_No response_

### Infrahub SDK version

1.17.0

### Current Behavior

When running the command `infrahubctl graphql generate-return-types query1.gql` I got the following exception:

```
Error: Provided name  cannot be used as python identifier.
    raise InvalidConfiguration(
ariadne_codegen.exceptions.InvalidConfiguration: Provided name  cannot be used as python identifier.
```

### Expected Behavior

I expect the command to work

### Steps to Reproduce

- Setup project
```
mkdir some-project
cd some-project
uv init
uv add 'infrahub-sdk[all]'
export INFRAHUB_API_TOKEN=06438eb2-8019-4776-878c-0941b1f1d1ec
```

- Create a query1.gql with the following content:

```gql
query Test {
  BuiltinTag {
    edges {
      node {
        name {
          value
        }
      }
    }
  }
}
```

- Run command `infrahubctl graphql export-schema`
- Then `infrahubctl graphql generate-return-types query1.gql`

=> This fails with `Error: Provided name  cannot be used as python identifier.` 

### Additional Information

The problem is that directory.name is empty when the GraphQL file is in the current directory.

https://github.com/opsmill/infrahub-sdk-python/blob/1f1cb0c94a3334ee7de0930dab80702bcae645ce/infrahub_sdk/ctl/graphql.py#L160

So either we force people to put their queries in a folder, or we adjust code to account for local folder.

### LINKED ISSUE #713: bug: TypeError: Redefinition of reserved type 'String' - Graphql generate-return-types
### Component

infrahubctl

### Infrahub SDK version

1.0.17

### Current Behavior

When running `infrahubctl graphql generate-return-types queries` I got the following exception: `Error: Redefinition of reserved type 'String'`

```
KeyError: '__typename'

During handling of the above exception, another exception occurred:
    raise TypeError(f"Redefinition of reserved type {name!r}")
TypeError: Redefinition of reserved type 'String'
```

### Expected Behavior

I expect this command to work

### Steps to Reproduce

- Setup project

```
mkdir some-project
cd some-project
mkdir queries
uv init
uv add 'infrahub-sdk[all]'
export INFRAHUB_API_TOKEN=06438eb2-8019-4776-878c-0941b1f1d1ec
```

- Add the following `query1.gql` in `queries` folder

```gql
query Test {
  BuiltinTag {
    edges {
      node {
        __typename
        name {
          value
        }
      }
    }
  }
}
```

- Run command `infrahubctl graphql export-schema`
- Then `infrahubctl graphql generate-return-types queries`

=> This will fail with error described above

- Remove `__typename` from query
- Then `infrahubctl graphql generate-return-types queries`

=> This works

### Additional Information

The __typename field is not explicitly defined in the GraphQL schema. In GraphQL, __typename is an introspection meta-field that's automatically available on all types - it doesn't need to be (and shouldn't be) explicitly defined in the schema. The problem is that ariadne-codegen is trying to look up __typename in the schema's type map and failing because it's not there (it's a built-in introspection field). When it fails to find it, it tries to create a fallback GraphQLScalarType(name="String"), but this causes the "Redefinition of reserved type 'String'" error because String is a reserved built-in scalar type in GraphQL. This is a bug in ariadne-codegen, not in your query or schema.

BaptisteGi:
* **Documentation**
  * Updated Python SDK guide with revised workflow examples for generating Pydantic models from GraphQL queries using a directory-based approach and enhanced step-by-step guidance.

* **Improvements for generate-return-types ctl command**
  * Support for __typename in queries
  * Support queries in current folder

## Summary by CodeRabbit

* **Documentation**
  * Updated Python SDK typing guide with a directory-based workflow, clearer step-by-step examples, schema export guidance, and a warning about unique query filenames.

* **Improvements**
  * Code generation now strips GraphQL __typename fields from operations and fragments before generating models.

* **Tests**
  * Added extensive unit and integration tests covering query discovery, parsing, generation flows, and typename-stripping behavior.

* **Fixtures / Schema**
  * Added GraphQL fixtures and a self-contained schema for tag-related types to support tests.

### DIFF
--- a/infrahub_sdk/graphql/utils.py
+++ b/infrahub_sdk/graphql/utils.py
@@ -1,5 +1,88 @@
 import ast
 
+from graphql import (
+    FieldNode,
+    FragmentDefinitionNode,
+    FragmentSpreadNode,
+    InlineFragmentNode,
+    OperationDefinitionNode,
+    SelectionNode,
+    SelectionSetNode,
+)
+
+
+def strip_typename_from_selection_set(selection_set: SelectionSetNode | None) -> SelectionSetNode | None:
+    """Recursively strip __typename fields from a SelectionSetNode.
+
+    The __typename meta-field is an introspection field that is not part of the schema's
+    type definitions. When code generation tools like ariadne-codegen try to look up
+    __typename in the schema, they fail because it's a reserved introspection field.
+
+    This function removes all __typename fields from the selection set, allowing
+    code generation to proceed without errors.
+    """
+    if selection_set is None:
+        return None
+
+    new_selections: list[SelectionNode] = []
+    for selection in selection_set.selections:
+        if isinstance(selection, FieldNode):
+            # Skip __typename fields
+            if selection.name.value == "__typename":
+                continue
+            # Recursively process nested selection sets
+            new_field = FieldNode(
+                alias=selection.alias,
+                name=selection.name,
+                arguments=selection.arguments,
+                directives=selection.directives,
+                selection_set=strip_typename_from_selection_set(selection.selection_set),
+            )
+            new_selections.append(new_field)
+        elif isinstance(selection, InlineFragmentNode):
+            # Process inline fragments
+            new_inline = InlineFragmentNode(
+                type_condition=selection.type_condition,
+                directives=selection.directives,
+                selection_set=strip_typename_from_selection_set(selection.selection_set),
+            )
+            new_selections.append(new_inline)
+        elif isinstance(selection, FragmentSpreadNode):
+            # FragmentSpread references a named fragment - keep as-is
+            new_selections.append(selection)
+
+    return SelectionSetNode(selections=tuple(new_selections))
+
+
+def strip_typename_from_operation(operation: OperationDefinitionNode) -> OperationDefinitionNode:
+    """Strip __typename fields from an operation definition.
+
+    Returns a new OperationDefinitionNode with all __typename fields removed
+    from its selection set and any nested selection sets.
+    """
+    return OperationDefinitionNode(
+        operation=operation.operation,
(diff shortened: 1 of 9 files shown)


infrahub_sdk/graphql/utils.py
```diff
@@ -1,5 +1,88 @@
...
+            new_selections.append(new_field)
+        elif isinstance(selection, InlineFragmentNode):
+            # Process inline fragments
+            new_inline = InlineFragmentNode(
+                type_condition=selection.type_condition,
+                directives=selection.directives,
+                selection_set=strip_typename_from_selection_set(selection.selection_set),
+            )
+            new_selections.append(new_inline)
+        elif isinstance(selection, FragmentSpreadNode):
```
ajtmccarty, line 50: should this just be an `else`? if there are items in `selection_set` that are not `FieldNode`, `InlineFragmentNode`, or `FragmentSpreadNode`, then they will be silently deleted

ajtmccarty (approved):
not sure exactly how much testing we have for infrahubctl commands, but it would be good to have a test for this that actually calls `generate_return_types`

infrahub_sdk/graphql/utils.py
BaptisteGi: I checked `SelectionNode` from graphql-core and apparently it could only be `FieldNode`, `InlineFragmentNode` or `FragmentSpreadNode`.

To make it extra safe I added exception if we reach the final `else` statement

ajtmccarty: 👍

PR #727, opened 2025-12-30
Add `value_to_enum_name` filter

gmazoyer:
This Jinja2 filter aims to convert a value into a name when it is part of an enum. This is useful when the name of an enum member has more meaning, e.g. in a string which is exposed to users.

An optional `enum_path` parameter can be specified. When using a valid import string for an enum class, that class will be imported to look for the value within the enum members. This mean that this filter can be used with any enum classes, including the ones found in the Infrahub server code.

Note: I'd like to use this filter to move our current `display_labels` for permission nodes into Jinja2 based `display_label`.

## Summary by CodeRabbit

* **New Features**
  * Added an INFRAHUB filter set to templates, including a filter that converts values into enum member names and exposing these filters in the available filter catalog.
* **Tests**
  * Added comprehensive tests for filter ordering and template integration of the new filter, covering many edge/error cases and various enum input scenarios.

### DIFF
--- a/infrahub_sdk/template/filters.py
+++ b/infrahub_sdk/template/filters.py
@@ -1,4 +1,7 @@
 from dataclasses import dataclass
+from enum import Enum
+from importlib import import_module
+from typing import Any
 
 
 @dataclass
@@ -6,6 +9,42 @@ class FilterDefinition:
     name: str
     trusted: bool
     source: str
+
+
+def value_to_enum_name(value: Any, enum_path: str | None = None) -> str:
+    """Convert a value to its enum member name using the specified enum class.
+
+    This filter takes a raw value and converts it to the corresponding enum member name by dynamically importing the
+    enum class.
+
+    For example, `{{ decision__value | value_to_enum_name("infrahub.permissions.constants.PermissionDecisionFlag") }}`
+    will return: `"ALLOW_ALL"` for value `6`.
+    """
+    if isinstance(value, Enum):
+        return value.name
+
+    if not enum_path:
+        return str(value)
+
+    try:
+        module_path, class_name = enum_path.rsplit(".", 1)
+        module = import_module(module_path)
+        enum_type = getattr(module, class_name)
+    except (ValueError, ImportError, AttributeError):
+        return str(value)
+
+    # Verify that we have a class and that this class is a valid Enum
+    if not (isinstance(enum_type, type) and issubclass(enum_type, Enum)):
+        return str(value)
+
+    try:
+        enum_member = enum_type(value)
+        if enum_member.name is not None:
+            return enum_member.name
+    except (ValueError, TypeError):
+        pass
+
+    return str(value)
 
 
 BUILTIN_FILTERS = [
--- a/infrahub_sdk/template/filters.py
+++ b/infrahub_sdk/template/filters.py
@@ -17,34 +17,36 @@ def value_to_enum_name(value: Any, enum_path: str | None = None) -> str:
     This filter takes a raw value and converts it to the corresponding enum member name by dynamically importing the
     enum class.
 
-    For example, `{{ decision__value | value_to_enum_name("infrahub.permissions.constants.PermissionDecisionFlag") }}`
+    For example, `{{ decision__value | value_to_enum_name("infrahub.core.constants.PermissionDecision") }}`
     will return: `"ALLOW_ALL"` for value `6`.
     """
-    if isinstance(value, Enum):
+    if isinstance(value, Enum) and not enum_path:
         return value.name
 
+    raw_value = value.value if isinstance(value, Enum) else value
+
     if not enum_path:
-        return str(value)
+        return str(raw_value)
 
     try:
         module_path, class_name = enum_path.rsplit(".", 1)
         module = import_module(module_path)
         enum_type = getattr(module, class_name)
     except (ValueError, ImportError, AttributeError):
-        return str(value)
+        return str(raw_value)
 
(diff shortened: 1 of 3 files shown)


infrahub_sdk/template/filters.py
```diff
@@ -8,6 +11,48 @@ class FilterDefinition:
...
+    except (ValueError, TypeError):
+        pass
+
+    return str(value)
+
+
+INFRAHUB_FILTERS = {"value_to_enum_name": value_to_enum_name}
+INFRAHUB_FILTER_DEFINITIONS = [
+    FilterDefinition(name=name, trusted=True, source="infrahub-sdk-python") for name in sorted(INFRAHUB_FILTERS.keys())
+]
```
gmazoyer, line 55: I wrote it like that so that we can easily add filters in the future.

infrahub_sdk/template/filters.py
```diff
@@ -8,6 +11,50 @@ class FilterDefinition:
...
+
+    raw_value = value.value if isinstance(value, Enum) else value
+
+    if not enum_path:
+        return str(raw_value)
+
+    try:
+        module_path, class_name = enum_path.rsplit(".", 1)
+        module = import_module(module_path)
+        enum_type = getattr(module, class_name)
```
ajtmccarty, line 36: is allowing the user to import any Python module inside of the SDK a security concern?
I guess the user can already do this b/c they are using the SDK in their own code, but we also use the SDK within Infrahub and I wonder if there is a way for the user to force the Infrahub server to import something malicious

gmazoyer: We use Jinja2's `SandboxedEnvironment` to render template, so I wonder if this could help prevent having malicious code messing around (I would not expect it to be 100% secure though). That said we also allow users to import their own Jinja2 filters so I guess these could also be entry points to run any kind of code.

ajtmccarty (approved):

infrahub_sdk/template/filters.py
```diff
@@ -8,6 +11,50 @@ class FilterDefinition:
     source: str
 
 
+def value_to_enum_name(value: Any, enum_path: str | None = None) -> str:
+    """Convert a value to its enum member name using the specified enum class.
+
+    This filter takes a raw value and converts it to the corresponding enum member name by dynamically importing the
+    enum class.
+
+    For example, `{{ decision__value | value_to_enum_name("infrahub.core.constants.PermissionDecision") }}`
```
dgarros: Not sure out the format of this filter, it feels a bit complicated to use for the end user

ogenstad: I don't know if it will be useful at all to end users. I think this is pretty much a workaround for https://github.com/opsmill/infrahub/blob/infrahub-v1.6.2/backend/infrahub/core/node/permissions.py#L34 so that we can clean up the "identifier" fields for computed attributes and make them into a computed attribute instead. As such I'm not sure if we should have this in the SDK or if it will just confuse users. Of course after saying that there's little value for end users to have this we could get a feature request for it tomorrow, but given the problem it sets out to solve I think it's fine that it might be a bit complicated to use.

infrahub_sdk/template/filters.py
ogenstad: This shouldn't be executed in any environment where the user can choose their own code to import and not already have that ability. Based on that I think it should be fine to do it this way.

infrahub_sdk/template/filters.py
```diff
@@ -8,6 +11,50 @@ class FilterDefinition:
...
+
+    if not enum_path:
+        return str(raw_value)
+
+    try:
+        module_path, class_name = enum_path.rsplit(".", 1)
+        module = import_module(module_path)
+        enum_type = getattr(module, class_name)
+    except (ValueError, ImportError, AttributeError):
+        return str(raw_value)
```
ogenstad: If we get an import error I think it would be better to have a proper failure. Potentially there would be some future migration issues if we remove a decision from the permissions and the old value no longer exists in the enum, but that should preferably be handled by a migration. So it could be that by catching all of these errors and returning some value we mask a user error if they have made one. Where if we fail early the error should be logged somewhere.

ogenstad (approved):
LGTM, one small comment though.

```python
def value_to_enum_name(value: Any, enum_path: str | None = None) -> str:
    """Convert a value to its enum member name using the specified enum class.
    This filter takes a raw value and converts it to the corresponding enum member name by dynamically importing the
    enum class.
    For example, `{{ decision__value | value_to_enum_name("infrahub.core.constants.PermissionDecision") }}`
```

As this is a docstring within the SDK it could make more sense to link to an enum that lives within the SDK itself. On the other hand it might be worth highlighting that it can come from anywhere. Please ignore this comment if you think it's good as it is.

infrahub_sdk/template/filters.py
dgarros: Let's not include that in the SDK then, let's create a dedicated filter for the permission enum then.

gmazoyer:
> As this is a docstring within the SDK it could make more sense to link to an enum that lives within the SDK itself. On the other hand it might be worth highlighting that it can come from anywhere. Please ignore this comment if you think it's good as it is.

Sadly I havent found a good enum candidate in the SDK code, hence me keeping that enum we use in the Infrahub server code. Though, I added some more explanation in the docstring to explain that the enum can come from anywhere.

ogenstad:
Hey @gmazoyer, I wrote about this one in Slack. We can hold off just a bit with this.

PR #744, opened 2026-01-11
Fix pytest plugin for integration tests and empty YAML files

FragmentedPacket:
- Add missing instantiate_transform() call in integration test runner
- Handle empty YAML files gracefully in loader to avoid TypeError
- Make input field optional in InfrahubInputOutputTest model
- Default input to None for InfrahubIntegrationTest since integration tests get input from live GraphQL queries, not files

## Release Notes

* **Bug Fixes**
  * Resolved initialization order issue in integration tests to ensure proper setup before query execution.
  * Strengthened test data validation to gracefully handle missing or empty test data, preventing unexpected errors.

* **New Features**
  * Extended test framework to support test scenarios without requiring input files, offering greater flexibility.

### DIFF
--- a/infrahub_sdk/pytest_plugin/models.py
+++ b/infrahub_sdk/pytest_plugin/models.py
@@ -31,7 +31,7 @@ class InfrahubInputOutputTest(InfrahubBaseTest):
     directory: Path | None = Field(
         None, description="Path to the directory where the input and output files are located"
     )
-    input: Path = Field(
+    input: Path | None = Field(
         Path("input.json"),
         description="Path to the file with the input data for the test, can be a relative path from the config file or from the directory.",
     )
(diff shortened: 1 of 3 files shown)


infrahub_sdk/pytest_plugin/models.py
```diff
@@ -68,7 +68,7 @@ def update_paths(self, base_dir: Path) -> None:
         else:
             self.directory = base_dir
 
-        if not self.input or not self.input.is_file():
+        if self.input is not None and not self.input.is_file():
```
ajtmccarty, line 71: this is a slight functional change, just want to make sure that is intentional
previously if `self.input` was falsy, then this conditional would run, now it will not

ajtmccarty (approved):

infrahub_sdk/pytest_plugin/models.py
FragmentedPacket: Yes, this is due to integration tests not requiring an `input.json` due to live querying of the data for these.

We originally had to define a dummy `input.json` for integration tests, and this should not require it any longer for integration tests.

PR #732, opened 2025-12-31
Fix code for HFID casting of strings that aren't UUIDs 

FragmentedPacket:
This allows the user to pass in a single for single HFID values as per the documentation

Fixes #731 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HFID/reference normalization: single strings, lists, and UUIDs are now consistently interpreted and stored for single- and multi-reference fields when creating objects.

* **Tests**
  * Added unit tests covering single/many cardinality and mixed reference formats to validate normalization behavior.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #746, opened 2026-01-12
Re-implement casting of HFIDs to a list if the rel peer schema has hf…

FragmentedPacket:
Reimplements #732 but only casting if the peer schema has an HFID defined.

Relevant CI runs in Infrahub: https://github.com/opsmill/infrahub/pull/8096

## Summary by CodeRabbit

* **New Features**
  * Human-friendly ID (HFID) support for relationships: peers now expose HFID info and reference values are normalized during node creation when applicable.

* **Tests**
  * Expanded unit tests covering HFID normalization, object loading with HFIDs, and related-node payload generation.

* **Schema**
  * Test fixtures updated to include human_friendly_id on sample nodes.


infrahub_sdk/spec/object.py
```diff
@@ -119,6 +151,7 @@ async def get_relationship_info(
         info.peer_kind = value["kind"]
 
     peer_schema = await client.schema.get(kind=info.peer_kind, branch=branch)
+    info.peer_has_hfid = bool(peer_schema.human_friendly_id)
```
ajtmccarty: `peer_has_hfid` looks like it should be a `property` on the `RelationshipInfo` class

tests/unit/sdk/spec/test_object.py
```diff
@@ -263,3 +272,256 @@ async def test_parameters_non_dict(client_with_schema_01: InfrahubClient, locati
...
+    - string input → {"id": "..."} in GraphQL
+
+    This test verifies the normalization produces the correct format.
+    """
+    if test_case.peer_has_hfid:
+        # When peer has HFID, use normalization
+        processed_value = normalize_hfid_reference(test_case.input_value)
+    else:
+        # When peer has no HFID, pass value as-is (no normalization)
+        processed_value = test_case.input_value
```
ajtmccarty: this doesn't actually test the logic that checks if the peer schema has an HFID defined. for that, this test probably needs to call `get_relationship_info`. although that looks like it will require some mocking

infrahub_sdk/spec/object.py
FragmentedPacket: Updated

tests/unit/sdk/spec/test_object.py
```diff
@@ -263,3 +273,259 @@ async def test_parameters_non_dict(client_with_schema_01: InfrahubClient, locati
...
+    PeerHfidTestCase(
+        name="peer_has_no_hfid_when_empty_list",
+        peer_human_friendly_id=[],
+        expected_peer_has_hfid=False,
+    ),
+]
+
+
+@pytest.mark.parametrize("test_case", PEER_HFID_TEST_CASES, ids=lambda tc: tc.name)
+async def test_peer_has_hfid_from_schema(test_case: PeerHfidTestCase) -> None:
```
ajtmccarty: this test has a little too much mocking, so it ends up barely testing anything.
I think to actually test this you would need a test on `create_node`, but I don't think we have any of those in the SDK
so perhaps the easiest option is to just add a positive and negative `functional` or `integration` test to infrahub community that uses this version of the SDK

infrahub_sdk/spec/object.py
ajtmccarty: sorry I thought `peer_schema` was part of `RelationshipInfo`, so you could add a `peer_has_hfid` property to `RelationshipInfo`. this seems like it works fine though so you can just leave it

ajtmccarty (approved):

Issue #697, opened 2025-12-17
bug: Problems with typing related to generated protocols


ogenstad:
### Infrahub SDK version

v1.17.0

### Current Behavior

If you create a file that looks like this:

run.py
```python
from infrahub_sdk import InfrahubClient
from infrahub_sdk.protocols import BuiltinTag, CoreGenericRepository


async def show_them() -> None:
    client = InfrahubClient()
    tags = await client.all(kind=BuiltinTag)
    for tag in tags:
        print(tag.name)

    red_tag = await client.get(kind=BuiltinTag, name__value="red")
    print(f"Found tag: {red_tag.name}")
    repo = await client.get(
        kind=CoreGenericRepository,
        name__value="edge",
        branch="main",
    )
    print(f"Found repository: {repo.name} at location {repo.location}")
```

Then try to run it through a type checker we get an error

```bash
❯ mypy run.py
run.py:7: error: Only concrete class can be given where "type[BuiltinTag]" is expected  [type-abstract]
        tags = await client.all(kind=BuiltinTag)
                                     ^~~~~~~~~~
run.py:11: error: Only concrete class can be given where "type[BuiltinTag]" is expected  [type-abstract]
        red_tag = await client.get(kind=BuiltinTag, name__value="red")
                                        ^~~~~~~~~~
run.py:14: error: Only concrete class can be given where "type[CoreGenericRepository]" is expected  [type-abstract]
            kind=CoreGenericRepository,
                 ^~~~~~~~~~~~~~~~~~~~~
Found 3 errors in 1 file (checked 1 source file)
```

Probably no one has noticed this before due to the lack of a py.type file in the repo (#543).

In the few places where we actually use this within the SDK the problem is ignored with an inline exception to the typing: https://github.com/opsmill/infrahub-sdk-python/blob/v1.17.0/infrahub_sdk/testing/repository.py#L101

### Expected Behavior

This is expected to work without any reported issues.

### Steps to Reproduce

Follow the above steps

ogenstad:
Resolved in #763.

PR #749, opened 2026-01-19
IHS-183: Fix typing errors for protocols

solababs:
Fixes https://github.com/opsmill/infrahub-sdk-python/issues/697

Replace protocols with concrete classes to address `type-abstract` mypy errors

## Summary by CodeRabbit

* **Chores**
  * Updated Ruff linting tool from v0.11.9 to v0.14.10

* **Refactor**
  * Strengthened and expanded type annotations across the SDK for clearer type checking and IDE support
  * Removed type-suppression directives to surface compile-time type issues
  * Converted several protocol-like bases to concrete bases with default behaviors and adjusted schema validation messaging

### DIFF
--- a/infrahub_sdk/node/related_node.py
+++ b/infrahub_sdk/node/related_node.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import re
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, cast
 
 from ..exceptions import Error
 from ..protocols_base import CoreNodeBase
@@ -11,7 +11,7 @@ from .metadata import NodeMetadata, RelationshipMetadata
 if TYPE_CHECKING:
     from ..client import InfrahubClient, InfrahubClientSync
     from ..schema import RelationshipSchemaAPI
-    from .node import InfrahubNode, InfrahubNodeSync
+    from .node import InfrahubNode, InfrahubNodeBase, InfrahubNodeSync
 
 
 class RelatedNodeBase:
@@ -34,7 +34,7 @@ class RelatedNodeBase:
         self._properties_object = PROPERTIES_OBJECT
         self._properties = self._properties_flag + self._properties_object
 
-        self._peer: InfrahubNode | InfrahubNodeSync | CoreNodeBase | None = None
+        self._peer: InfrahubNode | InfrahubNodeSync | InfrahubNodeBase | CoreNodeBase | None = None
         self._id: str | None = None
         self._hfid: list[str] | None = None
         self._display_label: str | None = None
@@ -43,8 +43,10 @@ class RelatedNodeBase:
         self._source_typename: str | None = None
         self._relationship_metadata: RelationshipMetadata | None = None
 
-        if isinstance(data, (CoreNodeBase)):
-            self._peer = data
+        # Check for InfrahubNodeBase instances using duck-typing (_schema attribute)
+        # to avoid circular imports, or CoreNodeBase instances
+        if isinstance(data, CoreNodeBase) or hasattr(data, "_schema"):
+            self._peer = cast("InfrahubNodeBase | CoreNodeBase", data)
             for prop in self._properties:
                 setattr(self, prop, None)
             self._relationship_metadata = None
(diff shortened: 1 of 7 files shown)


infrahub_sdk/node/related_node.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, branch: str, schema: RelationshipSchemaAPI, data: Any | dict,
         self._properties_object = PROPERTIES_OBJECT
         self._properties = self._properties_flag + self._properties_object
 
-        self._peer = None
+        self._peer: InfrahubNode | InfrahubNodeSync | InfrahubNodeBase | CoreNodeBase | None = None
```
ajtmccarty: could this be `InfrahubNodeBase | CoreNodeBase | None`? looks like the other two `InfrahubNode` classes inherit from `InfrahubNodeBase`

infrahub_sdk/schema/__init__.py
```diff
@@ -207,14 +208,14 @@ def _get_schema_name(schema: type[SchemaType | SchemaTypeSync] | str) -> str:
         if isinstance(schema, str):
             return schema
 
-        if hasattr(schema, "_is_runtime_protocol") and getattr(schema, "_is_runtime_protocol", None):
+        if isinstance(schema, type) and issubclass(schema, (CoreNode, CoreNodeSync)):
```
ajtmccarty: could this be ` issubclass(schema, (CoreNodeBase))`?

infrahub_sdk/protocols_base.py
```diff
@@ -171,8 +171,7 @@ class AnyAttributeOptional(Attribute):
     value: float | None
 
 
-@runtime_checkable
-class CoreNodeBase(Protocol):
+class CoreNodeBase:
```
ajtmccarty, line 174: would be worth updating the PR description to explain why these are no longer `Protocol`s

infrahub_sdk/node/related_node.py
```diff
@@ -43,8 +43,10 @@ def __init__(self, branch: str, schema: RelationshipSchemaAPI, data: Any | dict,
         self._source_typename: str | None = None
         self._relationship_metadata: RelationshipMetadata | None = None
 
-        if isinstance(data, (CoreNodeBase)):
-            self._peer = data
+        # Check for InfrahubNodeBase instances using duck-typing (_schema attribute)
+        # to avoid circular imports, or CoreNodeBase instances
+        if isinstance(data, CoreNodeBase) or hasattr(data, "_schema"):
+            self._peer = cast("InfrahubNodeBase | CoreNodeBase", data)
```
ajtmccarty, line 49: maybe it is the best solution right now, but it smells
is the circular import error resolvable? I suspect it is probably pretty tangled, but maybe not

infrahub_sdk/protocols_base.py
```diff
@@ -189,23 +188,28 @@ def get_human_friendly_id(self) -> list[str] | None: ...
 
     def get_human_friendly_id_as_string(self, include_kind: bool = False) -> str | None: ...
 
-    def get_kind(self) -> str: ...
+    def get_kind(self) -> str:
+        return ""
```
ogenstad: These methods should not be called so as such I think we can just have a `raise NotImplementedError` instead of having these placeholder returns.

infrahub_sdk/protocols_base.py
solababs: Replace protocols with concrete classes to address `type-abstract` mypy errors

infrahub_sdk/node/related_node.py
solababs: No it is not

ogenstad:
One thing that nags me about this is that we still have all of the nodes within a file called `protocols.py` which could potentially be a bit misleading since they are not protocols any more so it might be a bit confusing. However the only reason we have this is for typing and I don't expect any external protocols to have been used in this context and even if they did we could see that it didn't work as intended for typing.

I don't think that it's worth renaming or moving around the files to correct this confusion. Just wanted to see if you guys had any thoughts around it? (@solababs, @ajtmccarty)

ajtmccarty:
maybe they could move to `models_base.py` and `models.py`, but I'm not sure if that would break backwards compatibility

ajtmccarty (approved):

ogenstad:
> maybe they could move to `models_base.py` and `models.py`, but I'm not sure if that would break backwards compatibility

I agree. We'll probably do more work in this area with proper pydantic models. Let's keep the slightly incorrect filenames for now.

ogenstad (approved):

PR #776, opened 2026-01-27
Prevent cascade delete from causing SDK tracking delete to fail

### LINKED ISSUE #265: bug: SDK Tracking feature errors out when handling parent/component deletion sequence
### Component

_No response_

### Infrahub SDK version

1.1.6

### Current Behavior

If the parent and all the components of the relationship are being deleted due to the difference in the generator, it fails with: `['NetworkPhysicalInterfaceL3Delete'] Unable to find the node 1821cc90-7fb1-f8b3-3b21-c51f9599f512 / NetworkPhysicalInterfaceL3 in the database.`

This example is all interfaces (component) to a device (parent) are being deleted as well as the parent object.

### Expected Behavior

It would just delete the parent node rather than all the component objects tied to the node to prevent it attempting to delete a component when the parent has been deleted and the deletion is cascading down.

### Steps to Reproduce

1. Spin up infrahub
2. Load `base` and `extensions/location_minimal` from https://github.com/opsmill/schema-library
3. Create a generator that will build a site, 2 devices, and 24 interfaces for each device (ranges would work to keep generator simple)
4. Run the generator and let it create the devices and interfaces
5. Modify the generator to only generate one of those devices
6. Execute generator and it should show an error that it fails to delete one of the interfaces due to it no longer existing

### Additional Information

While researching this further, I believe this code may need to account for parent/component relationships when processing previous children.

https://github.com/opsmill/infrahub-sdk-python/blob/stable/infrahub_sdk/query_groups.py#L102-L111

ogenstad:
Currently there's an issue with deleting nodes using a generator and the SDK tracking feature if the impacted nodes have a schema that also uses the cascade delete feature.

For example if we have a number of devices and interfaces that should be deleted and we happen to delete one of the devices first we'd cause the backend cascade delete feature to remove all of the interfaces attached to that device. When the SDK then tries to remove such an interface we previously raised an error. The change here is that we instead catch those errors and ignore GraphQL errors if they are raised because a node was not found.

A larger refactoring could be warranted here. Both in terms of error management and enrichment from the backend but also the ability to send in a list of all of the nodes we want to delete in one transaction to the backend instead of calling a delete mutation for each entry.

Fixes #265 

## Summary by CodeRabbit

* **Bug Fixes**
  * SDK tracking no longer stops when cascade-deleted nodes trigger delete errors; cleanup now continues safely when affected nodes were already removed.

* **Chores**
  * Added a changelog entry documenting the cascade-delete handling fix.

### DIFF
--- a/infrahub_sdk/query_groups.py
+++ b/infrahub_sdk/query_groups.py
@@ -4,7 +4,7 @@ from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any
 
 from .constants import InfrahubClientMode
-from .exceptions import NodeNotFoundError
+from .exceptions import GraphQLError, NodeNotFoundError
 from .utils import dict_hash
 
 if TYPE_CHECKING:
@@ -109,7 +109,13 @@ class InfrahubGroupContext(InfrahubGroupContextBase):
         if self.previous_members and self.unused_member_ids:
             for member in self.previous_members:
                 if member.id in self.unused_member_ids and member.typename:
-                    await self.client.delete(kind=member.typename, id=member.id)
+                    try:
+                        await self.client.delete(kind=member.typename, id=member.id)
+                    except GraphQLError as exc:
+                        if not exc.message or "Unable to find the node" not in exc.message:
+                            # If the node already has been deleted, skip the error as it would have been deleted
+                            # by the cascade delete of another node
+                            raise
 
     async def add_related_nodes(self, ids: list[str], update_group_context: bool | None = None) -> None:
         """
--- a/changelog/265.fixed.md
+++ b/changelog/265.fixed.md
@@ -1,0 +1,1 @@
+Allow SDK tracking feature to continue after encountering delete errors due to impacted nodes having already been deleted by cascade delete.


infrahub_sdk/query_groups.py
```diff
@@ -109,7 +109,17 @@ async def delete_unused(self) -> None:
...
+                    try:
+                        await self.client.delete(kind=member.typename, id=member.id)
+                    except GraphQLError as exc:
+                        if not exc.message or "Unable to find the node" not in exc.message:
+                            # If the node already has been deleted, skip the error as it would have been deleted
+                            # by the cascade delete of another node
+                            # Note this is a quick fix and could be improved by the delete order based on relationship
+                            # types along with cascade delete settings. Alternatively send a batch delete request.
+                            # Also see INFP-468 for better error handling from the backend side to avoid having to
+                            # parse the error message as above.
```
ajtmccarty: I'd say this comment should be `Node has already been deleted, ignore error`. everything else should be captured by Jira/GH issues

ajtmccarty (approved):

infrahub_sdk/query_groups.py
ogenstad: Fair enough. I've shortened the comment.

PR #792, opened 2026-02-02
IHS-193 Add support for file upload/download for `CoreFileObject`

gmazoyer:
This change adds support to create, update and retrieve nodes which schemas implement `CoreFileObject`.

The proposed user exposed API follows these key points:
- `node.upload_from_path(path)` to select a file from disk for upload (streams during upload)
- `node.upload_from_bytes(content, name)` to set content for upload (supports bytes or BinaryIO)
- `node.download_file(dest)`  to download a file to memory or stream to disk

Being able to stream a file to disk or from a disk is important in order to support large files and to avoid them being loaded completely into memory (which would be problematic for +1GB files in general).

The choice of using `upload_from_path` and `upload_from_bytes` is to prevent a collision with a potential attribute or relationship called `file` in the schema. That is also the reason why the `file` GraphQL parameter is outside the `data` one instead of inside it.

Here we introduce a couple of components to try to make our code SOLID (it's not much for now, but it's honest work):
- `FileHandler` / `FileHandlerSync` dedicated classes for file I/O operations
- `MultipartBuilder` GraphQL Multipart Request Spec payload building

It sure won't make our code SOLID but it won't add to the burden for now.

So given the user who loaded a schema, using our SDK will look like:

####  Upload a file when creating a node

```python
from pathlib import Path
from infrahub_sdk import InfrahubClientSync

client = InfrahubClientSync()

contract = client.create(
    kind="NetworkCircuitContract", contract_start="2026-01-01", contract_end="2026-12-31"
)

# Option 1: Select file from disk (will be streamed during upload)
contract.upload_from_path(path=Path("/tmp/contract.pdf"))

# Option 2: Upload from memory (for small files or dynamically generated content)
contract.upload_from_bytes(content=b"file content", name="contract.pdf")

# Save as usual
contract.save()
```

#### Download a file from a node

```python
from pathlib import Path

contract = client.get(kind="NetworkCircuitContract", id="abc123")

# Download to memory (suitable for small files)
content = contract.download_file()

# Or stream directly to disk (suitable for large files)
bytes_written = contract.download_file(dest=Path("/tmp/downloaded.pdf"))
```

## Summary by CodeRabbit

* **New Features**
  * GraphQL multipart file upload support and node-level file attach/clear APIs.
  * File download for file-like nodes (in-memory or streamed to disk).
  * Streaming GET responses and improved proxy/config handling for network requests.

* **Tests**
  * Comprehensive unit tests for multipart uploads, file preparation, downloads/streaming, and node file workflows.

### DIFF
--- a/infrahub_sdk/client.py
+++ b/infrahub_sdk/client.py
@@ -1057,6 +1057,45 @@ class InfrahubClient(BaseClient):
             timeout=timeout or self.default_timeout,
         )
 
+    @asynccontextmanager
+    async def _get_streaming(
+        self, url: str, headers: dict | None = None, timeout: int | None = None
+    ) -> AsyncIterator[httpx.Response]:
+        """Execute a streaming HTTP GET with HTTPX.
+
+        Returns an async context manager that yields the streaming response.
+        Use this for downloading large files without loading into memory.
+
+        Raises:
+            ServerNotReachableError if we are not able to connect to the server
+            ServerNotResponsiveError if the server didn't respond before the timeout expired
+        """
+        await self.login()
+
+        headers = headers or {}
+        base_headers = copy.copy(self.headers or {})
+        headers.update(base_headers)
+
+        proxy_config: ProxyConfig = {"proxy": None, "mounts": None}
+        if self.config.proxy:
+            proxy_config["proxy"] = self.config.proxy
+        elif self.config.proxy_mounts.is_set:
+            proxy_config["mounts"] = {
+                key: httpx.AsyncHTTPTransport(proxy=value)
+                for key, value in self.config.proxy_mounts.model_dump(by_alias=True).items()
+            }
+
+        async with httpx.AsyncClient(**proxy_config, verify=self.config.tls_context) as client:
+            try:
+                async with client.stream(
+                    method="GET", url=url, headers=headers, timeout=timeout or self.default_timeout
+                ) as response:
+                    yield response
+            except httpx.NetworkError as exc:
+                raise ServerNotReachableError(address=self.address) from exc
+            except httpx.ReadTimeout as exc:
+                raise ServerNotResponsiveError(url=url, timeout=timeout or self.default_timeout) from exc
+
     async def _request(
         self,
         url: str,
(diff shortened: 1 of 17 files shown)


infrahub_sdk/file_handler.py
```diff
@@ -0,0 +1,348 @@
...
+        if exc.response.status_code in {401, 403}:
+            response = exc.response.json()
+            errors = response.get("errors", [])
+            messages = [error.get("message") for error in errors]
+            raise AuthenticationError(" | ".join(messages)) from exc
+        if exc.response.status_code == 404:
+            response = exc.response.json()
+            detail = response.get("detail", "File not found")
+            raise NodeNotFoundError(node_type="FileObject", identifier=detail) from exc
+        raise exc
```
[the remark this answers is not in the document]
gmazoyer: This will be responses from our GraphQL endpoints, so 401, 403 and 404 will be JSON formatted anyway.

wvandeun:
I would prefer the `upload_from_path` and `upload_from_content` names. Maybe even `upload_from_bytes` for the latter?

infrahub_sdk/client.py
```diff
@@ -1008,6 +994,130 @@ async def execute_graphql(
 
         # TODO add a special method to execute mutation that will check if the method returned OK
 
+    async def execute_graphql_with_file(
```
ogenstad: Do we want this to be a public method? I.e. is it something that users would ever call directly?

infrahub_sdk/client.py
```diff
@@ -1008,6 +994,130 @@ async def execute_graphql(
...
+        self,
+        query: str,
+        variables: dict | None = None,
+        file_content: BinaryIO | None = None,
+        file_name: str | None = None,
+        branch_name: str | None = None,
+        at: str | Timestamp | None = None,
```
ogenstad: For this method we might not want/need the `at` param as we won't support uploading a file in the past.

infrahub_sdk/node/node.py
```diff
@@ -68,11 +71,16 @@ def __init__(self, schema: MainSchemaTypesAPI, branch: str, data: dict | None =
         # GenericSchemaAPI doesn't have inherit_from, so we need to check the type first
         if isinstance(schema, GenericSchemaAPI):
             self._artifact_support = False
+            self._file_object_support = False
         else:
             inherit_from = getattr(schema, "inherit_from", None) or []
             self._artifact_support = "CoreArtifactTarget" in inherit_from
+            self._file_object_support = "CoreFileObject" in inherit_from
```
ogenstad: Not really related to this PR but the if/else statement looks a bit redundant, just having:

```python
inherit_from = getattr(schema, "inherit_from", None) or []
self._artifact_support = "CoreArtifactTarget" in inherit_from
self._file_object_support = "CoreFileObject" in inherit_from
```
Should do the same thing, but I don't know if it helps with the readability in any way..

ogenstad (approved):
It would be good with some integration tests once we have things updated on the Infrahub side.

PR #802, opened 2026-02-06
Add integration tests for file objects

gmazoyer:
These tests won't pass yet, as we need the upcoming Infrahub 1.8 to run them, hence marking them as XFAIL for now.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive integration tests and fixtures for FileObject lifecycle: uploads (bytes & path), updates and upserts, downloads (memory & disk), metadata checks (name, size, checksum, storage id), schema-loading fixtures, node-creation helpers, and relationship-based queries — available for both asynchronous and synchronous test flows.

### DIFF
--- a/infrahub_sdk/testing/schemas/file_object.py
+++ b/infrahub_sdk/testing/schemas/file_object.py
@@ -1,7 +1,6 @@
 import pytest
 
 from infrahub_sdk import InfrahubClient, InfrahubClientSync
-from infrahub_sdk.exceptions import GraphQLError
 from infrahub_sdk.node import InfrahubNode, InfrahubNodeSync
 from infrahub_sdk.schema.main import AttributeKind, NodeSchema, RelationshipKind, SchemaRoot
 from infrahub_sdk.schema.main import AttributeSchema as Attr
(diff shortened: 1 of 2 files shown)


infrahub_sdk/testing/schemas/file_object.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @pytest.fixture(scope="class")
+    def circuit_main_sync(
+        self,
+        client_sync: InfrahubClientSync,
+        load_file_object_schema_sync: None,  # noqa: ARG002
+    ) -> InfrahubNodeSync:
+        obj = client_sync.create(kind=TESTING_CIRCUIT, circuit_id="CIRCUIT-SYNC-001", bandwidth=2000)
+        obj.save()
+        return obj
```
ogenstad: I wonder if it's valuable to have this included in the SDK or if it should just be a fixture within the integration tests directory. Not a bit issue for me but do you think we'll use it outside of the SDK tests?

ogenstad (approved):

infrahub_sdk/testing/schemas/file_object.py
gmazoyer: I think it kinda highlight that the line I drew between SDK integration tests and backend integration tests is a bit blurry to me. I think I will remove these completely from the SDK as they don't seem very well suited here and we have something similar in the backend integration tests anyway.

PR #805, opened 2026-02-06
Fix infrahubctl proposed change table generation

gmazoyer:
This was broken because the API server will now expose the `created_by` and `created_at` values inside the node metadata.

Also add `get_node_metadata` to `CoreNodeBase` protocol.

## Summary by CodeRabbit

* **New Features**
  * Added a public API to retrieve node-specific metadata.

* **Improvements**
  * Branch reporting now computes and displays creator and timestamp info from node metadata.
  * Proposed-change retrieval updated to use metadata-based inclusion for richer results.

* **Tests**
  * Updated unit tests to reflect the revised node metadata structure.

* **Documentation**
  * Added changelog entry describing the metadata-based branch report change.

### DIFF
--- a/changelog/+infrahubctl-proposedchange.changed.md
+++ b/changelog/+infrahubctl-proposedchange.changed.md
@@ -1,0 +1,1 @@
+Updated branch report command to use node metadata for proposed change creator information instead of the deprecated relationship-based approach. Requires Infrahub 1.7 or above.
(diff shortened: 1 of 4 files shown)


ogenstad (approved):
LGTM.

This will of course only work for installations after Infrahub 1.7. It might be that this being targeted for `infrahub-develop` would mean that it only gets included in an SDK version that will require Infrahub 1.8. Not related to this PR, just feels like it would be nice to have this resolved. For this version we should probably clearly state in the release notes of the SDK to highlight what release is required. Then after that we can hopefully avoid breaking changes between version.

PR #807, opened 2026-02-09
feat: IHS-190 (part of IFC-2162) Add namespace restriction parameter to generic schemas

polmichel:
To understand the global feature, see this PR: https://github.com/opsmill/infrahub/pull/8319

This PR is about updating `infrahub_sdk.schema.main.GenericSchemaAPI` with the new generic schema parameter "restricted_namespaces".

### Goal
See https://github.com/opsmill/infrahub/pull/8319

### Non-goals
Added another missing attribute to `infrahub_sdk.schema.main.GenericSchemaAPI`: hierarchical. This attribute was already into the internal Infrahub generic schema model.

### What changed
Generic schemas into SDK API.

### How-to review
This PR is quite small: commit by commit, in the chronological order.

### How to test

**Automated back-end tests:**

`uv run pytest tests/unit/ctl/test_schema_app.py tests/unit/sdk/test_schema.py
`
- new test loading a generic schema through infrahubctl command, Infrahub API is mocked
- new test ensuring that a schema not following the namespace restriction rule is rejected with the correct error message when loaded from python-sdk utilities. Infrahub API is mocked.

**Application testing scenarios:**

- Load a valid generic schema from SDK API using the new attribute "restricted_namespaces".
- Load an invalid schema from SDK API not following the namespace restriction rule.

### Impact & rollout
Deployment notes: safe to deploy

Closes IHS-190

## Summary by CodeRabbit

* **New Features**
  * Added optional "hierarchical" and "restricted namespaces" configuration options to the schema API.

* **Tests**
  * Added tests for loading generic schemas that include namespace restrictions.
  * Added tests to surface API error handling when schema load requests fail.

### DIFF
--- a/tests/unit/sdk/test_schema.py
+++ b/tests/unit/sdk/test_schema.py
@@ -1,4 +1,5 @@
 import inspect
+import re
 from io import StringIO
 from unittest import mock
 from unittest.mock import MagicMock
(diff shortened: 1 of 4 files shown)


ogenstad:
Small comment on this PR. It should target the `infrahub-develop` branch. The reason being is that anything within that branch covers features included in an upcoming Infrahub version. Where as the `develop` branch of the SDK might be for things that get included in an SDK release that's outside of the Infrahub release schedule.

tests/unit/sdk/test_schema.py
```diff
@@ -482,3 +483,56 @@ def test_schema_base__get_schema_name__returns_correct_schema_name_for_protocols
     assert InfrahubSchemaBase._get_schema_name(schema=BuiltinIPAddressSync) == "BuiltinIPAddress"
     assert InfrahubSchemaBase._get_schema_name(schema=BuiltinIPAddress) == "BuiltinIPAddress"
     assert InfrahubSchemaBase._get_schema_name(schema="BuiltinIPAddress") == "BuiltinIPAddress"
+
+
+async def test_schema_load_rejected_when_node_namespace_violates_generic_restricted_namespaces(
```
ogenstad: I'm not sure that this test actually tests anything meaningful, since we're just mocking the response. I.e. are we doing something here that isn't already covered by other tests?

ogenstad (approved):
LGTM, but as indicated by the comment I'm not sure about that second test.

tests/unit/sdk/test_schema.py
polmichel: This tests that the error message is going from the API response to the user through the SDK layer. This does not bring much value, I could understand if you would rather delete it.

ogenstad: I think in this case that the test name is a bit confusing as it really doesn't have anything to do with generics and restricted names but only how the SDK responds to errors in general. It's just that the dummy data happens to be one of this new kind, does that make sense?

polmichel: Yes, I generalized and simplified the unit test.

ogenstad (approved):

PR #810, opened 2026-02-09
Fix: exclude uninitialized hierarchy relationships from mutation data

minitriga:
When saving a hierarchical node with allow_upsert=True, the SDK was including `parent: null` in the GraphQL mutation for nodes without a parent (e.g. top-level fabric nodes). Infrahub rejects this because hierarchy relationships are not assignable via mutations.

Skip hierarchy relationships in _generate_input_data() unless they were explicitly initialized with data.

## Summary by CodeRabbit

* **Bug Fixes**
  * Mutation payloads now omit uninitialized hierarchical relationships (parent/children/ancestors/descendants) unless explicitly set, preventing unintended data from being sent.

* **Tests**
  * Added tests verifying hierarchical relationship input data is excluded when uninitialized and included when explicitly initialized, across both standard and sync client modes.

### DIFF
--- a/infrahub_sdk/node/node.py
+++ b/infrahub_sdk/node/node.py
@@ -258,6 +258,10 @@ class InfrahubNodeBase:
 
             rel: RelatedNodeBase | RelationshipManagerBase = getattr(self, item_name)
 
+            # Skip hierarchy relationships (parent/children) unless explicitly set
+            if rel_schema.hierarchical and not rel.initialized:
+                continue
+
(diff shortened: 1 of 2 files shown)


ajtmccarty (approved):
I am not the most familiar with all the inner workings of the SDK, but I think this looks reasonable
I'd give it a day for anyone else on @opsmill/backend  to review before merging

minitriga:
Closing this and creating a better issue to track this as there is a larger problem here: #821

PR #835, opened 2026-02-18
Add the ability to export the schema in yaml

### LINKED ISSUE #151: feature: Ability to export the schema in YAML format using infrahubctl
### Component

infrahubctl CLI

It would be good if infrahubctl could export the schema of a given Infrahub server in YAML format. The command should support filtering options for namespace, name etc so that it's possible to remove restricted namespaces from the export (as it wouldn't be possible to import these again)

Some filtering and cleanup would also be required to avoid having the schema be too verbose and include IDs etc.

### Describe the Use Case

Provide an easy way to export and reimport a schema

### Additional Information

_No response_

BeArchiTek:
Fixes #151 

## Summary by CodeRabbit

* **New Features**
  * CLI command to export Infrahub schemas as per-namespace YAML files with options for output directory, branch, and namespace filtering.

* **Library**
  * SDK exposes schema export capability producing per-namespace exports (nodes and generics), honoring restricted namespaces and empty-export cases.

* **Documentation**
  * Added usage docs and option reference for the schema export command.

* **Tests**
  * Added tests covering export behavior, filtering, restricted/empty cases, and client integration.

* **Chores**
  * New changelog entry documenting the export command.


infrahub_sdk/constants.py
```diff
@@ -1,5 +1,23 @@
...
+    "Core",
+    "Deprecated",
+    "Diff",
+    "Infrahub",
+    "Internal",
+    "Lineage",
+    "Schema",
+    "Profile",
+    "Template",
+]
```
ogenstad: Instead of doing it like this I'd suggest that we have this part be included within the schema component of the SDK. The backend will return this information (to indicate if the namespace is editable by the user or not). It could be that the SDK doesn't yet deal with that part of the API. But we should not introduce a new list for this and support them separately.

infrahub_sdk/ctl/schema.py
```diff
@@ -211,3 +219,62 @@ def _display_schema_warnings(console: Console, warnings: list[SchemaWarning]) ->
...
+            continue
+        if namespace and schema.namespace not in namespace:
+            continue
+        ns = schema.namespace
+        user_schemas.setdefault(ns, {"nodes": [], "generics": []})
+        schema_dict = schema_to_export_dict(schema)
+        if isinstance(schema, GenericSchemaAPI):
+            user_schemas[ns]["generics"].append(schema_dict)
+        else:
+            user_schemas[ns]["nodes"].append(schema_dict)
```
ogenstad: I think this logic should live under the `InfrahubClient.schema` functionallity. As it is today I think we have a bit too much logic within `infrahubctl` and it would be better to have the ctl be focused on presentation and have the core logic happen elsewhere.

BeArchiTek: ok will do

infrahub_sdk/constants.py
BeArchiTek: Can we do it as part of another PR later on ?

tests/unit/sdk/test_schema_export.py
```diff
@@ -0,0 +1,244 @@
...
+
+from typing import TYPE_CHECKING, Any
+
+import pytest
+
+from infrahub_sdk import Config, InfrahubClient, InfrahubClientSync
+
+if TYPE_CHECKING:
+    from pytest_httpx import HTTPXMock
+from infrahub_sdk.schema import (
```
ogenstad, line 8: This import should be above the "if TYPE_CHECKING" section. I would have expected the linters to pick up on that..

tests/unit/sdk/test_schema_export.py
```diff
@@ -0,0 +1,244 @@
...
+
+@pytest.mark.parametrize("client_type", ["async", "sync"])
+async def test_export_returns_user_schemas(httpx_mock: HTTPXMock, client_type: str) -> None:
+    response = _schema_response(
+        nodes=[_make_node_dict("Infra", "Device"), _make_node_dict("Dcim", "Rack")],
+        generics=[_make_generic_dict("Infra", "GenericInterface")],
+    )
+    httpx_mock.add_response(method="GET", url="http://mock/api/schema?branch=main", json=response)
+
+    if client_type == "async":
```
ogenstad: This doesn't follow the pattern of existing tests when using the default async client. I'd reuse the same approach or perhaps use the `clients` fixture that return both of them.

infrahub_sdk/schema/__init__.py
```diff
@@ -739,6 +793,28 @@ def fetch(
 
         return branch_schema.nodes
 
+    def export(
+        self,
+        branch: str | None = None,
+        namespaces: list[str] | None = None,
+    ) -> dict[str, dict[str, list[dict[str, Any]]]]:
```
ogenstad: The return value of `dict[str, dict[str, list[dict[str, Any]]]]:` makes it very hard for users to know what to expect from this method. I think this can be fine if we're talking about an initial version but moving forward I think we want an actual object being returned that matches what a schema looks like. Then on that object there could be a `.to_dict()` method which would deal with converting the schema into dictionary, which would do a lot of the work that `schema_to_export_dict` does today.

ogenstad (commented):
I've added some comments to this. If required for a project we can go forward with this. However the format change I mention to the .export() method would be considered a breaking change when we do that.

In order to allow for easier refactoring it would be good with an integration test where we loaded a schema to Infrahub and then exported it again to then compare the schema that we just loaded with what we got out from infrahub using `infrahubctl`.

infrahub_sdk/constants.py
BeArchiTek: As seen together - We'll follow up in a separate PR to consume this from the backend API once the SDK supports that part of the schema endpoint.

infrahub_sdk/ctl/schema.py
BeArchiTek: Done — the export logic now lives under `InfrahubClient.schema.export()` (both async and sync). The ctl command only handles CLI presentation (directory creation, YAML file writing, console output).

tests/unit/sdk/test_schema_export.py
BeArchiTek: Fixed — moved the `infrahub_sdk.schema` import above the `TYPE_CHECKING` block. Also added ruff to catch this going forward.

tests/unit/sdk/test_schema_export.py
BeArchiTek: Fixed — tests now use the `clients` fixture with `client_types` parametrize pattern matching the existing conventions in `test_schema.py`.

infrahub_sdk/schema/__init__.py
BeArchiTek: Agreed — the nested dict return type is not great for discoverability. Keeping it as-is for the initial version, but will plan a follow-up to introduce a proper schema export result object with a `.to_dict()` method.

infrahub_sdk/ctl/schema.py
```diff
@@ -211,3 +212,49 @@ def _display_schema_warnings(console: Console, warnings: list[SchemaWarning]) ->
...
+    timestamp = datetime.now(timezone.utc).astimezone().strftime("%Y%m%d-%H%M%S")
+    return Path(f"infrahub-schema-export-{timestamp}")
+
+
+@app.command()
+@catch_exception(console=console)
+async def export(
+    directory: Path = typer.Option(_default_export_directory, help="Directory path to store schema files"),
+    branch: str = typer.Option(None, help="Branch from which to export the schema"),
+    namespace: list[str] = typer.Option([], help="Namespace(s) to export (default: all user-defined)"),
```
ogenstad: Since this is a list "namespaces" might make more sense than "namespace"?

ogenstad (approved):

PR #836, opened 2026-02-18
Add specs for artifact composition

ogenstad:
# Why

Provide specifications for INFP-504

## What changed

Add Spec Kit specifications

## Summary by CodeRabbit

* **Documentation**
  * Added a new specification and a requirements checklist for composing rendered artifact content in templates via Jinja2-style filters, covering content-inlining and file-embedding scenarios, JSON/YAML parsing, a trust/security model that restricts certain filters during validation vs runtime, error-handling expectations, acceptance criteria, success metrics, testing guidance, assumptions, and open planning questions.

### DIFF
--- a/dev/specs/infp-504-artifact-composition/spec.md
+++ b/dev/specs/infp-504-artifact-composition/spec.md
@@ -99,13 +99,14 @@ The existing single `restricted: bool` parameter on `validate()` is insufficient
 - **FR-004**: The system MUST provide a `file_object_content` Jinja2 filter that accepts a `storage_id` string and returns the raw string content of the referenced file object, using the file-object-specific API path or metadata handling — this implementation is distinct from `artifact_content`.
 - **FR-005**: Both `artifact_content` and `file_object_content` MUST raise `JinjaFilterError` when the input `storage_id` is null or empty, or when the object store cannot retrieve the content for any reason (not found, network failure, auth failure).
 - **FR-006**: Both `artifact_content` and `file_object_content` MUST raise `JinjaFilterError` when invoked and no `InfrahubClient` was supplied to `Jinja2Template` at construction time. The error message MUST name the filter and explain that an `InfrahubClient` is required.
-- **FR-007**: Both `artifact_content` and `file_object_content` MUST be registered with `trusted=False` in the `FilterDefinition` registry so that `validate(restricted=True)` blocks them in the computed attributes execution context (Infrahub API server). They are only permitted to execute on Prefect workers, where an `InfrahubClient` is available.
+- **FR-007**: Both `artifact_content` and `file_object_content` MUST be registered with `trusted=False` in the `FilterDefinition` registry so that `validate(restricted=True)` blocks them in the computed attributes execution context (Infrahub API server). They are only permitted to execute on Prefect workers, where an `InfrahubClient` is available. Within Infrahub any Jinja2 based computed attributes that use these new filters should cause a schema violation when loading the schema.
 - **FR-008**: The system MUST provide `from_json` and `from_yaml` Jinja2 filters (adding them only if not already present in the environment) that parse a string into a Python dict/list. Applying them to an empty string MUST return an empty dict without raising. Applying them to malformed content MUST raise `JinjaFilterError`.
 - **FR-009**: `from_json` and `from_yaml` MUST be registered as trusted filters (`trusted=True`) since they perform no external I/O.
 - **FR-010**: All new filters MUST work correctly with `InfrahubClient` (async). `InfrahubClientSync` is not a supported client type for `Jinja2Template`.
 - **FR-011**: All `JinjaFilterError` instances MUST carry an actionable error message that identifies the filter name, the cause of failure, and any remediation hint (for example: "artifact_content requires an InfrahubClient — pass one via Jinja2Template(client=...)").
 - **FR-012**: A new `JinjaFilterError` exception class MUST be added to `infrahub_sdk/template/exceptions.py` as a subclass of `JinjaTemplateError`.
 - **FR-013**: Documentation MUST include a Python transform example demonstrating artifact content retrieval via `client.object_store.get(identifier=storage_id)`. No new SDK convenience method will be added.
+- **FR-014**: If the current user isn't allowed due to a permission denied error to query for the artifact or object file the filter should catch such permission error and raise a Jinja2 error specifically related to the permission issue.
 
 ### Key entities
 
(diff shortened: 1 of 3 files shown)


dev/specs/infp-504-artifact-composition/spec.md
```diff
@@ -0,0 +1,161 @@
...
+
+**Acceptance Scenarios**:
+
+1. **Given** a `Jinja2Template` constructed with a valid `InfrahubClient` and a template calling `storage_id | artifact_content`, **When** the template is rendered with a data dict containing a valid storage_id string, **Then** the output contains the raw string content fetched from the object store.
+2. **Given** the same setup but the storage_id is null or the object store cannot retrieve the content, **When** rendered, **Then** the filter raises a descriptive error indicating the retrieval failure.
+3. **Given** a `Jinja2Template` constructed *without* an `InfrahubClient` and a template calling `artifact_content`, **When** rendered, **Then** an error is raised with a message clearly stating that an `InfrahubClient` is required for this filter.
+4. **Given** a template using `artifact_content` and `validate(restricted=True)` is called, **Then** a `JinjaTemplateOperationViolationError` is raised, confirming the filter is blocked in local restricted mode.
```
dgarros, line 29: Not sure that I understand where this one is coming from

dev/specs/infp-504-artifact-composition/spec.md
```diff
@@ -0,0 +1,161 @@
...
+- What happens when `from_json` or `from_yaml` receives malformed content (invalid JSON/YAML syntax)? `JinjaFilterError` is raised — no silent fallback.
+- What if the same filter name is registered twice (for example, a user-supplied filter that shadows `artifact_content`)? Existing override behaviour should be preserved.
+- File-based templates use a regular `Environment` (not sandboxed); the new filters must be injected correctly in both cases.
+
+## Requirements *(mandatory)*
+
+### Functional requirements
+
+- **FR-001**: `Jinja2Template.__init__` MUST accept an optional `client` parameter of type `InfrahubClient | None` (default `None`). `InfrahubClientSync` is not supported.
+- **FR-002**: A dedicated class (for example, `InfrahubFilters`) MUST be introduced to hold the client reference and expose the Infrahub-specific filter callable methods. `Jinja2Template` instantiates this class when a client is provided and registers its filters into the Jinja2 environment.
```
dgarros, line 97: Not clear to me why we need that

dev/specs/infp-504-artifact-composition/spec.md
```diff
@@ -0,0 +1,161 @@
...
+- **FR-003**: The system MUST provide an `artifact_content` Jinja2 filter that accepts a `storage_id` string and returns the raw string content of the referenced artifact, using the artifact-specific API path.
+- **FR-004**: The system MUST provide a `file_object_content` Jinja2 filter that accepts a `storage_id` string and returns the raw string content of the referenced file object, using the file-object-specific API path or metadata handling — this implementation is distinct from `artifact_content`.
```
dgarros, line 99: is the storage id enough to pull the content of a file object ?

dgarros, line 99: Also, there is no mention of permission in this doc but the object file API will check the permission before returning the content, it feels like it should be mentioned in the spec, event if we just bypass it for now

dgarros, line 99: What if the object return is in binary format and not text based ?

dev/specs/infp-504-artifact-composition/spec.md
```diff
@@ -0,0 +1,161 @@
...
+- **FR-005**: Both `artifact_content` and `file_object_content` MUST raise `JinjaFilterError` when the input `storage_id` is null or empty, or when the object store cannot retrieve the content for any reason (not found, network failure, auth failure).
+- **FR-006**: Both `artifact_content` and `file_object_content` MUST raise `JinjaFilterError` when invoked and no `InfrahubClient` was supplied to `Jinja2Template` at construction time. The error message MUST name the filter and explain that an `InfrahubClient` is required.
+- **FR-007**: Both `artifact_content` and `file_object_content` MUST be registered with `trusted=False` in the `FilterDefinition` registry so that `validate(restricted=True)` blocks them in the computed attributes execution context (Infrahub API server). They are only permitted to execute on Prefect workers, where an `InfrahubClient` is available.
```
dgarros: Could instead of checking at runtime if once of these filters is used in a computed attribute, could we do a sanity check when the user add / update a computed attribute instead ?

dev/specs/infp-504-artifact-composition/spec.md
ogenstad: This part might be a bit weird within this context and goes into the technical aspects but it's really targeting what happens within Infrahub. The idea really is that if anyone uses one of these filters as a computed attribute we'd raise an error during schema import to indicate that it's not allowed.

dev/specs/infp-504-artifact-composition/spec.md
ogenstad: The idea is that these new filters would sit in a class that has access to the SDK itself. I.e. when we use something like `{{ storage_id | artifact_content }}` then `artifact_content` will be a method of a class which has the SDK initialized so that we can use the SDK to collect the information.

dev/specs/infp-504-artifact-composition/spec.md
ogenstad: Yes, although there are other ways that could be applicable to download these files. I'll add a comment around that. I.e. get by HFID and node id

ogenstad: Yes good point. I added this as a requirement as well that it should catch these from the API and provide the correct error back.

ogenstad: I think for now the best approach might be to raise an error, or do you have any other thoughts?

dev/specs/infp-504-artifact-composition/spec.md
ogenstad: Yes for sure, this would be on the side of the Infrahub project but the idea is to keep the current approach of blocking the schema from loading when this happens. But I agree it's vague within the SDK context here.

ogenstad:
@dgarros, I've pushed a new commit with some clarifications along with the comments above.

dev/specs/infp-504-artifact-composition/spec.md
wvandeun: But how would you detect that it is binary data. One challenge is that the storage API endpoint returns an incorrect content-type header, it seems to always return "text/plain".

If we also pass the artifact id or file object id we can validate that the content type defined on them is supported?

wvandeun (dismissed):

dev/specs/infp-504-artifact-composition/spec.md
ogenstad: I'm not completely sure, though I'd say that the first part should be classified as a bug in the API server if that's the case.

In most use cases it would be redundant for users to have to pass in two arguments and also query for both of them. I'd prefer to have an implementation where users don't need to do this even if it could be a bit trickier to implement. What do you think?

gmazoyer: Artifact are always returned as `text/plain` however file objects are not. Also you cannot download a file object through the same endpoint as the artifacts. So there is a way to detect if a file is not plain text by looking at the content type provided in the API response. For artifacts they will always be plain text anyway, we read from the storage as if the artifacts are text anyway.

gmazoyer (approved):

PR #863, opened 2026-03-07
don't allocate from pools on templates

ajtmccarty:
stop trying to process pools for templates b/c we will never provision from pools on a template

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented mutation responses from overwriting pool-allocated attributes and relationships for template nodes, ensuring template resources keep their original pool references while non-template nodes continue to apply server-allocated values.

* **Tests**
  * Added unit tests (sync & async) verifying template handling preserves pool-backed attributes/relationships and that regular nodes allocate from the pool as expected.

### DIFF
--- a/infrahub_sdk/node/node.py
+++ b/infrahub_sdk/node/node.py
@@ -286,7 +286,7 @@ class InfrahubNodeBase:
     def get_raw_graphql_data(self) -> dict | None:
         return self._data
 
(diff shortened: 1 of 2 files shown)


polmichel (approved):

ajtmccarty:
we are not taking the `from_pool` approach any longer. it's going to be `_from_resource_pool` all the way

PR #877, opened 2026-03-19
Add GraphQL fragment support in .infrahub.yml

ogenstad:
# Why

Customers with larger GraphQL queries might want to create smaller snippets with fragments. Both to make the queries themselves more readable but also to allow for reusing fragments in multiple queries. Because of this we're introducing a new section in a `.infrahub.yml` file for `graphql_fragments`

```yaml
queries:
  - name: blue_info
    file_path: blue_info.gql
  - name: character
    file_path: checks/character.gql

graphql_fragments:
  - name: fragments
    file_path: fragments (directory or points to a .gql file)
```

## What changed

* Introduced a new way to read and parse queries that uses the configuration object to determine if there are fragments involved in the query. The new function `render_query()` renders the query as is if it didn't contain any external fragments and if fragments are include they are added to the rendered output.
* Updates to `infrahubctl` to allow for the use of these fragments.
* Various tests

## How to test

The easiest is probably to start with an existing repository containing a `.infrahub.yml` file that you're familiar with and try to render transforms with `infrahubctl`.

Create a new section for `graphql_fragments` and move parts of one of the queries to a fragment file.

## Notes

* Ensure to also update any local `.graphqlrc.yml` files to match any new location where fragment files are stored (to allow for go to definition etc)
* Once this is in place we'll update the Infrahub repo to use this new feature of the SDK to allow for imports of GraphQL queries that use external fragments.

## Release Notes

* **New Features**
  * GraphQL queries now automatically inline referenced fragments for simplified query composition and maintainability.
  * CLI query execution and transforms seamlessly integrate fragment inlining functionality.

* **Improvements**
  * Enhanced error handling for fragment-related issues, including detection of missing, duplicate, and circular dependencies with clear error messages.

* **Documentation**
  * Added comprehensive specification and implementation guide for GraphQL fragment inlining capabilities.

### DIFF
--- a/dev/specs/infp-496-graphql-fragment-inlining/tasks.md
+++ b/dev/specs/infp-496-graphql-fragment-inlining/tasks.md
@@ -1,0 +1,157 @@
+# Tasks: GraphQL Fragment Inlining — SDK Scope
+
+**Input**: `python_sdk/dev/specs/infp-496-graphql-fragment-inlining/spec.md`
+**Parent tasks**: `specs/infp-496-graphql-fragment-inlining/tasks.md` (full feature view including backend)
+**Scope**: All work inside `python_sdk/` only (FR-015: all fragment logic lives in the SDK)
+
+**Path note**: All file paths below are relative to the **infrahub repo root** (e.g.,
+`python_sdk/infrahub_sdk/...`). The `python_sdk/` directory is a git submodule — changes inside
+it must be committed separately from the main infrahub repo.
+
+## Format: `[ID] [P?] [Story] Description`
+
+- **[P]**: Can run in parallel (different files, no dependencies)
+- **[Story]**: Which user story this task belongs to
+- Include exact file paths in descriptions
+
+---
+
+## Phase 1: Setup (Fixture Repository)
+
+**Purpose**: Create the fixture repository used by SDK unit tests and backend component tests.
+All fixture files live inside the `python_sdk` submodule.
+
+- [*] T001 Create fixture repo directory structure at `python_sdk/tests/fixtures/repos/fragment_inlining/` (subdirectories: `fragments/`, `queries/`)
+- [*] T002 [P] Create `python_sdk/tests/fixtures/repos/fragment_inlining/fragments/interfaces.gql` (defines `interfaceFragment`, `portFragment`) and `python_sdk/tests/fixtures/repos/fragment_inlining/fragments/devices.gql` (defines `deviceFragment` that spreads `...interfaceFragment`, and `chassisFragment`)
+- [*] T003 [P] Create `python_sdk/tests/fixtures/repos/fragment_inlining/queries/query_two_files.gql` (spreads `...interfaceFragment` and `...deviceFragment`), `query_no_fragments.gql` (no spreads), `query_transitive.gql` (spreads `...deviceFragment` only), `query_missing_fragment.gql` (spreads `...undeclaredFragment`)
+- [*] T004 Create `python_sdk/tests/fixtures/repos/fragment_inlining/.infrahub.yml` declaring `graphql_fragments` (both fragment files under `fragments/`) and `graphql_queries` (all four query files under `queries/`)
+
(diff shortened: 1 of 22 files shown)


dev/specs/infp-496-graphql-fragment-inlining/tasks.md
```diff
@@ -0,0 +1,157 @@
...
+## Phase 1: Setup (Fixture Repository)
+
+**Purpose**: Create the fixture repository used by SDK unit tests and backend component tests.
+All fixture files live inside the `python_sdk` submodule.
+
+- [ ] T001 Create fixture repo directory structure at `python_sdk/tests/fixtures/repos/fragment_inlining/` (subdirectories: `fragments/`, `queries/`)
+- [ ] T002 [P] Create `python_sdk/tests/fixtures/repos/fragment_inlining/fragments/interfaces.gql` (defines `interfaceFragment`, `portFragment`) and `python_sdk/tests/fixtures/repos/fragment_inlining/fragments/devices.gql` (defines `deviceFragment` that spreads `...interfaceFragment`, and `chassisFragment`)
+- [ ] T003 [P] Create `python_sdk/tests/fixtures/repos/fragment_inlining/queries/query_two_files.gql` (spreads `...interfaceFragment` and `...deviceFragment`), `query_no_fragments.gql` (no spreads), `query_transitive.gql` (spreads `...deviceFragment` only), `query_missing_fragment.gql` (spreads `...undeclaredFragment`)
+- [ ] T004 Create `python_sdk/tests/fixtures/repos/fragment_inlining/.infrahub.yml` declaring `graphql_fragments` (both fragment files under `fragments/`) and `graphql_queries` (all four query files under `queries/`)
```
ajtmccarty: I think phase 1 is done in here

tests/unit/sdk/graphql/test_query_renderer.py
```diff
@@ -0,0 +1,53 @@
...
+
+from infrahub_sdk.exceptions import FragmentNotFoundError
+from infrahub_sdk.graphql.query_renderer import render_query
+from infrahub_sdk.schema.repository import (
+    InfrahubRepositoryConfig,
+    InfrahubRepositoryFragmentConfig,
+    InfrahubRepositoryGraphQLConfig,
+)
+
+FIXTURE_REPO = str(Path(__file__).parents[3] / "fixtures" / "repos" / "fragment_inlining")
```
ajtmccarty: feel brittle, but not sure if there's a better way to do it

infrahub_sdk/graphql/fragment_renderer.py
ajtmccarty, line 1: I think that the functions in here and `query_renderer.py` look like they should be encapsulated in classes with single entry point methods. some logic has to be shared between this file and `query_renderer.py` so that classes would need to have that shared logic injected via another class. then the tests would call the classes instead of piecing the functions together.
certainly not a required change. could be that the separate functions are better for some later phase on this feature. and I know that you are less of a dependency injection lunatic than me. but when I see private functions imported to a different file I have to say something or they revoke my SOLID membership

ajtmccarty (approved):

infrahub_sdk/graphql/fragment_renderer.py
ogenstad: I agree with the private method. I'd moved the query renderer out to it's own file to simplify how users would use this feature, but looking at it I think I'll just get rid of the fragment_renderer.py file and move everything to query_renderer.py.

I'd argue that a class is mostly relevant if there's any shared data where the data itself needs to be modified in multiple ways. In general I don't think that a class based implementation provides additional readability compared to using pure functions though.

tests/unit/sdk/graphql/test_query_renderer.py
ogenstad: I agree, added some constants in `tests/constants.py` to make this cleaner.

ajtmccarty (approved):

PR #882, opened 2026-03-20
Fix/strip nested edged graphql wrapper types

### LINKED ISSUE #881: bug: SDK fails to deserialize NestedEdged GraphQL wrapper types in custom query responses
## Component

- Python SDK

## Infrahub SDK version

1.19.0

## Current Behavior

When using a custom GraphQL query (e.g., in a transform) that includes a cardinality-one relationship like `ip_prefix` on `IpamIPAddress`, the SDK fails to deserialize the response with one of two errors:

1. If `__typename` is included in the query under the relationship: `Error: Unable to find the schema 'NestedEdgedBuiltinIPPrefix'`
2. If `__typename` is removed: `ValueError: Unable to determine the type of the node, __typename not present in data`

Additionally, the `initialized` property on the `RelatedNode` for such relationships always returns `False`, even when the data is present in the GraphQL response.

## Expected Behavior

The SDK should correctly strip GraphQL wrapper type prefixes (`NestedEdged`, `Edged`, etc.) from `__typename` values when looking up schemas. The `RelatedNode` should be properly initialized with the data from the response.

## Steps to Reproduce

1. Define a GraphQL query that includes a cardinality-one relationship (e.g., `ip_prefix` on an IP address node)
2. Include `__typename` in the relationship's response fields
3. Execute the query via a transform or `execute_graphql`
4. The SDK will fail with `SchemaNotFoundError` when trying to look up `NestedEdgedBuiltinIPPrefix`

Example query fragment:
```graphql
ip_prefix {
  __typename
  node {
    ... on IpamIPPrefix {
      __typename
      id
    }
  }
}
```

## Additional Information

The root cause is in `InfrahubNode.from_graphql()`. The method extracts `__typename` from the GraphQL response data:

```python
node_kind = data.get("__typename") or data.get("node", {}).get("__typename", None)
```

For cardinality-one relationships, the GraphQL API returns a `NestedEdged{Kind}` wrapper type (e.g., `NestedEdgedBuiltinIPPrefix`). The outer `__typename` is truthy, so the inner node's actual `__typename` (e.g., `BuiltinIPPrefix`) is never reached. The schema lookup then fails because `NestedEdgedBuiltinIPPrefix` is a GraphQL wrapper type, not a registered schema kind.

The existing code in `RelatedNodeBase.__init__` already strips the `Related` prefix for legacy non-paginated responses, but does not handle `NestedEdged`, `Edged`, `NestedPaginated`, or `Paginated` prefixes.

minitriga:
# Why

  When a custom GraphQL query (e.g. in a transform) includes `__typename` on a cardinality-one relationship, the GraphQL API returns a wrapper type like `NestedEdgedBuiltinIPPrefix`. The
  SDK's `from_graphql()` uses this wrapper typename directly for schema lookup, which fails with `SchemaNotFoundError: Unable to find the schema 'NestedEdgedBuiltinIPPrefix'`. Removing
  `__typename` from the query instead produces `ValueError: Unable to determine the type of the node`. Either way, users cannot deserialize cardinality-one relationships from custom
  queries.

  This PR strips GraphQL wrapper type prefixes (`NestedEdged`, `Edged`, `NestedPaginated`, `Paginated`, `Related`) before schema lookup, so `NestedEdgedBuiltinIPPrefix` correctly resolves
   to `BuiltinIPPrefix`.

  Non-goals: this does not change how the backend generates these wrapper types or how the SDK generates its own queries (which already work correctly).

  Closes: #881

  ## What changed

  - **New utility function `strip_graphql_wrapper_prefix()`** in `node/constants.py` — strips known GraphQL wrapper type prefixes from a typename string to recover the actual schema kind.

  - **`InfrahubNode.from_graphql()` and `InfrahubNodeSync.from_graphql()`** now call `strip_graphql_wrapper_prefix()` on the resolved `node_kind` before schema lookup.

  - **`RelatedNodeBase.__init__`** — replaced the hardcoded `"Related"` prefix strip with the shared `strip_graphql_wrapper_prefix()`, consolidating all prefix handling in one place.

  - **No changes** to query generation, schema handling, or the `_init_relationships` path (which already worked correctly because it extracts `node_data` from the inner dict).

  ## How to review

  1. `infrahub_sdk/node/constants.py` — the new constant and utility function. The prefix list is ordered longest-first so `NestedEdged` matches before `Edged`.
  2. `infrahub_sdk/node/node.py` — two one-line additions in async and sync `from_graphql()`.
  3. `infrahub_sdk/node/related_node.py` — replacing the old `Related` prefix strip with the shared function.
  4. `tests/unit/sdk/test_node.py` — parametrized unit test for the stripping function, plus two integration tests covering `from_graphql` and `_process_relationships` with NestedEdged
  data.

  ## How to test

  ```bash
  cd python_sdk
  uv run pytest tests/unit/sdk/test_node.py::test_strip_graphql_wrapper_prefix tests/unit/sdk/test_node.py::test_from_graphql_with_nested_edged_typename
  tests/unit/sdk/test_node.py::test_process_relationships_with_nested_edged_cardinality_one -xvs

  All 12 new test cases pass. Full existing suite (226 tests) also passes with no regressions.

  Impact & rollout

  - Backward compatibility: Fully backward compatible. The Related prefix was already being stripped; this extends the same treatment to all wrapper prefixes. No API or schema changes.
  - Deployment notes: Safe to deploy independently. No coordination with backend required.

  Checklist

  - Tests added/updated
  - Changelog entry added (changelog/881.fixed.md)
  - External docs updated (if user-facing or ops-facing change)
  - Internal .md docs updated (internal knowledge and AI code tools knowledge)

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

## Bug Fixes

* Fixed SDK deserialization of GraphQL responses containing NestedEdged wrapper types for cardinality-one relationships in custom queries and transforms.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/changelog/881.fixed.md
+++ b/changelog/881.fixed.md
@@ -1,0 +1,1 @@
+Fix SDK failing to deserialize GraphQL responses containing NestedEdged wrapper types for cardinality-one relationships in custom queries and transforms.
--- a/infrahub_sdk/node/constants.py
+++ b/infrahub_sdk/node/constants.py
@@ -35,3 +35,15 @@ HIERARCHY_FETCH_FEATURE_NOT_SUPPORTED_MESSAGE = "Hierarchical fields are not sup
 
 HFID_STR_SEPARATOR = "__"
 PROFILE_KIND_PREFIX = "Profile"
+
+# GraphQL wrapper type prefixes that need to be stripped to get the actual schema kind.
+# Ordered longest-first so that "NestedEdged" is checked before "Edged", etc.
+GRAPHQL_WRAPPER_TYPE_PREFIXES = ("NestedEdged", "NestedPaginated", "Edged", "Paginated", "Related")
+
+
+def strip_graphql_wrapper_prefix(typename: str) -> str:
+    """Strip GraphQL wrapper type prefixes (NestedEdged, Edged, etc.) to get the actual schema kind."""
+    for prefix in GRAPHQL_WRAPPER_TYPE_PREFIXES:
+        if typename.startswith(prefix):
+            return typename[len(prefix) :]
+    return typename
--- a/infrahub_sdk/node/node.py
+++ b/infrahub_sdk/node/node.py
@@ -25,6 +25,7 @@ from .constants import (
     ARTIFACT_GENERATE_FEATURE_NOT_SUPPORTED_MESSAGE,
     FILE_DOWNLOAD_FEATURE_NOT_SUPPORTED_MESSAGE,
     PROPERTIES_OBJECT,
+    strip_graphql_wrapper_prefix,
 )
 from .metadata import NodeMetadata
 from .related_node import RelatedNode, RelatedNodeBase, RelatedNodeSync
@@ -618,6 +619,7 @@ class InfrahubNode(InfrahubNodeBase):
             node_kind = data.get("__typename") or data.get("node", {}).get("__typename", None)
             if not node_kind:
                 raise ValueError("Unable to determine the type of the node, __typename not present in data")
+            node_kind = strip_graphql_wrapper_prefix(node_kind)
             schema = await client.schema.get(kind=node_kind, branch=branch, timeout=timeout)
 
         return cls(client=client, schema=schema, branch=branch, data=cls._strip_alias(data))
@@ -1501,6 +1503,7 @@ class InfrahubNodeSync(InfrahubNodeBase):
             node_kind = data.get("__typename") or data.get("node", {}).get("__typename", None)
             if not node_kind:
                 raise ValueError("Unable to determine the type of the node, __typename not present in data")
+            node_kind = strip_graphql_wrapper_prefix(node_kind)
             schema = client.schema.get(kind=node_kind, branch=branch, timeout=timeout)
 
         return cls(client=client, schema=schema, branch=branch, data=cls._strip_alias(data))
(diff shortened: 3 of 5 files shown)


ajtmccarty:
I'm wondering if this is something that we should address on the backend. I think we need the `__typename` for GraphQL reasons (or maybe GraphQL creates the `__typename`, I'm not totally sure). but it seems reasonable that the backend would be able to tell you the `kind` of any objects it returns
@ogenstad any thoughts on this one?

ogenstad:
> I'm wondering if this is something that we should address on the backend. I think we need the `__typename` for GraphQL reasons (or maybe GraphQL creates the `__typename`, I'm not totally sure). but it seems reasonable that the backend would be able to tell you the `kind` of any objects it returns @ogenstad any thoughts on this one?

Graphene inserts the correct type for `__typename` where ever it is inserted in the query. For nodes it will typically reflect the actual node being returned. Given the current state I think the fix makes sense. However, I think the main problem is that there's too much magic going on within the SDK, where we're stripping out a prefix based on a string. While it mostly works it would completely fall apart if someone were to try to have a namespace named `Edged` which on the surface is completely valid. When parsing these results it would be better if the SDK used its knowledge about the schema to get the correct type.

@minitriga, I think what's missing from the original bug request is the full query that's being used and how this gets executed by the SDK. Is this something that we'd get with a normal .filters() query or how did you run into the bug?

minitriga:
@ogenstad I was supposed to close this. Issue was the query being used.

PR #885, opened 2026-03-23
Add plan to implement INFP-504

gmazoyer:
This PR only provides the implementation plan for INFP-504 - Artifact composition. It proposes a way to implement Jinja2 filters in order to be able to include the content of an artifact or a file object within another one.

## Release Notes

* **Documentation**
  * Added comprehensive specifications and quickstart guides for artifact and file content composition in templates
  * Documented new template filter capabilities and context-based execution control system for filter availability


dev/specs/infp-504-artifact-composition/data-model.md
```diff
@@ -0,0 +1,166 @@
...
+**Inheritance**: `Error` → `JinjaTemplateError` → `JinjaFilterError`
+
+### InfrahubFilters (new class)
+
+**Location**: `infrahub_sdk/template/infrahub_filters.py` (new file)
+
+```python
+class InfrahubFilters:
+    def __init__(self, client: InfrahubClient) -> None:
+        self.client = client
```
ogenstad, line 69: I'm not sure that it's correct, but the way I saw this was that a `client` param here would be optional. That way we can initialize a Template class the same way and then if we specifically want to use these filters we'd use some setter afterwards to add the client. However this might not be required depending on how we end up populating these filters into the template class in various circumstances.

PR #889, opened 2026-03-24
IFC-2275 Add Jinja2 filters for artifact composition

gmazoyer:
# Why

Customers building modular configuration pipelines need to compose larger artifacts from smaller sub-artifacts inside Jinja2 transforms, without duplicating template logic or GraphQL query fields. Today there is no way for a template to reference the rendered content of another artifact.

This PR implements the SDK side of the "Artifact of Artifacts" initiative, adding new Jinja2 filters and the supporting infrastructure to enable artifact content composition.

## What changed

- `artifact_content` that accepts a `storage_id` string, fetches artifact content from the object store via `ObjectStore.get()`
- `file_object_content` with the same pattern but uses the `/api/files/by-storage-id/{storage_id}` endpoint, with binary content-type rejection
- `from_json` / `from_yaml` to parse strings into Python dicts/lists, enabling chaining like `{{ sid | artifact_content | from_json }}`
- Replaced `FilterDefinition.trusted: bool` with `allowed_contexts: ExecutionContext` using Python's `Flag` enum
- Three contexts: `CORE` (API server computed attributes, most restrictive), `WORKER` (Prefect workers), `LOCAL` (CLI/unrestricted)
- New optional `client: InfrahubClient` constructor parameter
- New `set_client()` method for deferred client injection (per PR #885 feedback)
- Without a client, client-dependent filters raise `JinjaFilterError` with an actionable message
- `JinjaFilterError(filter_name, message, hint)` subclassing `JinjaTemplateError`
- New `get_file_by_storage_id()` method (async + sync) using `/api/files/by-storage-id/{storage_id}`
- Content-type validation: rejects non-text responses (`application/octet-stream`, etc.)

### Suggested review order
1. `infrahub_sdk/template/filters.py` for `ExecutionContext` enum and `FilterDefinition` migration
2. `infrahub_sdk/template/infrahub_filters.py` for `InfrahubFilters` class, `from_json`, `from_yaml`
3 `infrahub_sdk/template/__init__.py` for `Jinja2Template` wiring, `set_client()`, `validate()` changes
4. `infrahub_sdk/object_store.py` for `get_file_by_storage_id()`

## How to review

- `validate()` context logic iit is the core of the trust model change
- `_register_client_filters()` and `set_client()` ensure client-dependent filters are correctly bound or fallback

## How to test

```bash
uv run pytest tests/unit/sdk/test_infrahub_filters.py -v
uv run pytest tests/unit/sdk/test_template.py -v
[...]

PR #904, opened 2026-03-30
Add file_object_content_by_id and file_object_content_by_hfid filters

gmazoyer:
## Why & What

This is a change following https://github.com/opsmill/infrahub-sdk-python/pull/889.

Support all three file object retrieval endpoints: by storage ID (existing), by node UUID, and by HFID. Extract shared `ObjectStore` file fetching and content-type validation into helpers.

Make `InfrahubFilters` accept an optional client, checking at call time instead of maintaining separate no-client fallback functions.

## Checklist

- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Added template filters to retrieve file content by node ID and by HFID; filters validate required identifier and kind. Templates can defer providing a client and set it later.

* **Bug Fixes**
  * Unified input validation and clearer template errors for missing inputs, permission/auth failures, and non-text (binary) content rejection.

* **Documentation**
  * Expanded templating docs with execution contexts, per-context filter availability tables, and usage examples.

* **Tests**
  * Updated tests to exercise deferred-client behavior and HFID/kind validation.


infrahub_sdk/template/__init__.py
```diff
@@ -59,26 +58,20 @@ def __init__(
 
     def set_client(self, client: InfrahubClient) -> None:
         """Set or replace the InfrahubClient used by client-dependent filters."""
-        self._register_client_filters(client=client)
+        self._infrahub_filters.client = client
```
ogenstad: Would be a bit cleaner if the filters object also had a method to set the client instead of assigning to `.client` directly on the "child object".

docs/_templates/sdk_template_reference.j2
```diff
@@ -26,6 +52,67 @@ The following Jinja2 filters from <a href="https://netutils.readthedocs.io">Netu
...
 
+## Infrahub filters
+
+These filters are provided by the Infrahub SDK for artifact and file object content composition. The table shows which execution contexts each filter is allowed in.
+
+<!-- vale off -->
+| Name | CORE | WORKER | LOCAL |
+| ---- | ---- | ------ | ----- |
+{% for filter in infrahub %}
+| `{{ filter.name }}` | {% if filter.core %}✅{% else %}❌{% endif %} | {% if filter.worker %}✅{% else %}❌{% endif %} | {% if filter.local %}✅{% else %}❌{% endif %} |
```
ogenstad, line 64: Would it make sense to report on all filters using the same kind of table? I.e. for the previous table we only have the "trusted" option? Might be easier to understand if the format for the trust model is the same for all filter types?

ogenstad (commented):
LGTM. Looks like CodeRabbit has a point with the comment around "Edge case: empty strings within HFID list are not validated."

ogenstad (approved):

PR #905, opened 2026-03-30
improve InfrahubGeneratorDefintionConfig descriptions

wvandeun:
## Summary by CodeRabbit

* **Documentation**
  * Clarified generator configuration: GraphQL variable names now map to target attributes using double-underscore notation.
  * Specified that the generator target field is required and that group members become individual generator targets (one run per member).
  * Clarified the class name meaning (Python class extending the generator) and refined descriptions of execution-related options and response conversion.

### DIFF
--- a/infrahub_sdk/schema/repository.py
+++ b/infrahub_sdk/schema/repository.py
@@ -88,21 +88,28 @@ class InfrahubGeneratorDefinitionConfig(InfrahubRepositoryConfigElement):
     file_path: Path = Field(..., description="The file within the repository with the generator code.")
     query: str = Field(..., description="The GraphQL query to use as input.")
     parameters: dict[str, Any] = Field(
-        default_factory=dict, description="The input parameters required to run this check"
+        default_factory=dict,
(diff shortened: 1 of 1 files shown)


infrahub_sdk/schema/repository.py
```diff
@@ -88,21 +88,28 @@ class InfrahubGeneratorDefinitionConfig(InfrahubRepositoryConfigElement):
...
+    class_name: str = Field(
+        default="Generator",
+        description="The name of the Python class within file_path that extends InfrahubGenerator.",
     )
-    targets: str = Field(..., description="The group to target when running this generator")
-    class_name: str = Field(default="Generator", description="The name of the generator class to run.")
     convert_query_response: bool = Field(
         default=False,
-        description="Decide if the generator should convert the result of the GraphQL query to SDK InfrahubNode objects.",
+        description="When true, converts the raw GraphQL dict into SDK InfrahubNode objects accessible via self.nodes and self.store.",
```
ajtmccarty, line 104: the descriptions below indicate that the default is `true`. should this say that the default is `false`?

ajtmccarty (approved):

PR #910, opened 2026-03-31
Clean up of logic within Node

ogenstad:
# Why

The node initialization code in InfrahubNodeBase was determining capability flags (_artifact_support, _file_object_support, _hierarchy_support, _artifact_definition_support) by inspecting schema internals directly — using getattr to reach into inherit_from, string comparisons against hardcoded kind names, and isinstance checks to exclude schema types that don't carry certain fields. This logic belonged on the schema layer, not the node layer.

The change moves these capability checks to where the relevant data lives: BaseSchema provides safe defaults (False / []) for all schema types, and NodeSchemaAPI overrides them with real logic using its own fields (inherit_from, hierarchy). The same pattern is applied to hierarchical_relationship_schemas — rather than constructing four hardcoded RelationshipSchemaAPI objects inline in _init_relationships (duplicated across the async and sync classes), the schema itself now produces the correct pseudo-schemas for hierarchical nodes.

The result is that node.py no longer needs to know anything about the structure of the schema to determine what a node supports. Four uniform assignments replace a mix of getattr hacks, isinstance guards, and inline schema construction. The _init_relationships hierarchical block drops from ~70 lines (×2 for async/sync) to a single loop.

## What changed

* infrahub_sdk/schema/main.py: define various generic properties related to the schema
* infrahub_sdk/node/node.py: refactor to use the new properties to simplify the code.

## Summary by CodeRabbit

* **New Features**
  * Added helper properties on schemas to expose capability and cardinality information for more accurate feature detection.

* **Improvements**
  * Improved node capability detection to support broader and more flexible schema configurations.
  * Simplified hierarchical relationship initialization for more consistent behavior, fewer runtime checks, and easier maintenance.

### DIFF
--- a/infrahub_sdk/node/node.py
+++ b/infrahub_sdk/node/node.py
@@ -11,11 +11,8 @@ from ..file_handler import FileHandler, FileHandlerBase, FileHandlerSync, Prepar
 from ..graphql import Mutation, Query
 from ..schema import (
     GenericSchemaAPI,
-    ProfileSchemaAPI,
     RelationshipCardinality,
     RelationshipKind,
-    RelationshipSchemaAPI,
-    TemplateSchemaAPI,
 )
 from ..utils import compare_lists, generate_short_id
 from .attribute import Attribute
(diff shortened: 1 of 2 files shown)


infrahub_sdk/node/node.py
```diff
@@ -11,11 +11,8 @@
 from ..graphql import Mutation, Query
 from ..schema import (
     GenericSchemaAPI,
-    ProfileSchemaAPI,
     RelationshipCardinality,
     RelationshipKind,
-    RelationshipSchemaAPI,
-    TemplateSchemaAPI,
 )
```
ajtmccarty, line 16: 🥇

ajtmccarty (approved):
coderabbit might have a couple legitimate comments, but, other than that, it looks great. big improvement

ogenstad:
> coderabbit might have a couple legitimate comments, but, other than that, it looks great. big improvement

@ajtmccarty, yeah the reason for the major one was to get rid of an if-statement within the `__init__()` method that also included those isinstance statements (I'm not a big fan of having different code paths within these init methods..). I think this could perhaps be solved by adding some docstrings for those properties to indicate that it's to ensure that we have the same methods on all schema nodes but that they will always return false/negative. What do you think about that?

Will have to check on the hierarchy thing.

ajtmccarty:
> > coderabbit might have a couple legitimate comments, but, other than that, it looks great. big improvement
> 
> @ajtmccarty, yeah the reason for the major one was to get rid of an if-statement within the `__init__()` method that also included those isinstance statements (I'm not a big fan of having different code paths within these init methods..). I think this could perhaps be solved by adding some docstrings for those properties to indicate that it's to ensure that we have the same methods on all schema nodes but that they will always return false/negative. What do you think about that?
> 
> Will have to check on the hierarchy thing.

sounds good to me

PR #927, opened 2026-04-08
Raise timeout for integration tests

ogenstad:
# Why

With all of the new integration tests and how they run in Docker the current timeout of 30 minutes seems to not be enough.

## What changed

Raise timeout of integration tests from 30 to 40 minutes.

## Summary by CodeRabbit

* **Chores**
  * Improved CI/CD pipeline configuration for test reliability.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -324,7 +324,7 @@ jobs:
     needs: ["prepare-environment", "files-changed", "yaml-lint", "python-lint"]
     runs-on:
       group: "huge-runners"
-    timeout-minutes: 30
+    timeout-minutes: 40
     steps:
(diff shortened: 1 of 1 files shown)


solababs (approved):

.github/workflows/ci.yml
polmichel, line 365: Maybe should we also modify this one?
```suggestion
  #   timeout-minutes: 40
```

polmichel: If some day we uncomment this code

ogenstad: I'll leave that as it is for now. I'm sure we'll do additional reviews if we later want to enable that one.

PR #928, opened 2026-04-08
Fix authentication validation to allow password auth to override env api_token

### LINKED ISSUE #654: bug: Having a token exported in env creates pydantic exception when logging with password
### Component

Python SDK

### Infrahub SDK version

1.5

### Current Behavior

If I have this in my env variables:

```
export | grep INFRAHUB
INFRAHUB_ADDRESS=http://localhost:8000
INFRAHUB_API_TOKEN=06438eb2-8019-4776-878c-0941b1f1d1ec
```

Then try to run this piece of code:

```
# Test connection and authentication
def hello_world():
    client = InfrahubClientSync(
        config=Config(
            address="https://sandbox.infrahub.app",
            username="otto",
            password="infrahub",
        )
    )
```

It creates this exception

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for Config
  Value error, Unable to combine password with token based authentication [type=value_error, input_value={'address': 'https://sand... 'password': 'infrahub'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/value_error
```

### Expected Behavior

Either a message clearer or in my opinion what's is in the code should be followed

### Steps to Reproduce

- Export token in env
- Try to create a client object with username and password

### Additional Information

_No response_

ogenstad:
# Why

Setting authentication parameters directly in a `Config` object should override conflicting settings set from environment variables. I.e. if you have an environment variable for a token but start a Config and directly specify username and password this should override the environment variable setting.

This can be a source of annoyance when running tests as well as with local development.

Closes #654

Replaces #655

## What changed

* New logic within the config class to ensure that directly assigned values override environment variables
* New tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication precedence fixed: explicitly configured credentials now override environment-derived values; mixed token/password inputs are disambiguated so only the intended method is used.

* **Tests**
  * Added/updated tests covering credential precedence between explicit configuration and environment variables.

* **Chores**
  * Added changelog entry documenting direct-assignment authentication behavior.

### DIFF
--- a/changelog/654.fixed.md
+++ b/changelog/654.fixed.md
@@ -1,0 +1,1 @@
+Allow direct assignment of authentication method to the configuration to override settings from environment variables.
--- a/infrahub_sdk/config.py
+++ b/infrahub_sdk/config.py
@@ -5,7 +5,7 @@ from copy import deepcopy
 from typing import Any
 
 from pydantic import Field, PrivateAttr, field_validator, model_validator
-from pydantic_settings import BaseSettings, SettingsConfigDict
+from pydantic_settings import BaseSettings, InitSettingsSource, PydanticBaseSettingsSource, SettingsConfigDict
 from typing_extensions import Self
 
 from .constants import InfrahubClientMode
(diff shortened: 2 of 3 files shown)


polmichel (approved):
LGTM! Just a small comment/thought, if we define `INFRAHUB_API_TOKEN` and `INFRAHUB_PASSWORD` and if we only send the `username` through the config object, the code seems to prioritize the username/password authentication, which, in my opinion, makes sense.

It might be useful to add a unit test for this case to make this use case clearer? But this is obviously not blocking and can be done later on.
```
def test_password_auth_overrides_env_token_when_password_env_var_and_username_explicit(
    monkeypatch: pytest.MonkeyPatch,
) -> None:
    """Test that explicit username/password overrides INFRAHUB_API_TOKEN from environment when only username is provided
    through config object and password is provided through environment variable"""

    # Set environment variable for api_token and password
    monkeypatch.setenv("INFRAHUB_API_TOKEN", "token-from-env")
    monkeypatch.setenv("INFRAHUB_PASSWORD", "testpass")

    # Create configuration with explicit username
    config = Config(address="https://sandbox.infrahub.app", username="testuser")

    # Password auth should be active and api_token should be cleared
    assert config.username == "testuser"
    assert config.password == "testpass"
    assert config.api_token is None
    assert config.password_authentication is True

```

ogenstad:
> LGTM! Just a small comment/thought, if we define `INFRAHUB_API_TOKEN` and `INFRAHUB_PASSWORD` and if we only send the `username` through the config object, the code seems to prioritize the username/password authentication, which, in my opinion, makes sense.

Good point! I added this test.

polmichel (approved):

PR #931, opened 2026-04-09
Move CRUD commands under object subgroup

gmazoyer:
## Why & What

The get, create, update, and delete commands were registered at the root of infrahubctl, which felt out of place next to unrelated top-level commands like check, render, and transform. They naturally belong under `infrahubctl object` alongside load and validate, since they all operate on Infrahub objects. Tests updated accordingly.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry added (`uv run towncrier create ...`)
- [x] External docs updated (if user-facing or ops-facing change)
- [x] Internal .md docs updated (internal knowledge and AI code tools knowledge)

## Summary by CodeRabbit

* **Documentation**
  * Reorganized CLI command documentation: `get`, `create`, `update`, and `delete` commands now accessed via the `object` subcommand (e.g., `infrahubctl object get`).
  * Added developer guides for CLI architecture and documentation generation workflow.

* **Breaking Changes**
  * Root-level CLI commands (`get`, `create`, `update`, `delete`) moved to `infrahubctl object` subcommand group.


ogenstad (approved):
Thanks! LGTM, as a future refactoring I think it could make sense to also move the commands into some other module so that they are grouped together.

PR #937, opened 2026-04-10
feat(infrahubctl): add telemetry commands

fatih-acar:
## Summary by CodeRabbit

* **New Features**
  * Added a top-level telemetry CLI command group:
    * list — view telemetry snapshots with ISO 8601 start/end filters and limit; displays summaries and a message when no snapshots are found
    * export — export snapshots to JSON with output path and ISO 8601 date filters; handles full export across pages and confirms when complete

* **Documentation**
  * Added telemetry CLI docs covering usage, options, defaults, and examples

### DIFF
--- a/infrahub_sdk/ctl/telemetry.py
+++ b/infrahub_sdk/ctl/telemetry.py
@@ -1,0 +1,128 @@
+from __future__ import annotations
+
+import json
+from datetime import datetime
+from pathlib import Path
+from typing import Any
+from urllib.parse import urlencode
+
+import typer
+from rich.console import Console
+from rich.table import Table
+
+from ..async_typer import AsyncTyper
+from .client import initialize_client
+from .parameters import CONFIG_PARAM
+from .utils import catch_exception
+
+app = AsyncTyper()
+console = Console()
+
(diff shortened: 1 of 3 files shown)


infrahub_sdk/ctl/telemetry.py
```diff
@@ -0,0 +1,96 @@
...
+
+app = AsyncTyper()
+console = Console()
+
+
+@app.command(name="list")
+@catch_exception(console=console)
+async def list_snapshots(
+    start_date: str | None = typer.Option(None, help="Start date filter (ISO 8601)"),
+    end_date: str | None = typer.Option(None, help="End date filter (ISO 8601)"),
```
ajtmccarty: I think `typer` supports Python `datetime` objects. at least this makes it look that way https://typer.tiangolo.com/tutorial/parameter-types/datetime/

if that doesn't work, then maybe there should be some sort of validation of some sort before sending the request to the api.

ajtmccarty (approved):

infrahub_sdk/ctl/telemetry.py
fatih-acar: Good catch, thanks!

PR #942, opened 2026-04-13
Let the server validate in object CRUD commands

gmazoyer:
## Why & What

The `object create/delete` commands were doing more work than they needed to. Most of that work was either unreliable or not necessary. This cleans it up and records the underlying principle so we don't reintroduce the same patterns.

`create` used to do a pre-flight `resolve_node` round-trip to figure out whether the object already existed, just so it could print `Created` vs `Updated` after the upsert. It wasted a network call but also wrongly fell back to a hardcoded `"name"` field and leaned on `default_filter`, which is on its way out.

`resolve_node` had a "fallback" path at the end that called `client.get(id=<not-a-uuid>)` purely to piggyback on the SDK's error message. That's now a direct `raise`.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry added
- [ ] External docs updated
- [x] Internal .md docs updated (new `cli-design-principles.md`)

### DIFF
--- a/dev/knowledge/cli-design-principles.md
+++ b/dev/knowledge/cli-design-principles.md
@@ -1,0 +1,45 @@
+# CLI Design Principles
+
+Guidelines for writing `infrahubctl` commands. These complement the structural rules in `cli-architecture.md`.
+
+## Don't pre-validate what the server validates
+
+The CLI's job is to translate user intent into API calls, not to reimplement server-side logic. When in doubt, just fire the request and let the server respond.
+
+**Avoid**:
+
+- Extra round-trips to check whether an object already exists before creating it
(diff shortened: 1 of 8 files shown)


dev/knowledge/cli-design-principles.md
```diff
@@ -0,0 +1,45 @@
+# CLI Design Principles
+
+Guidelines for writing `infrahubctl` commands. These complement the structural rules in `cli-architecture.md`.
+
+## Don't pre-validate what the server validates
+
+The CLI's job is to translate user intent into API calls, not to reimplement server-side logic. When in doubt, just fire the request and let the server respond.
```
ajtmccarty, line 7: I agree!

ajtmccarty (approved):

PR #948, opened 2026-04-14
Bring `infrahub-develop` changes into `develop`

gmazoyer:


ogenstad:
Hm, I think I might have misunderstood you during the daily we typically never merge back `infrahub-develop` to `develop`. I.e. at the time when we release Infrahub 1.9 we should merge `infrahub-develop` into `stable` and bump the version of the SDK. What was it we wanted to do here?

gmazoyer:
> Hm, I think I might have misunderstood you during the daily we typically never merge back `infrahub-develop` to `develop`. I.e. at the time when we release Infrahub 1.9 we should merge `infrahub-develop` into `stable` and bump the version of the SDK. What was it we wanted to do here?

My bad, I thought we were doing `infrahub-develop` → `develop` → `stable`. I'll close this PR and just use the proper commit from `infrahub-develop` for the submodule used by infrahub for now.

PR #952, opened 2026-04-21
Add `infrahubctl marketplace` CLI for fetching schemas and collections

minitriga:
## Summary

Adds a new `infrahubctl marketplace` sub-app with a `get` command that pulls schemas and collections from the public Infrahub Marketplace (`marketplace.infrahub.app`) into a local directory or to stdout.

```bash
infrahubctl marketplace get acme/network-base               # any item, auto-detected
infrahubctl marketplace get acme/network-base -v 1.2.0      # pin a schema version
infrahubctl marketplace get acme/starter-pack               # collection resolved automatically
infrahubctl marketplace get acme/network-base --stdout      # print to stdout, status to stderr
infrahubctl marketplace get acme/foo -o ./infra/schemas     # custom destination
```

## What the command does

- **REST-based** — talks to the public marketplace REST API (`/api/v1/schemas/...`, `/api/v1/collections/...`); no GraphQL involved.
- **Auto-detect schema vs collection** — probes both endpoints in parallel. One command handles either item type without the caller needing to know which it is. On a rare `namespace/name` collision, schema wins with a warning; `--collection/-c` forces the other path.
- **Version pinning** — `--version <semver>` pins the download to a specific published schema version; default is whatever the marketplace resolves as latest (echoed back on success).
- **Stdout mode** — `--stdout/-s` writes content to real stdout and routes status, warnings, and errors to stderr, so the output can be piped into other tools (e.g. `infrahubctl schema check` once it grows stdin support). Collections are emitted as multi-doc YAML, with `---` separators only inserted when a schema doesn't already start with one. No files are written when `--stdout` is set.
- **Custom destination** — `--output-dir/-o` (default `./schemas`) controls where files land; missing parent directories are created.
- **Marketplace override** — `--marketplace-url`, `infrahubctl.toml` `marketplace_url`, or `INFRAHUB_MARKETPLACE_URL` env var let you point at staging or a local instance.
- **Clean error taxonomy** — distinct failure classes (invalid input, not found, network) with exit code `1` for deterministic failures and `2` for transient/network so CI can branch on it. No tracebacks on an unwritable output directory, a missing schema, or a connection refused.

## Options

| Option | Default | Purpose |
|--------|---------|---------|
| `IDENTIFIER` (positional) | required | `namespace/name` — a schema or collection |
| `-v, --version TEXT` | latest | Pin to a specific published schema version (schemas only) |
| `-c, --collection` | auto-detect | Force the collection code path |
| `-s, --stdout` | off | Print to stdout, status to stderr; skips disk writes |
| `-o, --output-dir PATH` | `./schemas` | Where to write files |
| `--marketplace-url TEXT` | `https://marketplace.infrahub.app` | Override the marketplace host |

## Behaviour notes

- Output lines state the resolved item type (`Downloaded schema …` / `Collection …`) so the user can detect an unintended match in a collision case.
- Passing `--version` alongside a collection identifier prints a warning and proceeds with the collection download, rather than failing.
- A missing version on an existing schema reports "no published version '<v>'" with a hint to drop `--version`, distinct from "schema not found".

## Out of scope

No `--load` convenience flag. `get` is fetch-only — chain with `infrahubctl schema load <dir>` (or pipe via `--stdout` once stdin support lands) to push the result into a running Infrahub.

## Test plan

- [x] `uv run pytest tests/unit/ctl/test_marketplace_app.py` — 20 tests covering auto-detect, collision, error classes, `--version`, `--output-dir`, `--marketplace-url`, `--collection` override, and `--stdout` mode (single schema, collection, separator injection).
- [x] `uv run invoke lint` — ruff, ty, mypy, vale, markdownlint all clean.
- [x] `uv run invoke generate-infrahubctl` — `docs/docs/infrahubctl/infrahubctl-marketplace.mdx` regenerated.
- [x] End-to-end smoke test against the live marketplace: `infrahubctl marketplace get nonexistent/nonexistent` returns the expected not-found message with exit 1.

## Artefacts

Full spec, plan, research, contracts, quickstart, and task breakdown under `specs/001-marketplace-api-update/` for reviewers who want the design rationale (auto-detect strategy evaluation, error taxonomy decisions, collision precedence).


infrahub_sdk/ctl/marketplace.py
```diff
@@ -0,0 +1,248 @@
...
+        return "schema", schema_resp  # type: ignore[return-value]
+    if collection_ok:
+        return "collection", collection_resp  # type: ignore[return-value]
+
+    def is_transport_failure(r: object) -> bool:
+        if isinstance(r, Exception):
+            return True
+        return isinstance(r, httpx.Response) and r.status_code >= 500
+
+    if is_transport_failure(schema_resp) and is_transport_failure(collection_resp):
```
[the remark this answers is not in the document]
minitriga: Fixed in 76f1948 — changed the guard to `or` so a transport failure on either probe (when no 200 winner exists) raises `network` rather than falling through to `not-found`. Added `test_autodetect_partial_probe_failure_is_network` to cover the schema-404 + collection-transport-failure case.

infrahub_sdk/ctl/marketplace.py
```diff
@@ -0,0 +1,248 @@
...
+    async with httpx.AsyncClient(follow_redirects=True) as client:
+        prefetched: httpx.Response | None = None
+        if collection is None:
+            item_type, prefetched = await _detect_item_type(client, base_url, namespace, name)
+        elif collection:
+            item_type = "collection"
+        else:
+            item_type = "schema"
+
+        if item_type == "collection":
```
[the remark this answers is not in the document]
minitriga: Fixed in 76f1948 — wrapped the post-detect download dispatch in `try/except httpx.HTTPError`, re-raising via `_fail("network", ...)` so transport failures and `raise_for_status` 5xxs during the actual fetch exit 2 instead of bubbling up to the default `@catch_exception` (exit 1). Added `test_versioned_download_network_error` and `test_collection_flag_network_error` to cover the two paths that make a fresh request after auto-detect (versioned schema fetch and explicit `--collection`).

dev/specs/001-marketplace-api-update/quickstart.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+This must exercise the same auto-detection behaviour against the overridden host.
+
+## Expected success output shape
+
+For a schema:
+
+```text
+Downloaded acme/network-base v1.2.0 -> schemas/network-base.yml
```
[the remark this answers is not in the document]
minitriga: Fixed in 76f1948 — updated the schema success example to `Downloaded schema acme/network-base v1.2.0 -> schemas/network-base.yml` to match the actual CLI output.

infrahub_sdk/ctl/config.py
```diff
@@ -27,6 +27,7 @@ class Settings(BaseSettings):
     server_address: str = Field(default="http://localhost:8000", validation_alias="infrahub_address")
     api_token: str | None = Field(default=None)
     default_branch: str = Field(default="main")
+    marketplace_url: str = Field(default="https://marketplace.infrahub.app")
```
ogenstad: I'd like to get rid of the config for `infrahubctl` as it's pretty much a limited replica of what we have from the SDK config. Can we move this entry to that file instead? Sometime soon I'd like to ditch this file alltogether.

infrahub_sdk/ctl/marketplace.py
```diff
@@ -0,0 +1,294 @@
...
+
+    if schema_ok:
+        if collection_ok:
+            _status(stdout).print(
+                f"[yellow]Note: '{namespace}/{name}' exists as both a schema and a collection. "
+                "Resolving as schema. Pass --collection to force the collection path."
+            )
+        return "schema", schema_resp  # type: ignore[return-value]
+    if collection_ok:
+        return "collection", collection_resp  # type: ignore[return-value]
```
ogenstad: Can we resolve these added type ignores and ensure that we return the correct types?

infrahub_sdk/ctl/marketplace.py
```diff
@@ -0,0 +1,294 @@
...
+
+    By default, auto-detects whether `namespace/name` is a schema or a collection.
+    Pass --collection to force the collection path when an identifier exists as both.
+    """
+    namespace, name = _parse_identifier(identifier)
+
+    resolved_url = marketplace_url or config.SETTINGS.active.marketplace_url
+    base_url = resolved_url.rstrip("/")
+
+    async with httpx.AsyncClient(follow_redirects=True) as client:
```
ogenstad: A consideration here is that when we use a fresh AsyncClient directly like this it doesn't benefit from the configuration we have within the SDK. In this case it would mean that users behind a proxy might not be able to use this part of the tool. As such we should have methods within the client to perform these actions instead. It could be that we need some updates there with regards to the `follow_redirects` param.

infrahub_sdk/ctl/marketplace.py
```diff
@@ -0,0 +1,294 @@
...
+err_console = Console(stderr=True)
+
+MarketplaceItemType = Literal["schema", "collection"]
+ErrorClass = Literal["invalid-input", "not-found", "network"]
+
+_ERROR_EXIT_CODES: dict[ErrorClass, int] = {
+    "invalid-input": 1,
+    "not-found": 1,
+    "network": 2,
+}
```
ogenstad: The above looks like a very loose contract between ErrorClass and the Literals. While there are only a few it still looks brittle as it would fail if anything else was added to ErrorClass later. Just looking at this in isolation this looks like it could have been an Enum where the error codes were the values.

infrahub_sdk/ctl/config.py
minitriga: Done — moved `marketplace_url` to `ConfigBase` in `infrahub_sdk/config.py` (readable via `INFRAHUB_MARKETPLACE_URL`) and removed it from `infrahubctl` `Settings`. `marketplace.py` now instantiates `ConfigBase` directly, so the direct `config` import is gone from that file too.

infrahub_sdk/ctl/marketplace.py
minitriga: Resolved — replaced the `schema_ok`/`collection_ok` booleans with inline `isinstance(r, httpx.Response) and r.status_code == 200` guards directly in the `if` conditions. The type checker narrows to `httpx.Response` naturally within each branch, so no `cast` or `TypeGuard` needed. mypy and ty both confirm no issues.

infrahub_sdk/ctl/marketplace.py
minitriga: Addressed — added `_make_http_client(sdk_cfg)` which builds the `AsyncClient` using `ConfigBase`'s proxy (`proxy` / `proxy_mounts`) and `tls_context`, matching what `InfrahubClient._build_proxy_config()` does internally. Also added `follow_redirects=True` as you noted. The `get()` command now instantiates `ConfigBase` once and passes it to both the client builder and the URL default.\n\nAgreed that moving the download logic into client methods would be the cleanest end state — happy to do that as a follow-up once the shape is settled.

infrahub_sdk/ctl/marketplace.py
minitriga: Replaced with `_ErrorClass(Enum)` where each member carries an `exit_code` property — the mapping is now co-located with the error definition rather than split across two declarations. Adding a new error class without a matching exit code is a type error rather than a silent runtime gap.

infrahub_sdk/config.py
```diff
@@ -70,6 +70,9 @@ class ConfigBase(BaseSettings):
     )
     proxy: str | None = Field(default=None, description="Proxy address")
     proxy_mounts: ProxyMountsConfig = Field(default=ProxyMountsConfig(), description="Proxy mounts configuration")
+    marketplace_url: str = Field(
```
ogenstad, line 73: Is this configuration actually used anywhere? It seems like the app is using either the parameter defined when running the command line or the entry in the infrahubctl config which shadows this one.

It looks like this is just dead code now, but could be fine if we later refactor infrahubctl and remove that config section alltogether.

infrahub_sdk/ctl/marketplace.py
```diff
@@ -0,0 +1,331 @@
...
+
+@app.command()
+@catch_exception(console=console)
+async def get(
+    identifier: str = typer.Argument(help="Schema or collection identifier in namespace/name format"),
+    version: str | None = typer.Option(
+        None, "--version", "-v", help="Specific schema version, for example 1.2.0. Default: latest published."
+    ),
+    collection: bool | None = typer.Option(
+        None,
```
ogenstad: By default we'll have a `falsy` statement here, i.e. `None` the function signature makes us believe that collection can have three states `True`, `False`, `None` which is not true, the only possible options are `True` or `None` as there doesn't seem to be a way to set it to `False`. As a result we have some dead code below that also acknowledges this. It seems like we compensate for the error here with code that we know will be unreachable below. This should just be a bool that defaults to `False`.

infrahub_sdk/ctl/marketplace.py
```diff
@@ -0,0 +1,331 @@
...
+        if collection is None:
+            item_type, prefetched = await _detect_item_type(client, resolved_url, namespace, name, stdout=stdout)
+            schema_confirmed_exists = item_type == "schema"
+        elif collection:
+            item_type = "collection"
+        else:
+            # Typer exposes only --collection, not --no-collection, so collection=False
+            # is unreachable from the CLI. Kept for defensive completeness against the
+            # bool | None type.
+            item_type = "schema"
```
ogenstad: Refactor once the param above has been changed to a pure `bool` and not a `bool | None`, then reverse the condition so the first option is the `True` one.

infrahub_sdk/ctl/marketplace.py
```diff
@@ -0,0 +1,331 @@
...
+) -> None:
+    """Download all schemas in a collection to disk or stdout.
+
+    When ``prefetched`` is supplied, reuses the response instead of re-fetching
+    the collection download URL.
+    """
+    if prefetched is not None:
+        resp = prefetched
+    else:
+        resp = await client.get(_collection_url(base_url, namespace, name))
```
ogenstad, line 224: In general code is easier to read if we lead with a truthy statement, so:

```python
if prefetched:
    resp = prefetched
else:
    resp = await client.get(_collection_url(base_url, namespace, name))
```

infrahub_sdk/ctl/marketplace.py
ogenstad: I'd also suggest that after the first check for collections we add the type as:

```python
if collection:
    item_type: MarketplaceItemType = "collection"
```

infrahub_sdk/ctl/marketplace.py
```diff
@@ -0,0 +1,331 @@
...
+                    client, resolved_url, namespace, name, output_dir, stdout=stdout, prefetched=prefetched
+                )
+            else:
+                await _download_schema(
+                    client,
+                    resolved_url,
+                    namespace,
+                    name,
+                    version,
+                    output_dir,
```
ogenstad: When we have this many parameters to a function I'd suggest to use keyword arguments for everything so that it's crystal clear regarding what parameter each one is instead on relying on positional parameters.

infrahub_sdk/ctl/marketplace.py
```diff
@@ -0,0 +1,331 @@
...
+            # is unreachable from the CLI. Kept for defensive completeness against the
+            # bool | None type.
+            item_type = "schema"
+
+        try:
+            if item_type == "collection":
+                if version:
+                    _status_console(stdout).print(
+                        "[yellow]Warning: --version is ignored when downloading a collection."
+                    )
```
ogenstad, line 319: I don't really have a strong opinion here, but if a user specifically requests a specific version but ends up on a collection instead I wonder if a better approach would be to fail with and error so that the user could choose if they want to remove the --version param instead or if this is fine. Only raising a thought here.

infrahub_sdk/ctl/marketplace.py
```diff
@@ -0,0 +1,331 @@
...
+        help="Base URL of the Infrahub Marketplace. Overrides configuration and environment.",
+    ),
+    _: str = CONFIG_PARAM,
+) -> None:
+    """Fetch a schema or collection from the Infrahub Marketplace.
+
+    By default, auto-detects whether `namespace/name` is a schema or a collection.
+    Pass --collection to force the collection path when an identifier exists as both.
+    """
+    namespace, name = _parse_identifier(identifier)
```
ogenstad: Here we assume that we have the correct order of the variables returned from _parse_identifier(identifier).

I think it could be a bit cleaner to have _parse_identifier return an object like this:

```python
  @dataclass(frozen=True, slots=True)
  class MarketplaceIdentifier:
      namespace: str
      name: str
```

That way there's no confusion and there are several other functions where we pass in both name and namespace which could be replaced by one variable.

infrahub_sdk/ctl/marketplace.py
```diff
@@ -0,0 +1,331 @@
...
+    the auto-detect probe), so a 404 on a versioned URL is reported as version-not-found
+    rather than the generic not-found message.
+    """
+    if prefetched is not None and version is None:
+        resp = prefetched
+    else:
+        resp = await client.get(_schema_url(base_url, namespace, name, version=version))
+
+    if resp.status_code == 404:
+        if version is not None and schema_confirmed_exists:
```
ogenstad: Assuming we don't want to support an empty string as a version I'd change this to:

```python
if version and schema_confirmed_exists:
```

infrahub_sdk/ctl/marketplace.py
minitriga: Fixed in 69fc5ba — changed to `collection: bool = False` with `is_flag=True`. The `elif collection:` / `else:` branches are now just `if collection: ... else: auto-detect`, removing the unreachable force-schema path entirely.

infrahub_sdk/ctl/marketplace.py
minitriga: Fixed in 69fc5ba — condition is now `if collection: item_type: MarketplaceItemType = "collection"` / `else: auto-detect`.

infrahub_sdk/ctl/marketplace.py
minitriga: Fixed in 69fc5ba — `if prefetched:` now leads.

infrahub_sdk/ctl/marketplace.py
minitriga: Fixed in 69fc5ba — added `item_type: MarketplaceItemType = "collection"` on the collection branch.

infrahub_sdk/ctl/marketplace.py
minitriga: Fixed in 69fc5ba — both `_download_schema` and `_download_collection` calls now use keyword arguments throughout.

infrahub_sdk/ctl/marketplace.py
minitriga: Fixed in 69fc5ba — `_parse_identifier` now returns a `MarketplaceIdentifier(namespace, name)` frozen dataclass, eliminating positional ordering assumptions.

infrahub_sdk/ctl/marketplace.py
minitriga: Fixed in 69fc5ba — changed to `if version and schema_confirmed_exists:`.

infrahub_sdk/config.py
minitriga: The `marketplace_url` field in `ConfigBase` isn't wired into the marketplace command today (which reads from `SETTINGS.active.marketplace_url`). It's a forward-looking field intended to survive once we retire the `infrahubctl.toml` / `Settings` layer — happy to drop it now if you'd prefer to keep things minimal.

infrahub_sdk/ctl/marketplace.py
minitriga: Kept the warn-and-proceed behaviour for now — feels less surprising for a user who runs `get acme/starter-pack --version 1.0.0` and just forgot collections don't have versions. Happy to flip it to a hard error if you'd prefer.

ogenstad (approved):

Issue #923, opened 2026-04-07
feature: Add query_name parameter to all(), filters(), and get() for observability


fatih-acar:
### Component

Python SDK

### Describe the Feature Request

# Feature Request: Add `query_name` parameter to `all()`, `filters()`, and `get()` for observability

## Context

Related to IFC-2395 — Enrich trace data with generator-side context.

Infrahub's backend already collects trace data at the platform level (database queries, API calls, timing) and exposes trace attributes for GraphQL query operation names. However, the SDK query methods (`all()`, `filters()`, `get()`) generate GraphQL queries internally without any user-supplied operation name. This makes it impossible to correlate a trace back to a specific call site in user code (e.g. inside a generator or transform).

## Problem

When a generator or transform issues multiple SDK queries, all of them appear in traces as anonymous GraphQL operations. There is no way to distinguish which `client.all(...)` or `client.get(...)` call produced a given trace span. This forces developers to either:

- Instrument their own code manually with custom timing/logging
- Drop down to `execute_graphql()` with hand-written queries just to control the operation name
- Rely on guesswork to map slow traces back to specific lines of code

## Proposed Solution

Add an optional `query_name: str | None = None` parameter to the `all()`, `filters()`, and `get()` methods on both `InfrahubClient` and `InfrahubClientSync`. When provided, this name should be used as the GraphQL operation name in the generated query, so that it appears in backend traces.

### Example Usage

```python
# Today — anonymous query, no way to identify in traces
devices = await client.all(kind="InfraDevice")

# Proposed — named query, shows up as "GetAllEdgeDevices" in trace spans
devices = await client.all(kind="InfraDevice", query_name="GetAllEdgeDevices")
```

```python
# In a generator, giving each query a meaningful name for trace correlation
site = await client.get(
    kind="LocationSite",
    name__value=site_name,
    query_name="SiteGenerator_FetchTargetSite",
)

vlans = await client.filters(
    kind="InfraVLAN",
    site__name__value=site_name,
    query_name="SiteGenerator_FetchSiteVLANs",
)
```

### Expected Behavior

1. When `query_name` is provided, the SDK should use it as the GraphQL operation name in the generated query string (e.g. `query SiteGenerator_FetchTargetSite { ... }` instead of `query { ... }`).
2. When `query_name` is `None` (default), behavior is unchanged — no breaking change.
[...]

PR #965, opened 2026-04-23
IHS-219: Add query_name parameter to all(), filters(), and get() for observability

solababs:
# Why

When the SDK issues GraphQL queries internally, they are anonymous (`query { ... }`), making it impossible to correlate a slow backend trace span back to a specific call site. Developers debugging performance issues have no way to tell which `all()` or `filters()` call produced a given trace.

Goal: give every SDK-generated query a meaningful GraphQL operation name so trace spans are identifiable without custom instrumentation.

Closes https://github.com/opsmill/infrahub-sdk-python/issues/923

## What changed

- `all()`, `filters()`, `get()`, and `count()` on both `InfrahubClient` and `InfrahubClientSync` now accept an optional `query_name: str | None = None` parameter.
- When `query_name` is supplied it becomes the GraphQL operation name (e.g. `query GetAllEdgeDevices { ... }`).
- When omitted, a default name is auto-generated from the method and kind: `All_CoreRepository`, `Filters_CoreRepository`, `Get_CoreRepository`, `Count_CoreRepository`. This means **all queries now have an operation name** — previously they were anonymous.

## How to test

```bash
uv run pytest tests/unit/sdk/test_client.py -k "query_name" -v
```

## Impact & rollout

- **Backward compatibility:** fully backward-compatible. `query_name` defaults to an auto-generated value, so existing call sites need no changes.
- **Performance:** no measurable impact — the name is a short string prepended to the query.
- **Deployment notes:** safe to deploy independently.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry added (`uv run towncrier create ...`)

### DIFF
--- a/infrahub_sdk/client.py
+++ b/infrahub_sdk/client.py
@@ -378,6 +378,7 @@ class InfrahubClient(BaseClient):
         prefetch_relationships: bool = ...,
         property: bool = ...,
         include_metadata: bool = ...,
+        query_name: str | None = ...,
         **kwargs: Any,
     ) -> SchemaType | None: ...
 
@@ -398,6 +399,7 @@ class InfrahubClient(BaseClient):
         prefetch_relationships: bool = ...,
         property: bool = ...,
         include_metadata: bool = ...,
+        query_name: str | None = ...,
         **kwargs: Any,
     ) -> SchemaType: ...
 
@@ -418,6 +420,7 @@ class InfrahubClient(BaseClient):
         prefetch_relationships: bool = ...,
         property: bool = ...,
         include_metadata: bool = ...,
+        query_name: str | None = ...,
         **kwargs: Any,
     ) -> SchemaType: ...
 
@@ -438,6 +441,7 @@ class InfrahubClient(BaseClient):
         prefetch_relationships: bool = ...,
         property: bool = ...,
         include_metadata: bool = ...,
+        query_name: str | None = ...,
         **kwargs: Any,
     ) -> InfrahubNode | None: ...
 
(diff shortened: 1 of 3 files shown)


infrahub_sdk/client.py
```diff
@@ -692,10 +717,13 @@ async def all(
             parallel (bool, optional): Whether to use parallel processing for the query.
             order (Order, optional): Ordering related options. Setting `disable=True` enhances performances.
             include_metadata (bool, optional): If True, includes node_metadata and relationship_metadata in the query.
+            query_name (str, optional): If provided is used a the GraphQL operation name else All_<kind> is used.
```
ajtmccarty: `is used a` -> `is used as`

infrahub_sdk/client.py
```diff
@@ -796,13 +828,16 @@ async def filters(
             parallel (bool, optional): Whether to use parallel processing for the query.
             order (Order, optional): Ordering related options. Setting `disable=True` enhances performances.
             include_metadata (bool, optional): If True, includes node_metadata and relationship_metadata in the query.
+            query_name (str, optional): If provided is used a the GraphQL operation name else Filters_<kind> is used.
```
ajtmccarty: same

infrahub_sdk/client.py
```diff
@@ -2111,10 +2152,13 @@ def all(
             parallel (bool, optional): Whether to use parallel processing for the query.
             order (Order, optional): Ordering related options. Setting `disable=True` enhances performances.
             include_metadata (bool, optional): If True, includes node_metadata and relationship_metadata in the query.
+            query_name (str, optional): If provided is used a the GraphQL operation name else All_<kind> is used.
```
ajtmccarty: same

ajtmccarty (approved):

PR #972, opened 2026-04-27
Add missing external identities protocol

solababs:
Add missing InternalExternalIdentity node to protocols


polmichel (approved):

ogenstad:
Hm, actually there are for things for the Internal namespace, we might have another bug in the generation code for this since I see that we also have the tokens relationship on CoreAccount and tokens is also in the `Internal` namespace, so I think the correct thing here would actually be to ensure that we don't render anything related to the `Internal` repo in the SDK.. Could be that we should have another issue for that, it should be safe to merge this to avoid having the generation spit out things over and over if we don't add them, but at some point it would be better to remove those entries.

PR #976, opened 2026-04-28
Add `infrahubctl graphql query-report` command

ogenstad:
# Why

Adds a new subcommand to infrahubctl for `infrahubctl graphql query-report`, it's to allow users to check a query either from a .infrahub.yml file or by specifying the name of a query that exists in Infrahub and returning a query report. For now the only entry in the report is that users can see if the query can be used for unique targets for artifact definitions.

### Example output

```bash
blue on  main [⇡] via 🐍 v3.12.12 (infrahub-sdk) on ☁️   took 2s
❯ /Users/patrick/Code/opsmill/infrahub/.venv/bin/infrahubctl graphql query-report gen
Query 'gen' (local: generators/gen.gql)
Targets unique nodes: false — all artifacts for the definition will be regenerated on any relevant node change.

blue on  main [⇡] via 🐍 v3.12.12 (infrahub-sdk) on ☁️
❯ /Users/patrick/Code/opsmill/infrahub/.venv/bin/infrahubctl graphql query-report character
Query 'character' (local: checks/character.gql)
Targets unique nodes: true — Infrahub will limit artifact regeneration to changed nodes only.
```

## What changed

* New command for infrahubctl that uses GraphQL query introduced in Infrahub 1.9.1
* Tests

### DIFF
--- a/infrahub_sdk/ctl/graphql.py
+++ b/infrahub_sdk/ctl/graphql.py
@@ -125,7 +125,6 @@ async def query_report(
     _: str = CONFIG_PARAM,
 ) -> None:
     """Run a GraphQL query through InfrahubGraphQLQueryReport and report its analysis."""
-
     client = initialize_client(branch=branch)
 
     if online:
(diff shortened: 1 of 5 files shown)


infrahub_sdk/ctl/graphql.py
```diff
@@ -100,6 +103,76 @@ def callback() -> None:
...
+    branch: str | None = typer.Option(None, help="Branch on which to run the report."),
+    _: str = CONFIG_PARAM,
+) -> None:
+    """Run a GraphQL query through InfrahubGraphQLQueryReport and report its analysis."""
+
+    client = initialize_client(branch=branch)
+
+    if online:
+        node = await client.get(
+            kind=CoreGraphQLQuery,  # type: ignore[type-abstract]
```
ogenstad, line 132: Temporary ignore until we can resolve the protocol issues within the SDK.

gmazoyer (approved):

PR #977, opened 2026-04-28
Upgrade ruff==0.15.12

ogenstad:
# Why

Keep linting tools up to date

## What changed

* Update ruff in pyproject and git precommit config
* Change code in asserts as per https://docs.astral.sh/ruff/rules/float-equality-comparison/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.15.0
+    rev: v0.15.12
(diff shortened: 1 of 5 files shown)


polmichel (approved):
LGTM! I assume another PR will target the `infrahub-sdk-python` project?

ogenstad:
> LGTM! I assume another PR will target the `infrahub-sdk-python` project?

I assume you meant `infrahub` but yes that will come. It's not really urgent though and might be nice to have it updated in the SDK first as without it when we do a `ruff check .` in the infrahub project we could see violations from the SDK if the SDK wasn't up to date first.

PR #986, opened 2026-04-30
Update infrahub_testcontainers-1.9.1

ogenstad:
# Why

Update infrahub-testcontainers to 1.9.1, because I want to have it in place for #976

## What changed

* infrahub-testcontainers dependency in lockfile

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -878,7 +878,7 @@ types = [
 
 [[package]]
 name = "infrahub-testcontainers"
-version = "1.9.0"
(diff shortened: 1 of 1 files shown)


gmazoyer (approved):
I think you can also remove XFAIL markers in `tests/integration/test_file_object.py`

ogenstad:
> I think you can also remove XFAIL markers in `tests/integration/test_file_object.py`

Ah nice. I'll try that as a follow up!

PR #997, opened 2026-05-07
Bump ariadne-codegen version

solababs:
Bump ariadne-codegen version

### DIFF
--- a/docs/docs/infrahubctl/infrahubctl-branch.mdx
+++ b/docs/docs/infrahubctl/infrahubctl-branch.mdx
@@ -18,13 +18,28 @@ $ infrahubctl branch [OPTIONS] COMMAND [ARGS]...
 
 **Commands**:
 
+* `list`: List all existing branches.
 * `create`: Create a new branch.
 * `delete`: Delete a branch.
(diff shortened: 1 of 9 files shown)


ogenstad (commented):
Good find! Is this a bug that should be reported to [`ariadne-codegen`](https://github.com/mirumee/ariadne-codegen)? I'm wondering if it might be fixed in a later version since the one we've pinned is a bit old.

Could be worth to upgrade to see if it resolves the issue, other than that based on the added code can we have some test that would fail if we removed this inject_typename?

ogenstad:
Nice, so I assume that the upgrade fixed the problem? I'm just wondering about those doc updates if something is wrong in the repo or what happened here. I assume that the pipeline failed at first without you touching those files?

solababs:
> Nice, so I assume that the upgrade fixed the problem? I'm just wondering about those doc updates if something is wrong in the repo or what happened here. I assume that the pipeline failed at first without you touching those files?

Yes the new typer version reorders some of the commands in chronological order

ogenstad (approved):
Ok, got it thanks!

PR #998, opened 2026-05-07
chore(deps): patch transitive vulns in docs and bump requests dev floor

petercrocker:
## Summary

- Adds an `overrides` block to `docs/package.json` pinning patched versions of 12 vulnerable transitive deps pulled in by Docusaurus, and regenerates `docs/package-lock.json` against the current npm registry.
- Bumps the `requests` dev-dependency floor to `>=2.33.0` to clear GHSA-gc5v-m9x4-r6x2 (insecure temp file reuse in `extract_zipped_paths`).

## Why now (vs. waiting for Docusaurus)

26 of the 27 open Dependabot alerts are in `docs/package-lock.json` — all transitive deps of Docusaurus, none declared directly in `docs/package.json`. `@docusaurus/core` is already on the latest 3.x line (`^3.10.0` → resolves to 3.10.1), so there's no upstream release to wait on; the patches are available today and the lockfile just hadn't been refreshed.

Scoped `name@range` overrides are used for `path-to-regexp`, `picomatch`, and `ajv` so that only the vulnerable ranges are bumped — parallel major versions in the dep tree are left untouched.

## Verification

- `npm audit` in `docs/`: **found 0 vulnerabilities**
- `uv lock --check`: passes
- Pinned versions in regenerated lockfile (spot-checked): postcss 8.5.14, follow-redirects 1.16.0, lodash 4.18.1, path-to-regexp 0.1.13 (vulnerable instance only; 1.9.0 and 3.3.0 left intact), serialize-javascript 7.0.5, picomatch 2.3.2 + 4.0.4, markdown-it 14.1.1, ajv 6.14.0 + 8.18.0, qs 6.15.1, webpack 5.106.2.

## Test plan

- [ ] CI green (lint + tests)
- [ ] `cd docs && npm install && npm run build` succeeds locally
- [ ] Dependabot alerts auto-close after merge to `stable`

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -32,15 +32,15 @@
       }
     },
     "node_modules/@algolia/abtesting": {
-      "version": "1.16.1",
-      "resolved": "https://registry.npmjs.org/@algolia/abtesting/-/abtesting-1.16.1.tgz",
-      "integrity": "sha512-Xxk4l00pYI+jE0PNw8y0MvsQWh5278WRtZQav8/BMMi3HKi2xmeuqe11WJ3y8/6nuBHdv39w76OpJb09TMfAVQ==",
+      "version": "1.18.1",
+      "resolved": "https://registry.npmjs.org/@algolia/abtesting/-/abtesting-1.18.1.tgz",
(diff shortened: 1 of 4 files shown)


ajtmccarty (approved):
backend bump looks good. don't know about the frontend, requesting their review

PR #1002, opened 2026-05-08
fix(batch): cancel in-flight tasks when InfrahubBatch.execute() exits

### LINKED ISSUE #1001: bug: InfrahubBatch.execute() orphans in-flight tasks when re-raising
### Component

Python SDK

### Infrahub SDK version

1.20.0 (`infrahub_sdk/batch.py` on `develop`)

### Current Behavior

`InfrahubBatch.execute()` creates one `asyncio.Task` per `BatchTask` and iterates with `asyncio.as_completed()`. When `return_exceptions=False` (the default) and any task raises, the generator re-raises immediately. The remaining N − 1 tasks are still running on the event loop; nothing awaits or cancels them.

```python
async def execute(self) -> AsyncGenerator:
    tasks = [
        asyncio.create_task(execute_batch_task_in_pool(...))
        for batch_task in self._tasks
    ]

    for completed_task in asyncio.as_completed(tasks):
        node, result = await completed_task
        if isinstance(result, Exception) and not self.return_exceptions:
            raise result   # siblings still running, no cleanup
        yield node, result
```

When the orphaned tasks eventually finish:

- **Successful:** their work was performed (e.g., `node.save()` did write to the server) but no caller observes it. The caller has been told the batch failed and won't retry.
- **Failed:** asyncio's task GC emits `Task exception was never retrieved` on stderr. There is no other surface for the caller to discover them.

This surfaces under transient server slowness. Two identical batch-driven builds against separate Infrahub deployments can produce divergent state — one complete, the other missing a portion of intended writes plus all downstream objects that depended on those writes — with build exit status 0 on both sides. Logs from such a run show proportional counts of `ServerNotResponsiveError` and `Task exception was never retrieved` in the affected window, suggesting both error families are two views of the same underlying event.

### Expected Behavior

After `execute()` returns or raises, no `asyncio.Task` it created is still running. Specifically:

- When a task raises and `return_exceptions=False`, sibling tasks should be cancelled (or otherwise drained) before the exception propagates to the caller.
- This must hold for all generator exit paths: explicit raise, caller `break`, generator close.

### Steps to Reproduce

```python
import asyncio
import pytest
from infrahub_sdk.batch import InfrahubBatch

async def test_orphan() -> None:
    side_effects: list[str] = []

    async def raise_fast() -> None:
        await asyncio.sleep(0)
        raise RuntimeError("fast failure")

    async def slow_side_effect(name: str) -> str:
        await asyncio.sleep(0.1)
        side_effects.append(name)
        return name

    batch = InfrahubBatch(max_concurrent_execution=10, return_exceptions=False)
    batch.add(task=raise_fast)
    for i in range(5):
        batch.add(task=slow_side_effect, name=f"slow-{i}")

    with pytest.raises(RuntimeError):
        async for _ in batch.execute():
            pass

    await asyncio.sleep(0.3)
    # Currently fails: side_effects == ['slow-0', 'slow-1', 'slow-2', 'slow-3', 'slow-4']
    assert side_effects == []
```

### Additional Information

Related but separate defect: with `return_exceptions=True`, failures are yielded as `(node, ExceptionInstance)` in the same tuple shape as successes. A naive consumer that records `node` per yield silently treats failed tasks as successful. Will file separately — that one needs a small API redesign rather than a one-line fix.

PhillSimonds:
# Why

`InfrahubBatch.execute()` orphans in-flight tasks when one task raises with `return_exceptions=False`. Sibling work continues silently on the event loop; the caller has been told the batch failed and cannot retry; failed siblings surface only as `Task exception was never retrieved` GC warnings. Under a transient server-slowness window, this can yield build exit status `0` against a partial graph.

Closes #1001

## What changed

- `infrahub_sdk/batch.py`: wrapped the `asyncio.as_completed` iteration in `try/finally` that cancels any still-running task and `gather(..., return_exceptions=True)` to drain. Covers all generator exit paths (raise, caller `break`, generator close), not only the original raise path.
- Tests:
  - `test_execute_does_not_orphan_inflight_tasks_when_raising` — regression guard for the orphan behavior. Asserts that sibling task side effects do not run to completion after `execute()` raises.
  - `test_return_exceptions_yields_exceptions_indistinguishably_from_successes` — pins the current `return_exceptions=True` contract (failures yield as `(node, ExceptionInstance)` in the same tuple shape as successes). Expected to change when that API is reworked; tracked separately.

## How to review

Start with `infrahub_sdk/batch.py`. The diff is small and self-contained. The two new tests in `tests/unit/sdk/test_batch.py` run without HTTPX mocks (~0.4s total).

## How to test

```bash
uv run pytest tests/unit/sdk/test_batch.py -v
```

Expected: 8 passed.

## Impact & rollout

- **Backward compatibility:** behavior change only on the failure path. Healthy batches behave identically. Callers that previously relied on orphaned siblings continuing in the background after their batch raised will no longer get that — that reliance was undefined behavior (and silent data loss).
- **Performance:** unchanged.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added (`changelog/1001.fixed.md`)
- [x] External docs updated (n/a — internal behavior fix)
- [x] Internal .md docs updated (n/a)

### DIFF
--- a/infrahub_sdk/batch.py
+++ b/infrahub_sdk/batch.py
@@ -80,11 +80,20 @@ class InfrahubBatch:
             for batch_task in self._tasks
         ]
 
-        for completed_task in asyncio.as_completed(tasks):
-            node, result = await completed_task
-            if isinstance(result, Exception) and not self.return_exceptions:
-                raise result
-            yield node, result
+        try:
+            for completed_task in asyncio.as_completed(tasks):
+                node, result = await completed_task
+                if isinstance(result, Exception) and not self.return_exceptions:
+                    raise result
+                yield node, result
+        finally:
+            # Ensure no task created here outlives execute(). Cancel any still
+            # running, then drain so their exceptions are retrieved instead of
+            # surfacing later as "Task exception was never retrieved".
+            for t in tasks:
+                if not t.done():
+                    t.cancel()
+            await asyncio.gather(*tasks, return_exceptions=True)
 
 
 class InfrahubBatchSync:
--- a/tests/unit/sdk/test_batch.py
+++ b/tests/unit/sdk/test_batch.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
+import asyncio
 from typing import TYPE_CHECKING
 
 import pytest
 
+from infrahub_sdk.batch import InfrahubBatch
 from infrahub_sdk.exceptions import GraphQLError
 
 if TYPE_CHECKING:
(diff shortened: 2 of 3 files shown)


infrahub_sdk/batch.py
```diff
@@ -80,11 +80,20 @@ async def execute(self) -> AsyncGenerator:
...
+        try:
+            for completed_task in asyncio.as_completed(tasks):
+                node, result = await completed_task
+                if isinstance(result, Exception) and not self.return_exceptions:
+                    raise result
+                yield node, result
+        finally:
+            # Ensure no task created here outlives execute(). Cancel any still
+            # running, then drain so their exceptions are retrieved instead of
+            # surfacing later as "Task exception was never retrieved".
```
ajtmccarty, line 92: I'd say that you get, at most, 2 lines for a comment
`When raising an error, all tasks not done are canceled and/or drained`

tests/unit/sdk/test_batch.py
```diff
@@ -125,3 +127,91 @@ async def test_batch_exception(
...
+
+
+async def test_return_exceptions_yields_exceptions_indistinguishably_from_successes() -> None:
+    """Pins down the current contract of ``execute()`` with ``return_exceptions=True``.
+
+    Failures are yielded as ``(node, ExceptionInstance)`` using the same tuple
+    shape as successes ``(node, result)``. The yielded ``node`` is whatever the
+    caller passed via ``batch.add(..., node=...)`` regardless of outcome, so a
+    consumer that does not ``isinstance``-check ``result`` cannot tell a failed
+    task from a successful one and will silently treat both as "created".
```
ajtmccarty, line 175: I think this is just the expected behavior when `return_exceptions=True`

ajtmccarty (commented):
The functional code change looks good. comments on a the code comment and the second test

PR #1008, opened 2026-05-10
fix(ctl): handle extensions paths in display_schema_load_errors (#1007)

### LINKED ISSUE #1007: bug: `infrahubctl schema load` - `display_schema_load_errors` crashes with `ValueError` when error path traverses an `extensions:` block
### Component

Python SDK, infrahubctl

### Infrahub SDK version

1.20.0

### Current Behavior

`infrahubctl schema load` calls `infrahub_sdk.ctl.schema.display_schema_load_errors` to render server-rejected schema validation errors. The function indexes `loc_path[2]` and `loc_path[4]` and casts both to `int`, assuming the loc tuple shape is `("body", "schemas", <schema_index>, "nodes"|"generics", <node_index>, ...)`.

When the rejected error originates inside an `extensions:` block (per [docs.infrahub.app/topics/schema-extensions](https://docs.infrahub.app/topics/schema-extensions)), the loc tuple has an extra segment:

```
("body", "schemas", <schema_index>, "extensions", "nodes"|"generics"|"relationships", <node_index>, <field>)
```

So `loc_path[4]` is a string like `"generics"` instead of the integer node index. `int("generics")` raises:

```
ValueError: invalid literal for int() with base 10: 'generics'
```

The traceback replaces the intended human-readable error message — the user never sees what the server actually rejected.

### Expected Behavior

The server's rejection should be rendered in the same human-readable format as non-extensions errors, e.g.:

```
Unable to load the schema:
  Node: DcimGenericDevice (extensions/generics) | Extra inputs are not permitted (extra_forbidden) — field 'include_in_menu' is not part of the extension schema
```

### Environment

- Infrahub server: 1.9.2 (any 1.5+ exhibits the rejection that triggers the formatter)
- `infrahub-sdk[all]` 1.20.0
- Python 3.12

### Steps

**1. Have a running Infrahub instance** with a schema that includes a generic that has `include_in_menu: true` (e.g., the `DcimGenericDevice` from `opsmill/schema-library`'s `base/dcim.yml`). Any kind whose existing schema metadata you want to override works.

**2. Author a schema-extension YAML** that attempts to override a metadata field on an existing kind:

```yaml
# repro_extension.yml
---
version: "1.0"
extensions:
  generics:
    - kind: DcimGenericDevice
      include_in_menu: false
```

**3. Run `infrahubctl schema load`:**

```bash
infrahubctl schema load repro_extension.yml
```

### Actual output (1.20.0)

```
Schema loading.....
Unable to load the schema:
  File ".../site-packages/infrahub_sdk/ctl/schema.py", line 87, in load
    display_schema_load_errors(response=response.errors, schemas_data=schemas_data)
  File ".../site-packages/infrahub_sdk/ctl/schema.py", line 68, in display_schema_load_errors
    node_index = int(loc_path[4])
                 ^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'generics'
```

The user sees no human-readable explanation of why the schema was rejected.

## Affected code

`infrahub_sdk/ctl/schema.py` lines 53–103 (in 1.20.0), function `display_schema_load_errors`:

```python
def display_schema_load_errors(response: dict[str, Any], schemas_data: list[SchemaFile]) -> None:
    console.print("[red]Unable to load the schema:")
    if "detail" not in response:
        handle_non_detail_errors(response=response)
        return

    for error in response["detail"]:
        loc_path = error.get("loc", [])
        if not valid_error_path(loc_path=loc_path):
            continue

        # if the len of the path is equal to 6, the error is at the root of the object
        # if the len of the path is higher than 6, the error is in an attribute or a relationships
        schema_index = int(loc_path[2])
        node_index = int(loc_path[4])           # <-- crashes here for extensions paths
        ...
```

And the path validator at line 116, which lets the malformed path through:

```python
def valid_error_path(loc_path: list[Any]) -> bool:
    return len(loc_path) >= 6 and loc_path[0] == "body" and loc_path[1] == "schemas"
```

`valid_error_path` does not check that `loc_path[3]` is `"nodes"` or `"generics"`, so `loc_path[3] == "extensions"` slips through and the indexing assumption breaks.

## Suggested fix

Branch on whether `loc_path[3] == "extensions"` and re-anchor the indexing:

```python
def display_schema_load_errors(response: dict[str, Any], schemas_data: list[SchemaFile]) -> None:
    console.print("[red]Unable to load the schema:")
    if "detail" not in response:
        handle_non_detail_errors(response=response)
        return

    for error in response["detail"]:
        loc_path = error.get("loc", [])
        if not valid_error_path(loc_path=loc_path):
            continue

        schema_index = int(loc_path[2])

        # Distinguish top-level vs extensions-block error paths.
        # Top-level:  body/schemas/<si>/nodes|generics/<ni>/<field>
        # Extensions: body/schemas/<si>/extensions/nodes|generics|relationships/<ni>/<field>
        if loc_path[3] == "extensions":
            container = loc_path[4]                 # 'nodes' / 'generics' / 'relationships'
            node_index = int(loc_path[5])
            tail = loc_path[6:]
        else:
            container = loc_path[3]                 # 'nodes' / 'generics'
            node_index = int(loc_path[4])
            tail = loc_path[5:]
        ...
```

`get_node` and the downstream `len(loc_path) == 6` / `> 6` branches need corresponding adjustments (the absolute lengths shift by one for the extensions path).

`valid_error_path` should additionally accept `loc_path[3] in ("nodes", "generics", "extensions")` so unrelated server errors don't fall through this codepath.

iddocohen:
# Why

  `infrahubctl schema load` crashes with an unhandled `ValueError` when the
  server rejects a schema that uses the `extensions:` block. The traceback replaces the
  human-readable rejection message, so users have no idea what the server
  actually rejected.

  **Goal:** Render extensions-block rejections in the same one-line format as
  top-level rejections.

  **Non-goals:** No new error categories, no changes to `infrahubctl schema
  check` or to the server's rejection format, no broader refactor of the CLI
  schema commands.

  Closes #1007

  ## What changed

  **Behavioral changes**

  - Schema rejections originating inside an `extensions:` block now render as
    `Node: <kind> (extensions/<container>) | <field> (<input>) | <msg> (<type>)`
    instead of crashing with `ValueError: invalid literal for int() with base 10`.
  - Unknown / malformed `loc_path` shapes (non-int schema index, unrecognised
    container) are skipped silently instead of crashing the renderer.
  - Top-level (non-extension) error rendering is byte-for-byte unchanged —
    preserved by the existing tests.

  **Implementation notes**

  - `display_schema_load_errors` now branches on `loc_path[3] == "extensions"`,
    computes `container / node_index / tail` once, and dispatches on `len(tail)`
    instead of `len(loc_path)` so the offset shift doesn't have to be repeated.
  - Per-error rendering extracted into `_render_schema_error` to keep the main
    loop within the project's branch-count limits.
  - `valid_error_path` now whitelists `loc_path[3] in {nodes, generics,
    extensions}` and requires `loc_path[2]` (and the node-index slot) to be `int`.
  - `get_node` gained `container` and `is_extension` kwargs (defaults match the
    old behaviour) and uses `payload.get(...)` so a schema file with only
    `extensions:` no longer raises `KeyError`.

  **What stayed the same**

  - API contract of `display_schema_load_errors` unchanged.
  - `get_node` signature is backward-compatible (new params have defaults).
  - Output format for non-extension errors is unchanged.

  ## How to review

  - `infrahub_sdk/ctl/schema.py` — start with `_render_schema_error`, then
    `valid_error_path`, then `get_node`.
  - `tests/unit/sdk/test_schema.py` — three new tests cover top-level extensions
    errors, nested-attribute extensions errors, and the malformed-path filter.
  - The pre-existing tests for non-extension paths are intentionally untouched
    and still pass — that's the regression guard for the format-stability claim
    above.

  The `valid_error_path` whitelist is the area I'd most want a second opinion
  on: if the server later adds a new top-level container, errors against it
  will now be silently skipped instead of bubbling up. The trade-off is
  intentional for a CLI rendering helper, but worth flagging.

  ## How to test

  ```bash
  # Repro from the issue (requires a running Infrahub 1.5+ with DcimGenericDevice
  # already loaded, e.g. opsmill/schema-library base/dcim.yml):
  cat > /tmp/repro_extension.yml <<'YAML'
  ---
  version: "1.0"
  extensions:
    generics:
      - kind: DcimGenericDevice
        include_in_menu: false
  YAML
  infrahubctl schema load /tmp/repro_extension.yml
  # Before: ValueError traceback. After: one-line "Node: DcimGenericDevice (extensions/generics) | ..." rendering.

  # Unit tests:
  uv run pytest tests/unit/sdk/test_schema.py -k display_schema_load_errors -v

  # Lint:
  ```

  ## Impact & rollout

  - **Backward compatibility:** No breaking changes. CLI output for non-extension
    errors is unchanged. `get_node` gained two optional kwargs with defaults.
  - **Performance:** N/A — only touched on the error path of `schema load`.
  - **Config/env changes:** None.
  - **Deployment notes:** Safe to deploy. Pure SDK/CLI fix, no server-side change.

  ## Checklist

  - [x] Tests added/updated
  - [ ] Changelog entry added (`uv run towncrier create +fix-schema-load-extensions.fixed.md`)

### DIFF
--- a/infrahub_sdk/ctl/schema.py
+++ b/infrahub_sdk/ctl/schema.py
@@ -101,15 +101,17 @@ def _render_schema_error(
     )
     path_suffix = f" (extensions/{container})" if is_extension else ""
     input_str = error.get("input")
+    err_msg = error.get("msg", "No error message")
+    err_type = error.get("type", "unknown")
 
     if len(tail) == 1:
         loc_type = tail[0]
-        error_message = f"{loc_type} ({input_str}) | {error['msg']} ({error['type']})"
+        error_message = f"{loc_type} ({input_str}) | {err_msg} ({err_type})"
     elif len(tail) > 1:
         loc_type = tail[0]
         attribute = tail[1]
         input_label = _resolve_attribute_label(error_data=node.get(loc_type, []), attribute=attribute)
-        error_message = f"{loc_type[:-1].title()}: {input_label} ({input_str}) | {error['msg']} ({error['type']})"
+        error_message = f"{loc_type[:-1].title()}: {input_label} ({input_str}) | {err_msg} ({err_type})"
     else:
         return
 
--- a/infrahub_sdk/ctl/schema.py
+++ b/infrahub_sdk/ctl/schema.py
@@ -4,7 +4,7 @@ import asyncio
 import time
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 
 import typer
 import yaml
@@ -148,11 +148,14 @@ def valid_error_path(loc_path: list[Any]) -> bool:
     return loc_path[3] in {"nodes", "generics"} and isinstance(loc_path[4], int)
 
 
+SchemaContainer = Literal["nodes", "generics", "relationships"]
+
+
 def get_node(
     schemas_data: list[SchemaFile],
     schema_index: int,
     node_index: int,
-    container: str = "nodes",
+    container: SchemaContainer = "nodes",
     is_extension: bool = False,
 ) -> dict | None:
     if schema_index >= len(schemas_data):
--- a/infrahub_sdk/ctl/schema.py
+++ b/infrahub_sdk/ctl/schema.py
@@ -94,6 +94,7 @@ def _render_schema_error(
         output.print("Node data not found.")
         return
 
+    # Extensions reference an existing node by `kind`; new top-level nodes are identified by `namespace+name`.
     node_label = (
         (node.get("kind") or node.get("name") or "")
         if is_extension
@@ -105,12 +106,16 @@ def _render_schema_error(
     err_type = error.get("type", "unknown")
 
     if len(tail) == 1:
+        # Error on a direct field of the node (e.g. `name`, `namespace`).
         loc_type = tail[0]
         error_message = f"{loc_type} ({input_str}) | {err_msg} ({err_type})"
     elif len(tail) > 1:
+        # Error nested inside a collection (e.g. attributes[2].kind, relationships[0].peer).
+        # loc_type is the collection name; attribute is either its index or the failing field name.
         loc_type = tail[0]
         attribute = tail[1]
         input_label = _resolve_attribute_label(error_data=node.get(loc_type, []), attribute=attribute)
+        # Trim the trailing 's' so "attributes" → "Attribute" in the rendered label.
         error_message = f"{loc_type[:-1].title()}: {input_label} ({input_str}) | {err_msg} ({err_type})"
     else:
         return
@@ -124,7 +129,9 @@ def _resolve_attribute_label(error_data: list[dict[str, Any]], attribute: Any) -
             if data.get(attribute) is not None:
                 return data.get("name", None)
         return None
-    return error_data[attribute].get("name", None)
+    if isinstance(attribute, int) and 0 <= attribute < len(error_data):
+        return error_data[attribute].get("name", None)
+    return None
 
 
 def handle_non_detail_errors(response: dict[str, Any], output: Console | None = None) -> None:
--- a/infrahub_sdk/ctl/schema.py
+++ b/infrahub_sdk/ctl/schema.py
@@ -59,44 +59,65 @@ def display_schema_load_errors(response: dict[str, Any], schemas_data: list[Sche
         loc_path = error.get("loc", [])
         if not valid_error_path(loc_path=loc_path):
             continue
+        _render_schema_error(error=error, loc_path=loc_path, schemas_data=schemas_data)
 
-        # if the len of the path is equal to 6, the error is at the root of the object
-        # if the len of the path is higher than 6, the error is in an attribute or a relationships
-        schema_index = int(loc_path[2])
+
+def _render_schema_error(error: dict[str, Any], loc_path: list[Any], schemas_data: list[SchemaFile]) -> None:
+    # Two layout shapes for loc_path. tail is the part after the node index.
+    # Top-level: body / schemas / <si> / (nodes|generics) / <ni> / [<subtype> / <attr>]
+    # Extensions: body / schemas / <si> / extensions / (nodes|generics|relationships) / <ni> / [<subtype> / <attr>]
+    schema_index = int(loc_path[2])
+    is_extension = loc_path[3] == "extensions"
+    if is_extension:
+        container = loc_path[4]
+        node_index = int(loc_path[5])
+        tail = loc_path[6:]
+    else:
+        container = loc_path[3]
         node_index = int(loc_path[4])
-        node = get_node(schemas_data=schemas_data, schema_index=schema_index, node_index=node_index)
+        tail = loc_path[5:]
 
-        if not node:
-            console.print("Node data not found.")
-            continue
+    node = get_node(
+        schemas_data=schemas_data,
+        schema_index=schema_index,
+        node_index=node_index,
+        container=container,
+        is_extension=is_extension,
+    )
 
-        if len(loc_path) == 6:
-            loc_type = loc_path[-1]
-            input_str = error.get("input", None)
-            error_message = f"{loc_type} ({input_str}) | {error['msg']} ({error['type']})"
-            console.print(
-                f"  Node: {node.get('namespace', None)}{node.get('name', None)} | {error_message}", markup=False
-            )
+    if not node:
+        console.print("Node data not found.")
+        return
 
-        elif len(loc_path) > 6:
-            loc_type = loc_path[5]
-            error_data = node[loc_type]
-            attribute = loc_path[6]
+    node_label = (
+        (node.get("kind") or node.get("name") or "")
+        if is_extension
+        else f"{node.get('namespace', None)}{node.get('name', None)}"
+    )
+    path_suffix = f" (extensions/{container})" if is_extension else ""
+    input_str = error.get("input")
 
-            if isinstance(attribute, str):
-                input_label = None
-                for data in error_data:
-                    if data.get(attribute) is not None:
-                        input_label = data.get("name", None)
-                        break
-            else:
-                input_label = error_data[attribute].get("name", None)
+    if len(tail) == 1:
+        loc_type = tail[0]
+        error_message = f"{loc_type} ({input_str}) | {error['msg']} ({error['type']})"
+    elif len(tail) > 1:
+        loc_type = tail[0]
+        attribute = tail[1]
+        input_label = _resolve_attribute_label(error_data=node.get(loc_type, []), attribute=attribute)
+        error_message = f"{loc_type[:-1].title()}: {input_label} ({input_str}) | {error['msg']} ({error['type']})"
+    else:
+        return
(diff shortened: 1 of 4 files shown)


ogenstad:
Not a review but just highlighting that we also want a towncrier newsfragment under /changelog so that we include what was fixed in the release notes.

Also in general when we're fixing bugs it's fine to target the `stable` release directly. That part is more important in the Infrahub repo than it is in the SDK though as in the SDK we can merge `develop` directly into `stable` for a patch release but that would never happen within the main Infrahub project.

ajtmccarty:
I think you'll need to rebase this on `stable` to remove the extra commits currently included

tests/unit/sdk/test_schema.py
ajtmccarty, line 1: this would be easier to test if this code was written using dependency injection and then you could skip the `patch()` calls. sometimes `patch`ing is necessary for testing, but we generally try to avoid it when you can achieve a more robust test by moving the test up a layer or two in the test pyramid. in this case testing just `display_schema_load_errors` is fast, but brittle b/c it assumes a certain response structure from the schema load endpoint. I think that these would be better as integration tests similar to those in `python_sdk/tests/integration/test_schema.py`, then the test really ensures that the response from the server and the client handling work together correctly. sadly, integration tests are slower, but I think it is worth it in this case b/c we have clearly left schema extensions behind for a while and don't want to again

infrahub_sdk/ctl/schema.py
```diff
@@ -110,12 +131,30 @@ def handle_non_detail_errors(response: dict[str, Any]) -> None:
...
+            and isinstance(loc_path[5], int)
+        )
+    return loc_path[3] in {"nodes", "generics"} and isinstance(loc_path[4], int)
+
+
+def get_node(
+    schemas_data: list[SchemaFile],
+    schema_index: int,
+    node_index: int,
+    container: str = "nodes",
```
ajtmccarty: optional, b/c we didn't have the before and it's not really your job to clean up the typing, but it would be nice if this were either an `Enum` or a `Literal` to make it clear exactly what values are expected here

infrahub_sdk/ctl/schema.py
```diff
@@ -59,44 +59,65 @@ def display_schema_load_errors(response: dict[str, Any], schemas_data: list[Sche
...
+        (node.get("kind") or node.get("name") or "")
+        if is_extension
+        else f"{node.get('namespace', None)}{node.get('name', None)}"
+    )
+    path_suffix = f" (extensions/{container})" if is_extension else ""
+    input_str = error.get("input")
+
+    if len(tail) == 1:
+        loc_type = tail[0]
+        error_message = f"{loc_type} ({input_str}) | {error['msg']} ({error['type']})"
```
ajtmccarty: are we sure that the `msg` and `type` keys exist here? if not this will raise a `KeyError`. it would be safer, and more reasable, to do `err_msg = error.get("msg", "" \ "No error message")` and then use `err_msg` and `err_type` in the f-string.

infrahub_sdk/ctl/schema.py
```diff
@@ -59,44 +59,65 @@ def display_schema_load_errors(response: dict[str, Any], schemas_data: list[Sche
...
-                f"  Node: {node.get('namespace', None)}{node.get('name', None)} | {error_message}", markup=False
-            )
+
+def _resolve_attribute_label(error_data: list[dict[str, Any]], attribute: Any) -> str | None:
+    if isinstance(attribute, str):
+        for data in error_data:
+            if data.get(attribute) is not None:
+                return data.get("name", None)
+        return None
+    return error_data[attribute].get("name", None)
```
ajtmccarty: not your fault, but `error_data.get(attribute,{}).get(...)` would be safer

infrahub_sdk/ctl/schema.py
```diff
@@ -59,44 +59,65 @@ def display_schema_load_errors(response: dict[str, Any], schemas_data: list[Sche
...
-            )
+    node_label = (
+        (node.get("kind") or node.get("name") or "")
+        if is_extension
+        else f"{node.get('namespace', None)}{node.get('name', None)}"
+    )
+    path_suffix = f" (extensions/{container})" if is_extension else ""
+    input_str = error.get("input")
+
+    if len(tail) == 1:
```
ajtmccarty, line 110: also not your fault, but adding a few one-line comments in here would make it easier to understand this pretty opaque error string parsing

ajtmccarty (commented):
thanks for the PR
most of these comments are little things. the testing structure one is larger though

infrahub_sdk/ctl/schema.py
iddocohen: I believe `error_data` is a `list `and not a `dict` hence `.get() `will not work but understood the spirt of the concern and will do an out of bound check.

tests/unit/sdk/test_schema.py
iddocohen: Completed a full refactor, particularly around where to display the error, which was needed some time to think about how.

infrahub_sdk/ctl/schema.py
```diff
@@ -49,73 +49,130 @@ def validate_schema_content_and_exit(client: InfrahubClient, schemas: list[Schem
...
+    if loc_path[3] == "extensions":
+        return (
+            len(loc_path) >= 7
+            and loc_path[4] in {"nodes", "generics", "relationships"}
+            and isinstance(loc_path[5], int)
+        )
+    return loc_path[3] in {"nodes", "generics"} and isinstance(loc_path[4], int)
+
+
+SchemaContainer = Literal["nodes", "generics", "relationships"]
```
ajtmccarty: generally, this kind of definition goes near the top of the file. but if it is only used in one place, you should be able to do `get_node(..., container: Literal["nodes", "generics", "relationships"] = "nodes", ...)`

tests/integration/test_schema.py
```diff
@@ -43,3 +48,71 @@ async def test_schema_load_many(
...
+        response = await client.schema.load(schemas=[broken_schema])
+
+        assert response.errors, "Server should reject a forbidden field on an extensions/nodes entry"
+        assert "detail" in response.errors
+
+        schemas_data = [SchemaFile(location=Path("broken.yml"), content=broken_schema)]
+        buffer = StringIO()
+        output = Console(file=buffer, width=1000, force_terminal=False)
+        display_schema_load_errors(response=response.errors, schemas_data=schemas_data, output=output)
+        rendered = buffer.getvalue()
```
ajtmccarty: I did not realize it would be such a pain to get the output. is it possible to wrap `client.schema.load` in something like this from the `rich` [documentation](https://rich.readthedocs.io/en/stable/console.html#capturing-output)?
```
from rich.console import Console
console = Console()
with console.capture() as capture:
    console.print("[bold red]Hello[/] World")
str_output = capture.get()
```

ajtmccarty (commented):
thanks for the updates. looking good. just a couple more little comments

tests/integration/test_schema.py
iddocohen: I tried to do that but it didn't work. I will ask Claude for help here. Stay tuned.

ajtmccarty (approved):

PR #1014, opened 2026-05-12
IHS-119: Fix upsert not working for numberpool and attribute in hfid

### LINKED ISSUE #339: bug: Upsert not working for nodes that are assigning an attribute via a NumberPool and attribute in HFID
### Component

Python SDK

### Infrahub SDK version

1.10.0

### Current Behavior

Whenever we attempt to create an object such as a VLAN in an idempotent way when an attribute can use a `CoreNumberPool` such as `vlan_id` using the `.save(allow_upsert=True)`, we run into the following error message:

```
[{'message': "invalid literal for int() with base 10: 'VLAN ID Pool - Test'", 'locations': [{'line': 3, 'column': 5}], 'path': ['InfraVLANUpsert']}]
```

If we omit `allow_upsert=True`, it assigns a VLAN ID, but we're not able to idempotently manage these objects so when using this in a generator, we have to attempt to fetch the VLAN from the DB and if it doesn't exist, and then create it rather than just creating the object and calling `.save(allow_upsert=True)`.

Below is the full exception with the GraphQL query:
```
>>> vlan = client.create(kind="InfraVLAN", name="atl1_ilo_1", vlan_id=vlan_pool, site=site, status="Active", role="Management")
>>> vlan.save(allow_upsert=True)
URL: https://demo.infrahub.app/graphql/main
QUERY:

mutation {
    InfraVLANUpsert(
        data: {
            name: {
                value: "atl1_ilo_1"
            }
            vlan_id: {
                from_pool: {
                    id: "1832e9b4-e8b4-fcab-d4e4-1065092c9fee"
                }
            }
            status: {
                value: "Active"
            }
            role: {
                value: "Management"
            }
            site: {
                id: "1808d320-afbd-9656-d0e7-c515cf3a1de4"
            }
            hfid: [
                "atl1_ilo_1",
                "VLAN ID Pool - Test",
            ]
        }
    ){
        ok
        object {
            id
            vlan_id {
                value
            }
        }
    }
}

  File "/Users/Mikhail/cloned/infrahub-sdk-python/infrahub_sdk/node.py", line 1741, in save
    self.create(allow_upsert=allow_upsert, timeout=timeout, request_context=request_context)
  File "/Users/Mikhail/cloned/infrahub-sdk-python/infrahub_sdk/node.py", line 1987, in create
    response = self._client.execute_graphql(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Mikhail/cloned/infrahub-sdk-python/infrahub_sdk/client.py", line 1674, in execute_graphql
    raise GraphQLError(errors=response["errors"], query=query, variables=variables)
infrahub_sdk.exceptions.GraphQLError: An error occurred while executing the GraphQL Query
mutation {
    InfraVLANUpsert(
        data: {
            name: {
                value: "atl1_ilo_1"
            }
            vlan_id: {
                from_pool: {
                    id: "1832e9b4-e8b4-fcab-d4e4-1065092c9fee"
                }
            }
            status: {
                value: "Active"
            }
            role: {
                value: "Management"
            }
            site: {
                id: "1808d320-afbd-9656-d0e7-c515cf3a1de4"
            }
            hfid: [
                "atl1_ilo_1",
                "VLAN ID Pool - Test",
            ]
        }
    ){
        ok
        object {
            id
            vlan_id {
                value
            }
        }
    }
}
, [{'message': "invalid literal for int() with base 10: 'VLAN ID Pool - Test'", 'locations': [{'line': 3, 'column': 5}], 'path': ['InfraVLANUpsert']}]
```

### Expected Behavior

Manage objects idempotently when we pull an attribute from a `CoreNumberPool` for the node. This is possible with `CoreIPPrefixPool` and `CoreIPAddressPool`.

### Steps to Reproduce

1. Log into https://demo.infrahub.app
2. Create a resource pool for `InfraVLAN` for the `vlan_id` attribute
3. Launch Python
```python
from infrahub_sdk.client import InfrahubClientSync, Config
client = InfrahubClientSync(config=Config(address="https://demo.infrahub.app", username="otto", password="infrahub"))
vlan_pool = client.get(kind="CoreNumberPool", name__value="VLAN ID Pool - Test")
site = client.get(kind="LocationSite", name__value="atl1")
vlan = client.create(kind="InfraVLAN", name="atl1_ilo_1", vlan_id=vlan_pool, site=site, status="Active", role="Management")
vlan.save(allow_upsert=True)
```

### Additional Information

_No response_

solababs:
# Why

`node.save(allow_upsert=True)` on a node whose human-friendly identifier (HFID) contains a `CoreNumberPool`-sourced attribute crashes with `"invalid literal for int() with base 10: 'VLAN ID Pool - Test'"`. The backend upsert resolver tries to use the unresolved pool reference as an HFID component, but the integer value is not assigned until after the server creates the node — making the upsert semantically impossible.

This PR adds a client-side guard that detects the pattern and raises a descriptive `ValidationError` before any network call is made.

Closes https://github.com/opsmill/infrahub-sdk-python/issues/339

## How to test

```bash
cd python_sdk

# All four upsert guard tests
uv run pytest tests/unit/sdk/pool/test_attribute_from_pool.py -v -k "upsert"
```

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added (`python_sdk/changelog/339.fixed.md`)
- [x] External docs updated (`docs/docs/python-sdk/guides/resource-manager.mdx`)

---
## Summary by cubic
Prevents crashes when upserting nodes whose HFID includes a `CoreNumberPool` attribute by failing fast with a clear `ValidationError`. Addresses IHS-119.

- **Bug Fixes**
  - Added `_validate_upsert()` and call it from `create()` (async and sync) so both `save(allow_upsert=True)` and `create(allow_upsert=True)` raise early when any HFID component is an unresolved pool-backed attribute (a `from_pool` dict with no value or a pool node object).
  - Upsert still works when `id` is set or when the pool-backed attribute is not in the HFID. Added `is_unresolved_pool_attribute()`, tests for async/sync paths, updated docs with alternatives, and a changelog entry.

<sup>Written for commit 5ac552fc9896a2f375e425177683091f29ab0a1b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1014?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/infrahub_sdk/node/node.py
+++ b/infrahub_sdk/node/node.py
@@ -595,6 +595,36 @@ class InfrahubNodeBase:
 
         raise ResourceNotDefinedError(message=f"The node doesn't have an attribute for {name}")
 
+    def _validate_upsert(self, allow_upsert: bool) -> None:
+        """Ensure an upsert can resolve the HFID before attempting to save.
+
+        An attribute sourced from a CoreNumberPool has no concrete value until the node is
+        created, so it cannot be used to look up an existing node by its human-friendly identifier.
+
+        Raises:
+            ValidationError: If an HFID attribute is sourced from an unresolved CoreNumberPool.
+
+        """
+        if not (allow_upsert and not self.id):
+            return
+
+        for hfid_path in self._schema.human_friendly_id or []:
+            attr_name = hfid_path.split("__")[0]
+            try:
+                attr = self._get_attribute(attr_name)
+            except ResourceNotDefinedError:
+                continue
+            if attr.is_unresolved_pool_attribute():
+                raise ValidationError(
+                    identifier=attr_name,
+                    message=(
+                        f"Attribute '{attr_name}' is sourced from a CoreNumberPool and is part of "
+                        "this node's human-friendly identifier. Upsert cannot resolve the HFID "
+                        "without a concrete value. Use an explicit id, or create the node first "
+                        "and update it in a separate call."
+                    ),
+                )
+
     @staticmethod
     def _build_rel_query_data(
         rel_schema: RelationshipSchemaAPI,
@@ -979,23 +1009,7 @@ class InfrahubNode(InfrahubNodeBase):
         timeout: int | None = None,
         request_context: RequestContext | None = None,
     ) -> None:
-        if allow_upsert and not self.id:
-            for hfid_path in self._schema.human_friendly_id or []:
-                attr_name = hfid_path.split("__")[0]
-                try:
-                    attr = self._get_attribute(attr_name)
-                except ResourceNotDefinedError:
-                    continue
-                if attr.is_unresolved_pool_attribute():
-                    raise ValidationError(
-                        identifier=attr_name,
-                        message=(
-                            f"Attribute '{attr_name}' is sourced from a CoreNumberPool and is part of "
-                            "this node's human-friendly identifier. Upsert cannot resolve the HFID "
-                            "without a concrete value. Use an explicit id, or create the node first "
-                            "and update it in a separate call."
-                        ),
-                    )
+        self._validate_upsert(allow_upsert=allow_upsert)
 
         if self._existing is False or allow_upsert is True:
             await self.create(allow_upsert=allow_upsert, timeout=timeout, request_context=request_context)
@@ -1968,23 +1982,7 @@ class InfrahubNodeSync(InfrahubNodeBase):
         timeout: int | None = None,
         request_context: RequestContext | None = None,
     ) -> None:
-        if allow_upsert and not self.id:
-            for hfid_path in self._schema.human_friendly_id or []:
-                attr_name = hfid_path.split("__")[0]
-                try:
-                    attr = self._get_attribute(attr_name)
-                except ResourceNotDefinedError:
-                    continue
-                if attr.is_unresolved_pool_attribute():
-                    raise ValidationError(
-                        identifier=attr_name,
-                        message=(
-                            f"Attribute '{attr_name}' is sourced from a CoreNumberPool and is part of "
-                            "this node's human-friendly identifier. Upsert cannot resolve the HFID "
-                            "without a concrete value. Use an explicit id, or create the node first "
-                            "and update it in a separate call."
-                        ),
-                    )
+        self._validate_upsert(allow_upsert=allow_upsert)
 
         if self._existing is False or allow_upsert is True:
             self.create(allow_upsert=allow_upsert, timeout=timeout, request_context=request_context)
--- a/infrahub_sdk/node/node.py
+++ b/infrahub_sdk/node/node.py
@@ -986,7 +986,7 @@ class InfrahubNode(InfrahubNodeBase):
                     attr = self._get_attribute(attr_name)
                 except ResourceNotDefinedError:
                     continue
-                if attr.value is None and attr.is_from_pool_attribute():
+                if attr.is_unresolved_pool_attribute():
                     raise ValidationError(
                         identifier=attr_name,
                         message=(
@@ -1975,7 +1975,7 @@ class InfrahubNodeSync(InfrahubNodeBase):
                     attr = self._get_attribute(attr_name)
                 except ResourceNotDefinedError:
                     continue
-                if attr.value is None and attr.is_from_pool_attribute():
+                if attr.is_unresolved_pool_attribute():
                     raise ValidationError(
                         identifier=attr_name,
                         message=(
--- a/infrahub_sdk/node/node.py
+++ b/infrahub_sdk/node/node.py
@@ -7,7 +7,13 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, BinaryIO, overload
 
 from ..constants import InfrahubClientMode
-from ..exceptions import FeatureNotSupportedError, NodeNotFoundError, ResourceNotDefinedError, SchemaNotFoundError
+from ..exceptions import (
+    FeatureNotSupportedError,
+    NodeNotFoundError,
+    ResourceNotDefinedError,
+    SchemaNotFoundError,
+    ValidationError,
+)
 from ..file_handler import FileHandler, FileHandlerBase, FileHandlerSync, PreparedFile, sha1_of_source
 from ..graphql import Mutation, Query
 from ..schema import (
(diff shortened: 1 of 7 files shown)


infrahub_sdk/node/node.py
```diff
@@ -1944,6 +1968,24 @@ def save(
...
         request_context: RequestContext | None = None,
     ) -> None:
+        if allow_upsert and not self.id:
+            for hfid_path in self._schema.human_friendly_id or []:
+                attr_name = hfid_path.split("__")[0]
+                try:
+                    attr = self._get_attribute(attr_name)
+                except ResourceNotDefinedError:
+                    continue
+                if attr.is_from_pool_attribute():
```
ogenstad: I'm wondering if we should forbid this all together if the attribute can come from a pool. It could be that we should have 

```python
if attr.value is not None and attr.is_from_pool_attribute():
```

I.e. that this would be problematic for upsert mutations if a value hadn't been assigned from a pool already. Was that something that you considered?

infrahub_sdk/node/node.py
```diff
@@ -973,6 +979,24 @@ async def save(
         timeout: int | None = None,
         request_context: RequestContext | None = None,
     ) -> None:
+        if allow_upsert and not self.id:
```
ogenstad: The bug report specifically mentions `Node.save()` as a problem area but I think we can have a similar situation in `Node.create()`. One additional comment here is that the code looks exactly the same for the sync and async version, with that in mind I think we can create a helper method that can be called from both of these to ensure that we keep the code in sync in the future.

ogenstad (commented):
Looks good overall, but as we're testing with mocked data it could be nice either have an integration test or at least validate against a real server.

tests/unit/sdk/pool/test_attribute_from_pool.py
```diff
@@ -201,3 +206,144 @@ async def test_attribute_with_pool_node_generates_mutation_query(
...
+}
+
+
+@pytest.mark.httpx_mock(assert_all_responses_were_requested=False)
+async def test_save_upsert_no_error_before_fix(
+    client: InfrahubClient,
+    vlan_schema_with_pool_hfid: NodeSchemaAPI,
+    httpx_mock: HTTPXMock,
+) -> None:
+    """FAILS before the guard is added: ValidationError must be raised when HFID attr is NumberPool-sourced."""
```
polmichel: Nitpick and optional comment: test description should in my opinion describe what's the expected behavior from the system, and not stating what has been done in order to reach this state (like fixing an older bug)

tests/unit/sdk/pool/test_attribute_from_pool.py
```diff
@@ -201,3 +206,144 @@ async def test_attribute_with_pool_node_generates_mutation_query(
...
+    with pytest.raises(ValidationError, match="vlan_id"):
+        await node.save(allow_upsert=True)
+
+
+async def test_save_upsert_proceeds_when_explicit_id_set(
+    client: InfrahubClient,
+    vlan_schema_with_pool_hfid: NodeSchemaAPI,
+    httpx_mock: HTTPXMock,
+) -> None:
+    """Guard is bypassed when an explicit node id is already set."""
```
polmichel: Nitpick and optional comment: I do not think knowing there is a "guard" is important. What's important is the use case that this test describes.

tests/unit/sdk/pool/test_attribute_from_pool.py
```diff
@@ -201,3 +206,144 @@ async def test_attribute_with_pool_node_generates_mutation_query(
...
+    """FAILS before the guard is added: ValidationError must be raised when HFID attr is NumberPool-sourced."""
+    httpx_mock.add_response(method="POST", json=UPSERT_MOCK_RESPONSE)
+    node = InfrahubNode(
+        client=client,
+        schema=vlan_schema_with_pool_hfid,
+        data={"name": "Test VLAN", "vlan_id": {"from_pool": {"id": POOL_ID}}},
+    )
+
+    with pytest.raises(ValidationError, match="vlan_id"):
+        await node.save(allow_upsert=True)
```
polmichel: Optional comment: after some discussions with @gmazoyer and @ajtmccarty, we ended up agreeing that exact error message matches would create better assertions in our test suites.

tests/unit/sdk/pool/test_attribute_from_pool.py
```diff
@@ -201,3 +206,144 @@ async def test_attribute_with_pool_node_generates_mutation_query(
...
+    vlan_schema_with_pool_hfid: NodeSchemaAPI,
+) -> None:
+    """save(allow_upsert=True) raises ValidationError naming the pool-sourced HFID attribute."""
+    node = InfrahubNode(
+        client=client,
+        schema=vlan_schema_with_pool_hfid,
+        data={"name": "Test VLAN", "vlan_id": {"from_pool": {"id": POOL_ID}}},
+    )
+
+    with pytest.raises(ValidationError, match="vlan_id"):
```
polmichel: Same optional comment about error message assertions

tests/unit/sdk/pool/test_attribute_from_pool.py
polmichel: I am not sure what's the purpose of this test: mocking the API behavior to make sure that if we remove the fix, the application is still failing?

tests/unit/sdk/pool/test_attribute_from_pool.py
```diff
@@ -201,3 +206,144 @@ async def test_attribute_with_pool_node_generates_mutation_query(
...
+
+
+@pytest.mark.httpx_mock(assert_all_responses_were_requested=False)
+async def test_save_upsert_no_error_before_fix(
+    client: InfrahubClient,
+    vlan_schema_with_pool_hfid: NodeSchemaAPI,
+    httpx_mock: HTTPXMock,
+) -> None:
+    """FAILS before the guard is added: ValidationError must be raised when HFID attr is NumberPool-sourced."""
+    httpx_mock.add_response(method="POST", json=UPSERT_MOCK_RESPONSE)
```
polmichel: The mock response is not even used -> `@pytest.mark.httpx_mock(assert_all_responses_were_requested=False)`

tests/unit/sdk/pool/test_attribute_from_pool.py
polmichel: I think we should remove this test

tests/unit/sdk/pool/test_attribute_from_pool.py
```diff
@@ -201,3 +206,144 @@ async def test_attribute_with_pool_node_generates_mutation_query(
...
+
+    assert node.id == "mock-vlan-uuid"
+
+
+async def test_save_upsert_proceeds_when_numberpool_attr_not_in_hfid(
+    client: InfrahubClient,
+    vlan_schema: NodeSchemaAPI,
+    httpx_mock: HTTPXMock,
+) -> None:
+    """Guard does not fire when the pool-sourced attribute is not part of the HFID."""
```
polmichel: Same comment about test description

tests/unit/sdk/pool/test_attribute_from_pool.py
polmichel: Same inside other tests

tests/unit/sdk/pool/test_attribute_from_pool.py
polmichel: Same inside other tests

polmichel (commented):
The coverage is great and the fix is nicely implemented.
That said, there is a test that is just noisy and, in my opinion, does not bring any value to our test suite. Maybe it would be better to remove it.

Optional nitpick comments about test descriptions and test assertions have also been added.

polmichel (approved):

PR #1030, opened 2026-05-17
fix: add labels to NodeNotFoundError for clearer messaging

DharmaBytesX:
The error message now shows Branch:, Kind:, and Identifier: labels
instead of raw pipe-separated values, helping users understand
which branch the lookup failed on.

**Before:**
```
        Unable to find the node in the database.
        main | InfraDevice | {'name__value': 'bad-device404'}
```

**After:**
```
        Unable to find the node in the database.
        Branch: main | Kind: InfraDevice | Identifier: {'name__value': 'bad-device404'}
```

Fixes opsmill/infrahub#5514

---
## Summary by cubic
Add labels to `NodeNotFoundError` and require keyword-only `branch_name`, `node_type`, and `identifier`, producing consistent messages like "Branch: ... | Kind: ... | Identifier: ...".
Updated file download paths to pass branch and node ID, and store/CTL lookups to include branch (falling back to the client’s default), so all `NodeNotFoundError` instances carry full context.

<sup>Written for commit 1ab16ae8a91c65bd4993d024a745604882a23fbb. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1030?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/infrahub_sdk/exceptions.py
+++ b/infrahub_sdk/exceptions.py
@@ -71,29 +71,22 @@ class ModuleImportError(Error):
 class NodeNotFoundError(Error):
     def __init__(
         self,
-        identifier: Mapping[str, list[str]],
+        *,
+        branch_name: str,
+        node_type: str,
+        identifier: Mapping[str, list[str]] | str,
         message: str = "Unable to find the node in the database.",
-        branch_name: str | None = None,
-        node_type: str | None = None,
     ) -> None:
-        self.node_type = node_type or "unknown"
-        self.identifier = identifier
         self.branch_name = branch_name
-
+        self.node_type = node_type
+        self.identifier = identifier
         self.message = message
         super().__init__(self.message)
 
     def __str__(self) -> str:
-        detail_parts = []
-        if self.branch_name:
-            detail_parts.append(f"Branch: {self.branch_name}")
-        if self.node_type:
-            detail_parts.append(f"Kind: {self.node_type}")
-        detail_parts.append(f"Identifier: {self.identifier}")
-        detail = " | ".join(detail_parts)
         return f"""
         {self.message}
-        {detail}
+        Branch: {self.branch_name} | Kind: {self.node_type} | Identifier: {self.identifier}
         """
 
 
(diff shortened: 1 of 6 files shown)


DharmaBytesX:
Hi! A small fix here to improve the NodeNotFoundError message — could someone give it a quick review when you get a chance? Thanks!

infrahub_sdk/exceptions.py
```diff
@@ -84,9 +84,16 @@ def __init__(
         super().__init__(self.message)
 
     def __str__(self) -> str:
+        detail_parts = []
+        if self.branch_name:
+            detail_parts.append(f"Branch: {self.branch_name}")
+        if self.node_type:
+            detail_parts.append(f"Kind: {self.node_type}")
```
ogenstad: I don't think we should have if-statements within the `__str__()` method of classes in general. I think the problem here is really that the parameters to the `__init__()` method are optional instead of being required. We should probably look at the places where we raise this error and check if we are always sending in all fields (or if we have enough information to do so) and then update the init method to require there parameters.

DharmaBytesX: Thanks for the review. I've reworked the change along the lines you suggested: `branch_name`, `node_type` and `identifier` are now required keyword-only parameters of `NodeNotFoundError.__init__`, and `__str__` no longer has any conditional logic.

I audited the call sites and updated them to always pass the three fields:
- `infrahub_sdk/client.py` was already passing all three.
- `infrahub_sdk/ctl/object/utils.py` now resolves the branch to `client.default_branch` when the caller did not provide one.
- `infrahub_sdk/file_handler.py`: `handle_response` and `handle_error_response` now take `branch` and `node_id`, plumbed from `FileHandler.download` / `_stream_to_file` in both the async and sync paths.
- `infrahub_sdk/store.py`: passes `self.branch_name`, and falls back to `"unknown"` only on the internal lookup paths where the caller genuinely never specified a kind.

Also added `tests/unit/sdk/test_exceptions.py` covering the new contract (keyword-only, all three required, rendered format).

PR #1048, opened 2026-05-28
fix(ctl): let typer.Exit propagate so CLI commands exit without a traceback

### LINKED ISSUE #1047: infrahubctl: typer.Exit leaks raw traceback on schema load failure
## Component

infrahubctl

## Infrahub SDK version

v1.20.1

## Current Behavior

When `infrahubctl schema load` receives an error response from the server, it
correctly prints the human-readable error, then exposes a raw Python traceback:

```
Unable to load the schema:
  Unable to find the schema 'TemplateDcimDevice' in the registry
Error: 1
  File ".../infrahub_sdk/ctl/utils.py", line 93, in async_wrapper
    return await func(*args, **kwargs)
  File ".../infrahub_sdk/ctl/schema.py", line 201, in load
    raise typer.Exit(1)
typer._click.exceptions.Exit: 1
```

The `Error: 1` line and the traceback below it are implementation noise — the
useful message has already been printed above.

**Root cause**: `typer.Exit` inherits from `BaseException`, not `Exception`. The
`catch_exception` decorator in `utils.py` only catches `(Error, Exception)`, so
`typer.Exit` propagates uncaught through `async_wrapper`. Because `typer.Exit` is
not `SystemExit`, Python's default handler prints the full traceback instead of
silently exiting.

## Expected Behavior

After printing the human-readable error, the command exits with code 1 and no
traceback. The terminal output should be:

```
Unable to load the schema:
  Unable to find the schema 'TemplateDcimDevice' in the registry
```

## Steps to Reproduce

1. Have a running Infrahub instance
2. Load a schema that triggers any server-side validation error (e.g. a
   relationship peer that doesn't exist)
3. Run `infrahubctl schema load <schema-dir>`
4. Observe the traceback after the error message

## Additional Information

The fix belongs in `AsyncTyper` or `catch_exception` — either intercept
`typer.Exit` / `BaseException` and call `sys.exit(code)`, or exclude
`typer.Exit` from the catch clause so Typer's own machinery handles it cleanly.

Related: opsmill/infrahub#9364 (the server-side error that surfaced this UX
issue)

ogenstad:
# Why

Newer versions of typer behave differently and produce an error that's not as friendly as we'd want and we don't want to show that error.

Closes #1047

## What changed

* Catch error from typer early and avoid showing it to the end user
* Tests

---
## Summary by cubic
Fixes `infrahubctl` so commands that raise `typer.Exit` exit cleanly without the "Error: 1" line or a traceback. Addresses IHS-237 by ensuring `infrahubctl schema load` prints only the human-readable error and exits with code 1.

- **Bug Fixes**
  - Let `typer.Exit` pass through in `catch_exception` (sync/async) and `handle_exception`, removing `click.exceptions.Exit` interception so Typer exits cleanly without a traceback.
  - Add tests for `typer.Typer` and `AsyncTyper` to verify clean output, correct exit code, and no "Error: 1" line.

<sup>Written for commit b39470721e82ccea8360f5c84e6f47f01036e1f1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1048?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/infrahub_sdk/ctl/utils.py
+++ b/infrahub_sdk/ctl/utils.py
@@ -9,7 +9,6 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, NoReturn, TypeVar
 
 import typer
-from click.exceptions import Exit
 from httpx import HTTPError
 from rich.console import Console
 from rich.logging import RichHandler
@@ -54,8 +53,6 @@ def init_logging(debug: bool = False) -> None:
 
 def handle_exception(exc: Exception, console: Console, exit_code: int) -> NoReturn:
     """Handle exception in a different fashion based on its type."""
-    if isinstance(exc, Exit):
-        raise typer.Exit(code=exc.exit_code)
     if isinstance(exc, AuthenticationError):
         console.print(f"[red]Authentication failure: {exc!s}")
         raise typer.Exit(code=exit_code)
@@ -91,6 +88,8 @@ def catch_exception(
             async def async_wrapper(*args: Any, **kwargs: Any) -> T:
                 try:
                     return await func(*args, **kwargs)
+                except typer.Exit:
+                    raise
                 except (Error, Exception) as exc:
                     return handle_exception(exc=exc, console=console, exit_code=exit_code)
 
@@ -100,6 +99,8 @@ def catch_exception(
         def wrapper(*args: Any, **kwargs: Any) -> T:
             try:
                 return func(*args, **kwargs)
+            except typer.Exit:
+                raise
             except (Error, Exception) as exc:
                 return handle_exception(exc=exc, console=console, exit_code=exit_code)
 
--- a/infrahub_sdk/ctl/utils.py
+++ b/infrahub_sdk/ctl/utils.py
@@ -53,6 +53,8 @@ def init_logging(debug: bool = False) -> None:
 
 def handle_exception(exc: Exception, console: Console, exit_code: int) -> NoReturn:
     """Handle exception in a different fashion based on its type."""
+    if isinstance(exc, typer.Exit):
+        raise exc
     if isinstance(exc, AuthenticationError):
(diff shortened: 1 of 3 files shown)


infrahub_sdk/ctl/utils.py
```diff
@@ -91,6 +88,8 @@ def decorator(func: Callable[..., T]) -> Callable[..., T | Coroutine[Any, Any, T
             async def async_wrapper(*args: Any, **kwargs: Any) -> T:
                 try:
                     return await func(*args, **kwargs)
+                except typer.Exit:
+                    raise
```
ajtmccarty: I am a little surprised this fixes the issue. it looks like it's just catching the exception and then raising it again

ajtmccarty (approved):

infrahub_sdk/ctl/utils.py
ogenstad: I can agree that it looks a bit confusing. What happens is that it's interrupting the catch just below where we send this into handle_exception. But I'll look at moving this instead into `handle_exception` which is a bit cleaner anyway.

ajtmccarty (approved):

infrahub_sdk/ctl/utils.py
ogenstad: @ajtmccarty, do you think it's cleaner with the changes in b39470721e82ccea8360f5c84e6f47f01036e1f1?

ajtmccarty: yes I think this is easier to understand now

Issue #1064, opened 2026-06-05
bug: assigning to a relationship attribute fails mypy (assignment) for id strings and nodes


FragmentedPacket:
### Component

Python SDK, infrahubctl

### Infrahub SDK version

1.20.1

### Current Behavior

Generated protocols type relationship attributes as `RelatedNode` (the *read* type). At runtime the SDK accepts assigning either an id string or a node object to set a relationship (`node.rel = "<id>"` or `node.rel = other_node`). Under mypy both forms fail with `assignment`, because the annotation only models what you read back, not what is assignable.

Root cause — generated protocols declare relationships as a single read type, e.g. in `infrahub_sdk/protocols.py`:

```python
class BuiltinIPAddress(CoreNode):
    ip_namespace: RelatedNode
    ip_prefix: RelatedNode
```

A bare attribute annotation uses the same type for get and set, but the runtime `__setattr__` accepts `str | CoreNode | RelatedNode`. mypy therefore rejects the runtime-valid assignment forms. This affects every consumer that sets relationships and runs mypy.

Reproduces on both mypy 2.1.0 and mypy 1.17.1 (not a mypy-version regression).

### Expected Behavior

Setting a relationship via an id string or a node — the documented way to establish relationships before `save()` — should type-check.

### Steps to Reproduce

```python
from infrahub_sdk import InfrahubClient
from infrahub_sdk.protocols import BuiltinIPAddress


async def main(client: InfrahubClient) -> None:
    addr = await client.get(kind=BuiltinIPAddress, id="x")
    addr.ip_prefix = "some-prefix-id"            # set by id string

    other = await client.get(kind=BuiltinIPAddress, id="y")
    addr.ip_namespace = other                    # set by node
```

Run mypy (infrahub-sdk 1.20.1):

```
error: Incompatible types in assignment (expression has type "str", variable has type "RelatedNode")  [assignment]
    addr.ip_prefix = "some-prefix-id"
                     ^~~~~~~~~~~~~~~~
error: Incompatible types in assignment (expression has type "BuiltinIPAddress", variable has type "RelatedNode")  [assignment]
    addr.ip_namespace = other
                        ^~~~~
```

(The same error also commonly appears when assigning `RelatedNode.id`, i.e. `str | None`, to another relationship attribute.)

### Additional Information

**Impact**

- Affects all relationship-setting code (generators, data loaders, migrations).
- Forces `# type: ignore[assignment]` on every relationship assignment.

[...]

PR #763, opened 2026-01-22
IHS-183: Fix typing errors for protocols

solababs:
Fixes https://github.com/opsmill/infrahub-sdk-python/issues/697

Replace protocols with concrete classes to address type-abstract mypy errors

## Summary by CodeRabbit

* **Refactor**
  * Enhanced flexibility for related node handling to support duck-typing patterns.
  * Refactored internal class hierarchy for improved type safety and explicit error handling.
  * Updated schema validation logic for more robust type checking.

* **Chores**
  * Updated Ruff development tool to latest version.
  * Removed unnecessary type-checking suppressions throughout codebase.

* **Tests**
  * Added test coverage for schema name resolution.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1065, opened 2026-06-05
Make RelatedNode/RelationshipManager generic over peer type (#1063)

### LINKED ISSUE #1063: bug: RelatedNode.peer is typed InfrahubNode, losing the peer type during relationship traversal
### Component

Python SDK, infrahubctl

### Infrahub SDK version

1.20.1

### Current Behavior

`RelatedNode.peer` is annotated to return `InfrahubNode` (the dynamic, untyped node) rather than the typed peer. As a result, any relationship traversal through `.peer` loses all static type information: the next attribute access resolves to the catch-all union `Attribute | RelationshipManager | RelatedNode`, so `node.rel.peer.<attr>.value` fails under mypy with `union-attr`. Chains such as `interface.device.peer.rack.peer.name.value` produce a cascade of errors.

Root cause:

- `infrahub_sdk/node/related_node.py` — `def peer(self) -> InfrahubNode: ...`

`RelatedNode` is not parameterized by the peer type, so `peer` cannot return anything more specific than `InfrahubNode`, and `InfrahubNode.__getattr__` returns the union `Attribute | RelationshipManager | RelatedNode`. This affects every consumer that walks relationships using the generated typed protocols and runs mypy.

Reproduces on both mypy 2.1.0 and mypy 1.17.1 (not a mypy-version regression).

### Expected Behavior

`reveal_type(artifact.definition.peer)` should be `CoreArtifactDefinition` (the typed peer), and `artifact.definition.peer.name.value` should type-check as `str`.

### Steps to Reproduce

```python
from infrahub_sdk.protocols import CoreArtifact


async def main(client: InfrahubClient) -> None:
    artifact = await client.get(kind=CoreArtifact, id="x")
    reveal_type(artifact.definition.peer)        # -> infrahub_sdk.node.node.InfrahubNode
    _ = artifact.definition.peer.name.value      # union-attr
```

Run mypy (infrahub-sdk 1.20.1):

```
note:  Revealed type is "infrahub_sdk.node.node.InfrahubNode"
error: Item "RelationshipManager" of "Attribute | RelationshipManager | RelatedNode" has no attribute "value"  [union-attr]
error: Item "RelatedNode" of "Attribute | RelationshipManager | RelatedNode" has no attribute "value"  [union-attr]
    _ = artifact.definition.peer.name.value
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

`reveal_type` shows the peer comes back as `InfrahubNode` instead of `CoreArtifactDefinition`, so the typed schema is lost after the first hop.

### Additional Information

**Impact**

- The single largest source of type errors for relationship-heavy code. In one downstream repo this one issue accounts for ~56 of ~101 mypy errors after the 1.16 → 1.20 upgrade.
- Forces `cast(...)` at every `.peer` boundary or blanket `# type: ignore[union-attr]`.

**Suggested fix**

Make `RelatedNode` generic over the peer type:

```python
PeerT = TypeVar("PeerT", bound=CoreNode)

class RelatedNode(RelatedNodeBase, Protocol[PeerT]):
    @property
    def peer(self) -> PeerT: ...
```

and have `infrahubctl protocols` emit relationship attributes as `RelatedNode[CoreArtifactDefinition]` (and `RelationshipManager[...]` similarly) instead of bare `RelatedNode`. This restores type information across traversals and removes the need for downstream casts. This is likely the highest-value of the related typing fixes, since correct relationship typing is the main reason to use the generated protocols.

**Downstream workaround (interim)**

Cast at each `.peer` boundary to the known peer protocol (more truthful than a blanket ignore, and keeps downstream access checked):

```python
from typing import cast
from infrahub_sdk.protocols import CoreArtifactDefinition

definition = cast("CoreArtifactDefinition", artifact.definition.peer)
_ = definition.name.value  # now checks as str
```

For multi-hop chains, introduce a typed local per hop:

```python
# instead of: src_interface.peer.device.peer.rack.peer.name.value
interface = cast("NetworkInterface", endpoint.peer)
device = cast("NetworkDevice", interface.device.peer)
rack_name = cast("LocationRack", device.rack.peer).name.value if device.rack.initialized else ""
```

Once `RelatedNode` is generic, all of these casts can be deleted.

---

Filed separately: companion issues for `type-abstract` on `kind=` and for relationship-attribute assignment being rejected (the assignment fix composes with the generic-`RelatedNode` change proposed here).

FragmentedPacket:
Fixes #1063. Stacked on top of #763 (base branch is #763's `re-sb-20260119-fix-typing-for-protocols-ihs-183`); retarget to `develop` once #763 merges.

## Problem

`RelatedNode.peer` / `RelationshipManager` are typed against the dynamic `InfrahubNode`, so traversing a relationship loses the peer's type. Any chain through `.peer` / `.peers` collapses to the catch-all `Attribute | RelationshipManager | RelatedNode` union and fails under mypy/ty with `union-attr`. This became visible to downstream consumers when `py.typed` was added in 1.19.0.

## Change

- `RelatedNode`, `RelatedNodeSync`, `RelationshipManager`, `RelationshipManagerSync` are now generic over the peer type (`PeerT` / `PeerTSync`), **defaulting to `InfrahubNode` / `InfrahubNodeSync`** (PEP 696) so all existing un-parameterised usage is unchanged.
- `infrahubctl protocols` parameterises each relationship: `device: RelatedNode[NetworkDevice]`, `interfaces: RelationshipManager[NetworkInterface]`, etc. Core peers are syncified in sync output (`tags: RelationshipManagerSync[BuiltinTagSync]`); locally generated classes keep their name.
- `Generic[...]` is a typing-only change — zero runtime cost.

## Verification

- SDK: `mypy` (125 files) ✅, `ty` ✅, `ruff` ✅, generator test updated, 252 node/relationship/store unit tests ✅.
- Against a downstream repo (the AI/DC solution) on the released line: relationship traversal `union-attr` errors went from **52 → 0**. (Remaining downstream errors are a separate relationship-`assignment` issue, #1064, plus genuinely-unsound accesses now correctly surfaced, e.g. accessing `NetworkInterface`-only attributes on a generic `NetworkEndpoint` peer.)

## Notes

- The bundled `infrahub_sdk/protocols.py` is left as-is (bare relationships still resolve via the `InfrahubNode` default); it can be regenerated with the new generator separately.

---
## Summary by cubic
Make `RelatedNode`/`RelatedNodeSync` and `RelationshipManager`/`RelationshipManagerSync` generic over their peer type so `.peer`, `.peers`, and indexing return the concrete peer; regenerate SDK docs to show `PeerT`/`PeerTSync` returns, fixing #1063. Defaults keep existing code working by falling back to `InfrahubNode`/`InfrahubNodeSync`.

- **New Features**
  - Peer type defaults (PEP 696) to `InfrahubNode`/`InfrahubNodeSync` for backward compatibility.
  - `infrahubctl protocols` outputs `RelatedNode[Peer]` and `RelationshipManager[Peer]`; sync output uses `PeerSync` when available.
  - Hierarchical fields are typed: `parent: RelatedNode[Self]`, `children: RelationshipManager[Self]`.

<sup>Written for commit da2e3f39682b25d06f88c2b57066cc95286fe772. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1065?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ajtmccarty (commented):
looks like a big typing improvement, but there are conflicts

ajtmccarty (approved):

PR #1067, opened 2026-06-07
Type cardinality-one relationships as an assignable descriptor (#1064)

### LINKED ISSUE #1064: bug: assigning to a relationship attribute fails mypy (assignment) for id strings and nodes

FragmentedPacket:
Fixes #1064. Stacked on #1063 (base branch is `ihs-relatednode-generic-1063`); retarget once #1063 / #763 merge.

## Problem

Generated cardinality-one relationships are typed read-only as `RelatedNode`, but the documented way to set a relationship is `node.rel = "<id>"` or `node.rel = peer_node`. Under mypy/ty (since `py.typed` in 1.19.0) every such assignment fails with `Incompatible types in assignment ... variable has type "RelatedNode"` (`assignment`).

## Change

- Adds a typing-only descriptor `RelationshipAttribute` / `RelationshipAttributeSync` (in `infrahub_sdk/node/related_node.py`). It reads back as `RelatedNode[Peer]` (so `.peer` keeps the peer type from #1063) and accepts assignment of `str | list[str] | Peer | None`, mirroring the runtime `InfrahubNode.__setattr__`, which wraps the assigned value in a `RelatedNode`.
- `infrahubctl protocols` now emits cardinality-one relationships as `RelationshipAttribute[Peer]` (cardinality-many stays `RelationshipManager[Peer]`).
- The descriptor only appears in generated protocols and is never instantiated at runtime — purely to give `node.rel` separate read/assign types.

## Verification

- SDK: `mypy` (125 files) ✅, `ty` ✅, `ruff` ✅, generator test updated, 258 node/relationship/protocol unit tests ✅.
- Against the downstream AI/DC repo: the relationship-`assignment` errors (`node.rel = "<id>"` / `= peer`) drop to **0**. Remaining downstream errors are unrelated (untyped `client.create(kind="...")` results that need a cast, and an `Optional` not handled) — not relationship assignments.

## Depends on

#1063 (generic `RelatedNode`) — `RelationshipAttribute.__get__` returns `RelatedNode[Peer]`.

---
## Summary by cubic
Make cardinality-one relationships assignable in type-checked code by adding a typing-only `RelationshipAttribute[...]` descriptor. Reads still return `RelatedNode[Peer]`, so `.peer` keeps its type and `node.rel = "<id>"` or a peer node now type-checks.

- **Bug Fixes**
  - Added `RelationshipAttribute`/`RelationshipAttributeSync` that accept `str | list[str] | Peer | None` and read as `RelatedNode[Peer]`; typing-only, never used at runtime.
  - Protocols now emit `RelationshipAttribute[Peer]` for cardinality-one (including `parent`); many stays `RelationshipManager[Peer]`.
  - Exported via `infrahub_sdk.node`; removes mypy/ty assignment errors without changing runtime behavior.

- **Migration**
  - Regenerate protocols with `infrahubctl protocols` to pick up the new typing; existing generated files won’t update automatically.

<sup>Written for commit c5887f0a9a4bc1b6df9723248eab0473d3ed8d2e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1067?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ogenstad (approved):
LGTM, this will be a great improvement! One thing I think is missing here is that the protocols will need to be recreated once this is in place and we should highlight this in the changelog newsfragment so that it's clear that it won't just work out of the box.

PR #1079, opened 2026-06-15
fix(node): preserve unfetched optional one-cardinality relationships on upsert

### LINKED ISSUE #1080: bug: from_graphql + save(allow_upsert=True) silently clears unfetched optional one-cardinality relationships
### Component

Python SDK

### Infrahub SDK version

1.21.1 (`infrahub_sdk/node/` on `stable`)

### Current Behavior

A node hydrated by `InfrahubNode.from_graphql` from a GraphQL response that did **not** fetch a given optional one-cardinality relationship will silently null-clear that relationship on the next `save(allow_upsert=True)`. The relationship payload sent to Infrahub looks like `<rel_name>: null` even though the caller never assigned to it — leaving the peer on the other side as an orphan in subsequent queries.

The root cause is at `infrahub_sdk/node/node.py:344-349`:

```python
if rel_schema.cardinality == RelationshipCardinality.ONE and rel_schema.optional and not rel.initialized:
    # Only include None for existing nodes to allow clearing relationships
    if self._existing:
        data[item_name] = None
    continue
```

`RelatedNode.initialized` is derived (`bool(id) or bool(hfid)`) and conflates two distinct cases:

1. **Never loaded** — the GraphQL response didn't include this rel.
2. **Explicitly cleared** — the user wrote `node.rel = None`.

Both produce `id is None AND hfid is None`. The inline comment shows the gate was written to support case (2), but case (1) silently falls through it.

### Expected Behavior

A node hydrated from a partial GraphQL payload that omits an optional one-cardinality relationship must **not** include that relationship in the upsert mutation. Only explicit caller mutations (`node.rel = None` or `node.rel = <something>`) should produce a relationship entry in the payload.

### Steps to Reproduce

Demonstrable as a unit test against `_generate_input_data` (no live server required):

```python
from infrahub_sdk import InfrahubNode

# location_schema has optional one-cardinality rel `primary_tag`
data = {
    "id": "existing-node-id",
    "name": {"value": "JFK1"},
    "description": {"value": "JFK Airport"},
    "type": {"value": "SITE"},
    # primary_tag deliberately absent — the upstream query didn't fetch it
}
node = InfrahubNode(client=client, schema=location_schema, data=data)
payload = node._generate_input_data()["data"]["data"]

assert "primary_tag" not in payload  # FAILS today — payload contains primary_tag: None
```

The real-world hit was an Infrahub demo branch where a generator iterated over BGP neighbors with a cascade query that fetched only `id`, `hfid`, `__typename`, and one attribute on each peer. Calling `save(allow_upsert=True)` on those hydrated peers silently wiped each one's `local_interface` relationship. The peers became orphan nodes — they existed in the database but were no longer reachable from interface-anchored queries (GraphQL count for `local_interface` peers dropped to 0 on the affected branch).

### Notes

- **Cardinality-many** is already safe via `RelationshipManager.initialized = data is not None` (`infrahub_sdk/node/relationship.py:149`) — fetched-empty and never-fetched are already distinguishable on that path.
- **Hierarchical relationships** were handled separately in #810.
- **Attribute side** already mitigates the analogous case via `Attribute.value_has_been_mutated` (`infrahub_sdk/node/attribute.py:73, 109, 122-123`).

So the fix is to mirror the same discipline on `RelatedNodeBase`. Submitted in #1079.

iddocohen:
Closes #1080

## Summary

A node hydrated by `from_graphql` from a partial GraphQL response would silently null-clear any optional cardinality-one relationship the response did not fetch, on the next `save(allow_upsert=True)`. This is the kind of bug that surfaces as data corruption in downstream queries — orphan nodes that exist but are no longer reachable from their previous side of the relationship.

The root cause is a conflation in the serialization gate at `infrahub_sdk/node/node.py:344-349`. For an `_existing` node, an "uninitialized" optional one-cardinality rel is emitted as `null`. But `RelatedNode.initialized` is a derived property (`bool(id) or bool(hfid)`), and that derivation cannot tell apart:

1. **Never loaded** — the GraphQL response didn't include this rel.
2. **Explicitly cleared** — the user wrote `node.rel = None`.

Both produce `id is None AND hfid is None`. The gate was written to support case (2), but case (1) silently falls through it and clears data the user never intended to touch.

## Fix

Mirror the existing `Attribute.value_has_been_mutated` pattern on relationships:

- Add `_peer_has_been_mutated: bool = False` to `RelatedNodeBase.__init__`.
- Flip it to `True` in both async and sync `Node.__setattr__` when the user assigns to a cardinality-one rel.
- Tighten the serialization gate to require `self._existing AND rel._peer_has_been_mutated` before emitting `null`.

Net effect: nodes hydrated from partial payloads no longer leak `rel: null` into upsert mutations. Explicit `node.rel = None` still clears the rel.

## Scope

- Cardinality-one optional relationships only.
- Cardinality-many is already safe: `RelationshipManager.initialized = data is not None` (relationship.py:149) — fetched-empty and never-fetched are already distinguishable there.
- Hierarchical relationships were handled separately in #810.

## Test plan

- [x] New test `test_update_input_data_existing_node_preserves_untouched_optional_relationship` fails on the old code, passes after the fix (both `[standard]` and `[sync]` variants).
- [x] Existing `test_update_input_data_existing_node_with_optional_relationship` updated to actually call `node.primary_tag = None` (matching what its docstring already claimed) — verifies the explicit-clear path still works.
- [x] Full SDK unit suite passes (967/967).
- [x] `uv run invoke format lint-code` clean.
- [ ] Integration tests TBD via CI.


changelog/1080.fixed.md
ajtmccarty, line 1: the changelog file name could reference the issue being closed so it gets linked automatically

ajtmccarty (approved):
I ran the full test suite of the community version against these changes to the SDK in this draft PR (https://github.com/opsmill/infrahub/pull/9583). I got a few failures in there, but I don't think that they are related to the changes in this PR.

changelog/1080.fixed.md
iddocohen: I asked Claude to check too, and it confirmed that this change has nothing to do with the failure (re-run it too). Seems to be a previous one. 

Changed the log accordingly too.

PR #1090, opened 2026-06-18
feat(sdk): add graph traversal support for Infrahub 1.10+

minitriga:
## Overview

Adds graph-traversal support to the SDK (requires Infrahub 1.10+), letting users discover how nodes are connected without knowing the relationship path in advance. Purely additive — no existing signatures change.

## What's new

Three client methods, each with a sync equivalent:

- **`traverse_paths(source, destination, ...)`** — find the shortest path(s) between two nodes.
- **`path_exists(source, destination, ...)`** — boolean convenience wrapper for checks (requests a single path).
- **`reachable_nodes(source, target_kinds, ...)`** — find every node of given kinds reachable from a source, with the path to each.

Ergonomics:

- `source`/`destination` accept a node UUID string **or** an `InfrahubNode` instance.
- Kind filters accept kind-name strings **or** generated protocol classes (mix freely).
- Each result `PathNode` is a lightweight identity exposing `.fetch()` to resolve the full node (store-backed).
- Calling against a pre-1.10 server raises a clear `VersionNotSupportedError` instead of an opaque GraphQL failure.

## Implementation notes

- New `infrahub_sdk/graph_traversal/` package: typed pydantic result models (`models.py`) and GraphQL query strings + input builders (`query.py`). Pure logic is isolated so it's unit-tested without HTTP.
- New `VersionNotSupportedError` exception.
- Added to docs generation (`tasks.py`), a how-to guide, generated SDK reference, and a changelog fragment.

## Testing

- `pytest tests/unit/sdk/test_graph_traversal.py` — 23 passed (pure-logic builders/parsing + client methods mocked at the transport boundary, async + sync).
- `ruff check` clean; `lint-docs` 0 errors.

---
## Summary by cubic
Adds graph traversal to the Python SDK for Infrahub 1.10+, enabling shortest-path and dependency discovery across any relationships. Supports INFP-531 (Phase 1) by exposing traversal APIs the UI and GraphQL layer can use.

- **New Features**
  - SDK: `InfrahubClient.traverse_paths()`, `reachable_nodes()`, and `path_exists()` (with sync equivalents) to find shortest paths, discover reachable nodes by kind, and check connectivity.
  - Ergonomics: accept node IDs or `InfrahubNode`; kind filters accept strings or protocol classes; results include `PathNode.fetch()` to resolve full nodes from the client store.
  - Compatibility: pre-1.10 servers raise `VersionNotSupportedError` with a clear message.

- **Bug Fixes**
  - Clear errors for unsaved nodes used as traversal inputs (`NodeNotSavedError`, wrapped with role-specific context).
  - Safer schema kind resolution when `save` is missing in `_get_schema_name`.
  - More robust related-node detection using `isinstance(InfrahubNodeBase)` to avoid false positives.
  - Docs: added full Args in traversal docstrings; guide clarifies default-excluded kinds and how to detect truncated results; wording tweak to satisfy Vale (avoid “UUIDs”).
  - Include GraphQL `operationName` in requests for better tracing.
  - Tests: register traversal annotations in the async/sync parity map.

<sup>Written for commit 1fbfe9b49682d7645744e0cb08eb9c9c9bf6c977. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1090?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


docs/docs/python-sdk/guides/graph_traversal.mdx
```diff
@@ -0,0 +1,330 @@
...
+
+  print(f"{result.count} path(s) found")
+  ```
+
+  </TabItem>
+</Tabs>
+
+### Constraining the traversal
+
+All limits and filters are optional; when omitted, the server applies its own defaults. Unknown kinds (for example IP namespaces) are excluded by default and reported back in `result.excluded_kinds`.
```
ajtmccarty: I don't know exactly what "Unknown kinds" means here. Any IP namespaces are excluded by default b/c they act as hubs and are generally not valuable in this kind of path tracing (although they can be re-included using `included_kinds`). other internal namespaces are also excluded by default: Builtin, Internal, Core, etc.

docs/docs/python-sdk/guides/graph_traversal.mdx
```diff
@@ -0,0 +1,330 @@
...
+
+:::note
+
+`.fetch()` issues one request per node. For large results where you need many full nodes, prefer fetching only the ones you actually need, or batch them with `client.get` inside an `InfrahubBatch`.
+
+:::
+
+### Detecting truncated results
+
+The server caps the number of paths/results. Compare `count` to your requested limit to know whether more may exist:
```
ajtmccarty: the `count` is actually always the number of paths returned for either of the path traversal queries. it won't be a reliable method for determining if the `max_targets` was reached. that would require checking the UUIDs of the start objects for each path

docs/docs/python-sdk/sdk_ref/infrahub_sdk/client.mdx
```diff
@@ -119,6 +119,64 @@ count(self, kind: str | type[SchemaType], at: Timestamp | None = None, branch: s
...
+```python
+reachable_nodes(self, source: str | InfrahubNode, target_kinds: list[str | type[SchemaType]]) -> ReachableNodesResult
+```
+
+Find all nodes of the given kinds reachable from a source node.
+
+``source`` accepts either a node UUID string or an ``InfrahubNode`` instance.
+``target_kinds`` accepts kind-name strings and/or generated protocol classes.
+
+Requires Infrahub 1.10 or later. See https://docs.infrahub.app for details.
```
ajtmccarty: is this docs link actually helpful?

infrahub_sdk/client.py
```diff
@@ -117,6 +126,29 @@ def get_kind_as_string(kind: str | type[SchemaType | SchemaTypeSync]) -> str:
...
+    """Return a node UUID from either an id string or an InfrahubNode instance.
+
+    Raises:
+        Error: If a node instance without an id is provided.
+
+    """
+    if isinstance(node, str):
+        return node
+    if node.id is None:
+        raise Error("Cannot use a node without an id as a traversal source or destination.")
```
ajtmccarty: the function name is pretty generic but the error message is specific to path traversal. I think this method should raise some sort of `NodeNotSaved` error and then the path traversal caller can catch it and raise an error that mentions path traversal if necessary

ajtmccarty (approved):
looks good. few small comments

infrahub_sdk/client.py
minitriga: Fixed in b7e2b18: `_resolve_node_id` now raises a generic `NodeNotSavedError`, and the traversal callers wrap it via `_resolve_traversal_node_id(..., role=...)` to add path-traversal context.

docs/docs/python-sdk/guides/graph_traversal.mdx
minitriga: Fixed in b7e2b18: reworded to explain IP namespaces are excluded by default (they act as hubs), along with internal namespaces (Builtin, Internal, Core), reported in `excluded_kinds` and re-includable via `included_kinds`.

docs/docs/python-sdk/guides/graph_traversal.mdx
minitriga: Fixed in b7e2b18: clarified that `count` is the number of paths, not distinct nodes. The example now counts distinct node UUIDs to detect whether `max_results` was reached.

docs/docs/python-sdk/sdk_ref/infrahub_sdk/client.mdx
minitriga: Fixed in b7e2b18: removed the generic docs.infrahub.app link from the traversal docstrings.

docs/docs/python-sdk/sdk_ref/infrahub_sdk/client.mdx
```diff
@@ -119,6 +119,64 @@ count(self, kind: str | type[SchemaType], at: Timestamp | None = None, branch: s
```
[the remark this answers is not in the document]
minitriga: Fixed in b7e2b18: added full `Args:` sections to the traversal docstrings and regenerated the reference, so all keyword-only arguments are now documented (mdxify omits post-`*` params from the signature line).

infrahub_sdk/node/related_node.py
```diff
@@ -58,9 +58,11 @@ def __init__(self, branch: str, schema: RelationshipSchemaAPI, data: Any | dict,
...
         self._peer_has_been_mutated: bool = False
 
-        # Check for InfrahubNodeBase instances using duck-typing (_schema attribute)
-        # to avoid circular imports, or CoreNodeBase instances
-        if isinstance(data, CoreNodeBase) or hasattr(data, "_schema"):
+        # Detect node instances. InfrahubNodeBase is imported lazily here to avoid a
+        # circular import (node.py imports this module at load time).
+        from .node import InfrahubNodeBase as _InfrahubNodeBase  # noqa: PLC0415
+
+        if isinstance(data, (CoreNodeBase, _InfrahubNodeBase)):
```
ajtmccarty, line 65: this indicates a deeper structural problem in the SDK, but not really something to be solved in this PR. we might use some sort of duck-typing workaround in other places to avoid this mid-function import. something like if `hasattr(data,  "_schema")`, which might or might not work here

minitriga: Yeah I dont think it can be helped here. I'll open an issue to track this.

minitriga: Tracked in #1092 for a proper structural fix (the node.py ↔ related_node.py circular dependency). Keeping the isolated lazy import here for now.

ajtmccarty (approved):

ogenstad:
A note on target branch here, whenever we are targeting a feature for an upcoming Infrahub release we typically target the PR to the `infrahub-develop` branch. The idea is that if needed we can merge the `develop` branch into `stable` at any time to create a patch release of the SDK and when doing so we don't want to include anything from the next Infrahub release. As it is now we should be getting close regardless so `develop` might be fine.

PR #1108, opened 2026-06-28
feat(sdk): client.get_many — batched multi-kind node fetch

iddocohen:
## Summary

Add `InfrahubClient.get_many` (and the sync twin): one GraphQL
operation, one aliased block per kind, one round-trip.

```python
result = await client.get_many({
    "InfraAutonomousSystem": {"ids": [as_id_1, as_id_2], "attributes": ["asn"]},
    "InfraDevice":           {"ids": [dev_id_1],          "attributes": ["role"]},
})
# result["InfraDevice"] -> list[InfrahubNode]
```

Today this shape costs N round-trips with per-kind `client.filters` (each pulling the full default selection set), or a hand-rolled `execute_graphql` with manual alias plumbing and a hand-written hydration loop. After this change, it's a single call.

## Why this is needed

The pattern — read attributes off a heterogeneous set of `(kind, id)` tuples in one shot — recurs across the Infrahub ecosystem. Concrete domains that will adopt `get_many` immediately:

| Domain | Where the pattern shows up |
| --- | --- |
| **Generators** (`infrahub-managing-generators` skill, opsmill demos) | Design-driven generators read a topology spec, the rack layout, IPAM allocations, vendor catalog — all heterogeneous, all needed before any synthesis. Today: a flurry of `client.get` / `client.filters` calls. |
| **Python transforms** | Anything beyond ""render this one node"" needs to stitch attributes from several kinds. Today the workaround is either many round-trips or a hand-rolled `execute_graphql`. |
| **Checks** (other than reachability) | Impact-assessment / blast-radius checks fan out from a changed node to its dependents across kinds; verdicts depend on attributes from each. Same compile-then-execute shape. |
| **infrahub-sync / terraform-provider-generator / infrahub-ansible** | Bulk export / inventory build today does per-kind passes. `get_many` collapses a multi-kind sweep to one call. |
| **Change-impact / branch-diff reporting** | The diff API returns `(kind, id)` tuples for changed nodes. To render a human report you need attributes per id — current options are N gets or hand-built GraphQL. |
| **CLI tooling, dashboards, ad-hoc scripts** | Any ""give me a summary across these entities"" command is the same shape. |
| **Migrations / one-shot data fixes** | ""For these 10k nodes across 5 kinds, read attribute X, mutate Y."" Per-kind today; batched after. |

The litmus test for adoption is simple: search the opsmill org for `execute_graphql(` and look at what those queries do. Whenever the query is a hand-built alias-block multi-fetch (rather than a genuinely custom GraphQL operation the SDK can't express), that's a future caller of `get_many`. The canonical example today is [`_fetch_attributes` in opsmill/infrahub-demo-reachability-check](https://github.com/opsmill/infrahub-demo-reachability-check/blob/main/checks/path_assertion.py#L403-L485) — that method drops from ~70 lines of query construction + hydration to ~10 lines of domain logic once `get_many` lands.

## What ships

- `infrahub_sdk.client.compile_get_many_query` — pure helper that turns a kind→{ids, attributes} spec into `(query, variables, kinds_ordered)`. Validates kind and attribute names as GraphQL identifiers up front and collects every problem before raising `ValidationError`.
- `InfrahubClient.get_many` / `InfrahubClientSync.get_many` — async and sync. Reuse `execute_graphql` for transport, `InfrahubNode.from_graphql` for hydration, and `client.store.set` for store population. `populate_store=True` matches `client.get` semantics.

Reuses existing primitives end-to-end. No new exception classes, no new result types, no new dependencies. The return is `dict[str, list[InfrahubNode]]`.

## Error handling

Compile-time problems (empty spec, missing/empty `ids`, malformed kind/attribute identifiers, scalar `str`/`bytes` passed where a list is expected) raise `ValidationError` with a per-problem message list **before any HTTP call is made**. Server-side rejections (unknown kind or attribute on the loaded schema) propagate as `GraphQLError` from `execute_graphql`.

## Back-compat

Additive only. `get`, `filters`, `all`, `traverse_paths`, `execute_graphql`, and `PathNode.fetch()` are untouched.

## Test plan

- [x] Pure-function tests over `compile_get_many_query`: query shape, alias allocation, variable allocation, id/attribute dedup + sort, kind-order preservation, every compile-error path, scalar `str`/`bytes` rejection for `ids` and `attributes`
- [x] Round-trip tests via `httpx_mock` parametrized over `[""standard"", ""sync""]`: hydration, `populate_store=True/False`, compile errors short-circuit before the HTTP call
- [x] Full SDK unit suite: 1030+ tests passing
- [x] `uv run invoke format lint-code` clean (ruff, ty, mypy)
- [x] Integration tests (`integration-tests-latest-infrahub`) passing in CI

## Related issues

- [opsmill/infrahub#9067](https://github.com/opsmill/infrahub/issues/9067) — the same N+1 problem on the frontend (one GraphQL query per related node on the task details page). Not directly fixed by this PR (frontend is TypeScript), but the design here is the blueprint for the TypeScript port.

## Follow-ups (not in this PR)

- `PathTraversalResult.fetch_hop_attributes(spec)` sugar — one-liner over `get_many` for the reachability-check shape.
- `include_attributes=` keyword on `traverse_paths` — sugar that calls the above before returning.

### DIFF
--- a/docs/docs/python-sdk/sdk_ref/infrahub_sdk/client.mdx
+++ b/docs/docs/python-sdk/sdk_ref/infrahub_sdk/client.mdx
@@ -216,6 +216,41 @@ Requires Infrahub 1.10 or later.
 - `VersionNotSupportedError`: If the server does not support graph traversal (pre-1.10).
 - `GraphQLError`: When the GraphQL response contains errors (e.g. unknown node).
 
+#### `get_many`
+
+```python
+get_many(self, spec: Mapping[str, Mapping[str, Sequence[str]]]) -> dict[str, list[InfrahubNode]]
+```
+
+Fetch nodes of multiple kinds in a single GraphQL operation.
+
+``spec`` maps each kind to a dict with a non-empty ``ids`` list (node UUIDs)
+and an optional ``attributes`` list (attribute names to populate on the
+returned nodes). One aliased subselection is emitted per kind, so the cost
+is a single round-trip regardless of how many kinds the spec contains.
+
+Returned nodes only have the requested attributes populated; relationships
+are not fetched. For full nodes, follow up with :meth:`get` or :meth:`filters`.
+
+**Args:**
+
+- `spec`: Mapping of kind name to a sub-mapping with keys ``ids`` (list of
+node UUIDs, required and non-empty) and ``attributes`` (list of
+attribute names, optional - defaults to none, in which case only
+the id and typename are returned).
+- `branch`: Name of the branch to query from. Defaults to ``default_branch``.
+- `at`: Time of the query. Defaults to now.
+- `timeout`: Overrides the default GraphQL timeout, in seconds.
+- `populate_store`: When True (default), every hydrated node is added to
+``client.store`` so later lookups by id can skip the network.
+
+**Raises:**
+
+- `ValidationError`: If the spec cannot be compiled into a query (empty,
+missing ``ids``, malformed kind/attribute identifiers).
+- `GraphQLError`: If the server rejects the query (for example, an unknown
+kind or attribute name in the loaded schema).
+
 #### `all`
 
 ```python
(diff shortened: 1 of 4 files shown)


ogenstad (commented):
As mentioned in Slack I think we should have an extra round of design for this before moving forward.

PR #1135, opened 2026-07-04
feat: generate user-facing write/read schema models + offline validation

dgarros:
## Summary
Adds generated, user-facing **write** and **read** schema models to the SDK plus an offline `validate_schema()` so a schema can be checked for correctness with no running Infrahub server. Part of the user-facing schema separation (opsmill/infrahub, INFP-234).

## Key Changes
- New generated models `infrahub_sdk/schema/generated/{write,read}.py` — the write model is exactly what `/api/schema/load` accepts (constrained fields carry their allowed values as `Literal[...]`; `extra="forbid"`); read = write + visible-but-not-settable fields.
- `validate_schema()` (`infrahub_sdk/schema/validate.py`) — field-level offline validation with dotted error locations; also gates schema `extensions`.
- Drift/presence guard test so the generated models can't silently go stale.
- Regenerated `protocols.py` (pre-existing drift vs current backend schema).

## Related Context
Generated from the backend's schema definitions; consumed by the backend PR (opsmill/infrahub). Depends on nothing else in this repo.

## Test Plan
```bash
uv run pytest tests/unit/test_schema_offline_validation.py tests/unit/test_schema_generated_models.py
```
14 offline-validation cases + generated-model guard pass, pydantic-only (no server).

## Notes
Draft: the backend still ships a parallel hand-written model set; full consolidation of the SDK's hand-written schema models is tracked as a follow-up.

---
## Summary by cubic
Adds generated user-facing write/read schema models and an offline validator so schemas can be checked without a server. This separates user schema from internal models and satisfies INFP-234; validation covers extensions and now requires the root version.

- New Features
  - Committed generated models in `infrahub_sdk/schema/generated/{enums,write,read}.py`; export `InfrahubSchema{Write,Read}` and enums; read adds `Profile`/`Template`; write includes `extensions`; `extra="ignore"` with `use_enum_values=True`.
  - Attributes and computed attributes are kind‑discriminated unions; typed `parameters` and `choices`; enums live in `generated/enums.py`.
  - `validate_schema()` checks the full document (`nodes`, `generics`, `extensions`) via `InfrahubSchemaWrite`, returns `SchemaValidationResult` with dotted paths; drops non‑write/unknown fields; includes a generated‑model drift guard; requires `version`.

- Refactors
  - Public SDK schema classes now wrap the generated models; `*SchemaAPI` inherit generated read models; enums re‑exported; `CoreNodeBase._schema` is `MainSchemaTypesAPI`.
  - Read‑only `kind`/`hash` moved to generated read models; write nodes expose a non‑serializing `.kind`; `SchemaRoot` exposes `extensions` and validates payloads via `InfrahubSchemaWrite`.
  - Relationship helpers inlined on `RelationshipSchemaAPI`; `AttributeSchema` construction is lenient; removed `AttributeKind.STRING`; defaults match the server (relationship `min_count`/`max_count`=0; node `branch="aware"`, `generate_profile`=True, `generate_template`=False); regenerated `protocols.py`; skip `IPAddress` kind tests pending backend support.

<sup>Written for commit 973eaa909635f2f46aed7b495aec85b2fe40eb82. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1135?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


infrahub_sdk/schema/generated/write.py
```diff
@@ -0,0 +1,583 @@
...
+        description="Mark attribute as deprecated and provide a user-friendly message to display",
+        max_length=128,
+    )
+    display: SchemaAttributeDisplay = Field(
+        default=SchemaAttributeDisplay.DEFAULT,
+        description="Controls where the attribute is displayed. 'default' shows in the main view, 'extra' shows in an expanded/secondary section.",
+    )
+
+
+class TextAttributeWrite(AttributeSchemaBaseWrite):
```
ajtmccarty, line 235: can these not inherit from their corresponding `Read` model b/c `extra="forbid"` on the write ones?

infrahub_sdk/schema/main.py
```diff
@@ -1,154 +1,146 @@
...
+
+
+class _SchemaKindMixin:
+    """Capability flags shared by node/generic/profile/template read schemas.
+
+    ``kind`` itself is a computed field on the generated read base model; this mixin only carries
+    the derived capability helpers and declares ``kind`` for the type checker.
+    """
+
+    if TYPE_CHECKING:
```
ajtmccarty: I'm surprised this is necessary. if `_SchemaKindMixin` is only used in classes inheriting from `BaseNodeReadSchema` / `BaseNodeWriteSchema`, then I would think that `name`, `namespace`, etc. fields on the `BaseNode...` class would cover this case.
A Mixin with a `TYPE_CHECKING` conditional inside of it feels like a kludge. is there some other way to handle this?

tests/unit/test_schema_offline_validation.py
```diff
@@ -0,0 +1,413 @@
...
+    schema = _valid_schema()
+    # `inherited` is a read-level attribute field; a user must not be able to set it.
+    schema["nodes"][0]["attributes"][0]["inherited"] = True
+
+    result = validate_schema(schema=schema)
+
+    assert result.valid is False
+    assert any("inherited" in message for message in result.messages), result.messages
+    # The message must locate the offending field within the payload.
+    assert any("nodes[0].attributes[0]" in message for message in result.messages), result.messages
```
ajtmccarty: I think these error assertions could be a little more restrictive. I think there should only be 1 error and it should be possible to assert more, or all, of its text
same for other error message assertions in this test file

tests/unit/test_schema_offline_validation.py
```diff
@@ -0,0 +1,413 @@
...
+    assert result.valid is False
+    # `kind` is the attribute union's discriminator, so an unknown kind is reported against the
+    # attribute itself, with the discriminator field and the invalid value named in the message.
+    assert "nodes[0].attributes[0]" in _fields_named(result), result.messages
+    assert any("kind" in message and "NotARealKind" in message for message in result.messages), result.messages
+
+
+def test_enum_backed_relationship_cardinality_valid_value_passes() -> None:
+    # cardinality is typed with the RelationshipCardinality enum (use_enum_values keeps the runtime
+    # value a plain string); a valid enum value must still validate.
```
ajtmccarty: `# test string for RelationshipCardinality is valid`

tests/unit/test_schema_offline_validation.py
```diff
@@ -0,0 +1,413 @@
...
+    schema = _valid_schema()
+    schema["nodes"][0]["not_a_field"] = "boom"
+
+    result = validate_schema(schema=schema)
+
+    assert result.valid is False
+    assert any("not_a_field" in message for message in result.messages), result.messages
+
+
+def test_raise_on_error_raises_value_error_naming_field() -> None:
```
ajtmccarty: looks very similar to `test_out_of_enum_value_is_rejected_naming_field_and_value`

ajtmccarty (commented):
the main outstanding issue for me is finding some way to create the `SchemaAPI` classes without `Mixins` that have `TYPE_CHECKING` conditionals in them. there are also some comments from cubic that might need addressing

infrahub_sdk/schema/generated/write.py
dgarros: I thought about it, but I think it's making the generator a bit more complicated to have the write inherit from the read, since everything is generated. It's as easy to just generate them twice.

infrahub_sdk/schema/main.py
dgarros: Hey — I was surprised by this too, so I had Claude dig into it. Sharing what it found, because I actually think the current form is the right call.

**Why the block is there:** at runtime this is fine — every class the mixin lands on (`NodeSchemaAPI`, `GenericSchemaAPI`, `ProfileSchemaAPI`, `TemplateSchemaAPI`) inherits from a generated read model that extends `BaseNodeSchemaRead`, where `name` / `namespace` / `kind` / `attributes` / … actually live, so MRO resolves everything. The issue is purely static: the type checker analyzes `_SchemaKindMixin` **in isolation at its definition**, where it has no base that provides those attributes, so `self.kind` etc. come back "unresolved." The `TYPE_CHECKING` block is just the contract — "my host supplies these" — and emits zero runtime code. It's the documented idiom for typing mixins whose attributes come from the eventual host class.

(Detail: `kind` is declared as a `@property`, not `kind: str`, because on the generated base it's a read-only `@computed_field` — a plain annotation would model it as a settable field and clash in the MRO.)

**Why not just rely on the fields from `BaseNodeSchemaRead`:** that only works at runtime. The whole point of a mixin here is to share these helpers across four *distinct* generated read classes; the mixin has no single base to inherit the fields from at its own definition site, which is exactly what the checker looks at.

**Why the alternatives are worse:**

- *Make each mixin subclass its generated base* (so fields are inherited): this removes the block, but turns the plain, field-less mixins into full pydantic models with diamond inheritance, and couples each mixin to one specific generated base — losing the "compose onto any of them" property that motivated the mixin. It also drags back the field-collection subtleties (and the `kind` computed-field clash) we deliberately kept out. Net negative.
- *Protocol-typed `self` on every method*: doesn't actually remove the contract, just relocates it — and now every helper (including the `@property` getters) carries a `self: SomeProtocol` annotation. More noise, not less.
- *Generate the helpers into the read models*: this behavior is SDK-specific (e.g. `kind == "CoreArtifactDefinition"`), not derivable from the schema, so it would leak hand-written logic into files that are supposed to be pure generated schema.

So the `TYPE_CHECKING` block is the least-intrusive option and a recognized pattern rather than a one-off hack — I'd like to keep it as-is.

ajtmccarty: This sounds like a contract without any verification. the `_SchemaKindMixin` class seems like it indicates that any class using this Mixin needs to define `name`, `namespace`, `inherit_from`, and `kind` instance variables or properties, but I don't think there is any checking that actually verifies this. for example if `NodeSchemaAPI` does not define `name`, then I don't think it would be identified by type-checking. if I am wrong about that, then I will approve this pattern, but I can't approve something that has 3 strikes against it
- it's a Mixin, which I consider a code smell
- it uses TYPE_CHECKING, which I consider a crutch that hides import errors regardless of where it is used
- it isn't actually type checked

I think the right approach is to make `_SchemaKindMixin` and `_SchemaRelAttrMixin` one base class for the `Node/Generic/Profile/TemplateSchemaAPI` classes, even if it uses "diamond inheritance"

infrahub_sdk/schema/main.py
```diff
@@ -1,157 +1,155 @@
...
+class RelationshipSchema(RelationshipSchemaWrite):
+    """Constructible relationship write model (kept as a distinct public name)."""
+
 
-    inherited: bool = False
-    read_only: bool = False
-    allow_override: AllowOverrideType = AllowOverrideType.ANY
+class NodeSchema(NodeSchemaWrite):
+    # ``attributes`` accepts the constructible ``AttributeSchema`` (the generated discriminated union
+    # cannot be built from an ``AttributeSchema`` instance); dumping still validates server-side.
```
ajtmccarty: I'm not really sure what this means

infrahub_sdk/schema/generated/read.py
```diff
@@ -0,0 +1,602 @@
...
+        default=None,
+        description="List of values or range of values not allowed for the attribute, format is: '100,150-200,280,300-400'",
+        pattern=r"^(\d+(?:-\d+)?)(?:,\d+(?:-\d+)?)*$",
+    )
+
+
+class NumberPoolParametersRead(AttributeParametersRead):
+    model_config = ConfigDict(extra="ignore", use_enum_values=True)
+    end_range: int = Field(
+        default=9223372036854775807,
```
ajtmccarty, line 71: should this be a constant somewhere? maybe with a comment to explain what this number is

infrahub_sdk/schema/main.py
```diff
@@ -161,12 +159,18 @@ def cardinality_is_many(self) -> bool:
...
 
-class BaseSchemaAttrRel(BaseModel):
-    attributes: list[AttributeSchema] = Field(default_factory=list)
-    relationships: list[RelationshipSchema] = Field(default_factory=list)
+class _SchemaNodeBase(BaseNodeSchemaRead):
+    """Behavior shared by the node/generic/profile/template read models.
 
+    Subclasses ``BaseNodeSchemaRead`` so ``name``, ``namespace``, ``kind`` and the attribute/
+    relationship collections are real inherited fields — the helpers below are type-checked against
+    them, unlike a mixin whose fields would only be declared for the type checker. The node-like
```
ajtmccarty: you don't even have to mention `mixin`s here

tests/unit/test_schema_generated_models.py
```diff
@@ -0,0 +1,240 @@
...
+    assert [member.value for member in enum_cls] == _EXPECTED_ENUM_VALUES[enum_name]
+
+
+def test_attribute_kind_enum_present_without_deprecated_string_member() -> None:
+    assert issubclass(enums_module.AttributeKind, Enum)
+    assert issubclass(enums_module.AttributeKind, str)
+    values = [member.value for member in enums_module.AttributeKind]
+    # The deprecated "String" kind is dropped from the generated enum.
+    assert "String" not in values
+    assert "Text" in values
```
ajtmccarty, line 201: not sure this really needs a test

ajtmccarty (approved):

infrahub_sdk/schema/main.py
dgarros: Reworded in 973eaa9 — it now says only why the override exists: the generated write model types these as discriminated unions, which cannot be instantiated directly, so overriding them with the public constructible models keeps `NodeSchema(attributes=[AttributeSchema(...)])` working. Dropped the trailing remark about server-side validation, which was unrelated to the override.

infrahub_sdk/schema/main.py
dgarros: Dropped in 973eaa9 — the docstring no longer mentions mixins, since the mixin it replaced is gone.

tests/unit/test_schema_offline_validation.py
dgarros: Compressed to your one-liner in 973eaa9.

PR #1149, opened 2026-07-08
feat: retry HTTP 429 responses with jittered backoff

### LINKED ISSUE #1124: feature: retry requests with backoff on HTTP 429 responses
### Component

Python SDK

When the Infrahub server responds with an HTTP 429 (Too Many Requests), the
SDK client should automatically retry the request instead of surfacing the
error immediately.

Two behaviours are needed:

- **Exponential backoff with jitter** — on a 429, wait before retrying and
  progressively increase the wait on each subsequent failure, with a small
  random jitter added to avoid a thundering-herd of clients retrying in
  lockstep.
- **Respect the `Retry-After` header** — when the 429 response includes a
  `Retry-After` header, honour the delay it specifies rather than the
  computed backoff. This is the server communicating its own recovery time.

Both the async (`InfrahubClient`) and sync (`InfrahubClientSync`) clients
should share this behaviour.

### Describe the Use Case

Clients (including `infrahubctl` and the Infrahub Ansible collection, which
use this SDK internally) can generate bursts of requests — batching, syncs,
generators — that trip server-side rate limiting. Today a 429 bubbles up as
a failure, forcing every caller to implement their own retry logic. Handling
it in the SDK makes rate limiting transparent and lets the server reliably
shed and recover from load.

### Additional Information

Suggested implementation notes (non-binding):

- Cap the number of retries and the maximum backoff so requests eventually
  fail rather than hang indefinitely.
- Consider making retry parameters configurable via the existing pydantic
  `Config`.

dgarros:
# Why

When the Infrahub server returns HTTP 429 (Too Many Requests), the SDK surfaces the error immediately — so every caller (`infrahubctl`, the Ansible collection, custom scripts) has to hand-roll retry logic or just fail. Background workloads (batching, syncs, generators) are the traffic most likely to be rate-limited and had no cooperative way to back off.

**Goal:** the SDK transparently retries 429s with jittered exponential backoff, honours the server's `Retry-After`, and raises a dedicated catchable error once a configurable budget is exhausted — so callers ride through transient rate-limiting with no retry code of their own.

**Non-goals:** retrying other status codes (503, etc.); server-side rate limiting / prioritisation (INFP-636/635); changing the existing connectivity `retry_on_failure` mechanism.

Closes #1124 (Jira IHS-249)

## What changed

**Behavioural:**

- 429 responses are retried automatically (default 5 retries) with jittered exponential backoff — callers receive the eventual success transparently.
- The server's `Retry-After` header (delta-seconds **and** HTTP-date) is honoured in place of computed backoff, clamped to a max; malformed headers fall back to computed backoff.
- On sustained rate-limiting the SDK raises a new `RateLimitError` (carrying `url`, `attempts`, last `Retry-After`; chains the underlying `httpx.HTTPStatusError` as `__cause__`).
- Identical behaviour across `InfrahubClient` (async) and `InfrahubClientSync` (sync), applied to **all** request paths: queries, mutations, multipart uploads, streaming initiation, and auth.
- Each retry logs at WARNING (url, attempt, delay) — no headers/credentials logged.

**Config (all additive):** `rate_limit_retry_enabled` (True), `rate_limit_max_retries` (5), `rate_limit_backoff_base` (0.5), `rate_limit_backoff_max` (60). Set `rate_limit_retry_enabled=False` to fully restore the previous behaviour.

**Implementation notes:** a new pure-logic `RateLimitRetryHandler` (no I/O) owns parsing/backoff/decision; thin async & sync drivers wrap the three send sites. Multipart retries rewind/re-materialize the body per attempt.

**What stayed the same:** no existing method signatures changed; `retry_on_failure` untouched; no new dependencies (httpx + stdlib only).

## How to review

Suggested order: `infrahub_sdk/rate_limit.py` (pure logic) → `exceptions.py` + `config.py` → `infrahub_sdk/client.py` (drivers + three wired send sites, async & sync) → tests. Extra scrutiny on the async/sync driver parity and the streaming-init retry resource handling. Full design rationale in `dev/specs/ihs-249-sdk-429-retry/`.

## How to test

```bash
uv run pytest tests/unit/test_rate_limit.py tests/unit/sdk/test_rate_limit_retry.py
```

Expected: **62 passing** — handler unit tests + client-level tests parametrized across async/sync (transparent 429→200, `Retry-After` honouring, exhaustion→`RateLimitError`, disable/tune, all-paths coverage, multipart body-survives-retry, jitter growth/divergence).

## Impact & rollout

- **Backward compatibility:** one caller-visible change — a persistent 429 now raises `RateLimitError` after retries exhaust instead of `httpx.HTTPStatusError` (raw error still available via `__cause__`). Documented in `changelog/1124.changed.md`. Retry is on by default; `rate_limit_retry_enabled=False` opts out.
- **Performance:** no hot-path impact; worst-case blocking is bounded (~`max_retries × backoff_max` ≈ 300s with defaults).
- **Config/env changes:** four additive `rate_limit_*` Config fields (env-overridable).
- **Deployment notes:** safe to deploy; additive.

## Checklist

- [x] Tests added/updated (62 tests)
- [x] Changelog entry added (`1124.added.md`, `1124.changed.md`)
- [x] External docs updated (`config.mdx` regenerated)
- [x] Internal .md docs updated (`dev/specs/ihs-249-sdk-429-retry/`)

---
## Summary by cubic
Adds transparent retries for HTTP 429 responses with jittered exponential backoff across both clients, honoring `Retry-After`, and raising `RateLimitError` when the retry budget is exhausted. Encapsulates the retry loop inside `RateLimitRetryHandler` with a single handler instance per client, and hardens exhaustion and streaming edge cases.

- **New Features**
  - Automatically retries 429s with jittered exponential backoff; honors `Retry-After` (delta-seconds or HTTP-date) and clamps to a max.
  - Works across `InfrahubClient` and `InfrahubClientSync` for all paths: standard requests, multipart uploads (with safe body rewind), streaming initiation, and auth.
  - New `RateLimitError` is raised after retries are exhausted and chains the underlying `httpx.HTTPStatusError` when available.
  - Configurable via `rate_limit_retry_enabled` (True), `rate_limit_max_retries` (5), `rate_limit_backoff_base` (0.5), `rate_limit_backoff_max` (60). Hardened: caps backoff exponent, guards negative/overflow `Retry-After`, always raises on request-less 429 exhaustion, and closes streaming resources on failure. Internals: driver loop moved into `RateLimitRetryHandler.send/asend`; each client reuses a single handler.

- **Migration**
  - Persistent 429s now raise `RateLimitError` instead of `httpx.HTTPStatusError`. Catch `RateLimitError` (or the SDK base `Error`); the original `httpx.HTTPStatusError` is available via `__cause__` when present.
  - To revert to previous behavior, set `rate_limit_retry_enabled=False`.

<sup>Written for commit 07cda1ab85c4b527647ba78fe6b752680101bcae. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1149?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


infrahub_sdk/client.py
```diff
@@ -1472,16 +1496,92 @@ async def _get_streaming(
...
+
+        """
+        if not self.config.rate_limit_retry_enabled:
+            return await send()
+
+        handler = RateLimitRetryHandler(
+            max_retries=self.config.rate_limit_max_retries,
+            backoff_base=self.config.rate_limit_backoff_base,
+            backoff_max=self.config.rate_limit_backoff_max,
+        )
```
saltas888: I was thinking whether we can actually have the attempts state inside the class so that we don't always need to keep track of the attempts when using. However it seems thats intentionally designed to be stateless.

If we prefer it to be stateless, we should initialize the handler once globally. We do it now twice

.specify/feature.json
ajtmccarty, line 1: leftover

changelog/1124.added.md
ajtmccarty, line 1: should probably only be one of these changelogs

CLAUDE.md
```diff
@@ -2,5 +2,6 @@
 
 <!-- SPECKIT START -->
 For additional context about technologies to be used, project structure,
-shell commands, and other important information, read the current plan
+shell commands, and other important information, read the current plan:
+`specs/ihs-249-sdk-429-retry/plan.md`
```
ajtmccarty: leftover

.specify/feature.json
gmazoyer, line 1: This should be in the `.gitignore` file

CLAUDE.md
gmazoyer: This should not be part of the PR

infrahub_sdk/rate_limit.py
```diff
@@ -0,0 +1,93 @@
...
+
+        # delta-seconds form
+        try:
+            return max(0.0, float(int(value)))
+        except OverflowError:
+            # A pathological, arbitrarily long digit string overflows float(); fall back to
+            # computed backoff rather than crashing.
+            return None
+        except ValueError:
+            # Not a delta-seconds integer (e.g. an HTTP-date); fall through to date parsing.
```
gmazoyer: Those comments are not needed, error handling is pretty explicit by itself "A pathological, arbitrarily long digit string overflows" that's a very convoluted way for the bot to explain a simple concept :D

infrahub_sdk/client.py
```diff
@@ -1472,16 +1496,92 @@ async def _get_streaming(
         base_headers = copy.copy(self.headers or {})
         headers.update(base_headers)
 
+        request_timeout = timeout or self.default_timeout
         async with httpx.AsyncClient(**self._build_proxy_config(), verify=self.config.tls_context) as client:
+            open_stream: dict[str, AsyncExitStack] = {}
```
ajtmccarty, line 1503: does this need to be a dictionary? it just holds one variable. seems like it could be `stack: AsyncExitStack`

infrahub_sdk/client.py
ajtmccarty: I would also prefer to see all of this logic moved into `RateLimitRetryHandler`, but I believe that might not be easy to achieve b/c we have to support both sync and async. It'd be nice if there was some way to do accomplish this b/c all of this should be encapsulated in `await retry_handler.send()` or something

infrahub_sdk/client.py
```diff
@@ -1472,16 +1496,92 @@ async def _get_streaming(
...
+                        await stack.aclose()
             except httpx.NetworkError as exc:
                 raise ServerNotReachableError(address=self.address) from exc
             except httpx.ReadTimeout as exc:
-                raise ServerNotResponsiveError(url=url, timeout=timeout or self.default_timeout) from exc
+                raise ServerNotResponsiveError(url=url, timeout=request_timeout) from exc
+
+    async def _send_with_rate_limit_retry(
+        self,
+        send: Callable[[], Coroutine[Any, Any, httpx.Response]],
```
ajtmccarty: I find the callback approach to be difficult to read, hard to debug, and hard to test.
I think an interface like this 
```
class RequestSenderInterface:
    async def send(self) -> httpx.Response:
        ...
```
that could be passed into some version of `RateLimitRetryHandler.send()` would be a more SOLID approach and would allow easier, more modular testing of all the pieces (individual Sender instances, `RateLimitRetryHandler`, etc.). but this might be difficult with the current structure of the client and supporting both sync and async

infrahub_sdk/rate_limit.py
```diff
@@ -0,0 +1,93 @@
...
+        return max(0.0, delta)
+
+    def compute_backoff(self, attempt: int) -> float:
+        """Deterministic exponential ceiling: ``min(backoff_max, backoff_base * 2**attempt)``.
+
+        ``attempt`` is capped before exponentiation: ``2**attempt`` for a large ``attempt``
+        (reachable via a very high ``rate_limit_max_retries``) would overflow the float
+        conversion in the multiply before the ``min`` clamp could run. ``2**63`` already dwarfs
+        any realistic ``backoff_max`` (seconds), so capping there always yields the clamp.
+        """
```
ajtmccarty: a lot of docstring for little value

infrahub_sdk/rate_limit.py
```diff
@@ -0,0 +1,93 @@
...
+        return min(self.backoff_max, self.backoff_base * (2 ** min(attempt, 63)))
+
+    def jittered_delay(self, ceiling: float) -> float:
+        """Full jitter: ``random.uniform(0, ceiling)``."""
```
ajtmccarty: just repeats the code

tests/unit/test_rate_limit.py
```diff
@@ -0,0 +1,133 @@
...
+    for _ in range(50):
+        delay = handler.next_delay(attempt=3)
+        # compute_backoff(3) = 16.0 -> jittered in [0, 16].
+        assert 0.0 <= delay <= 16.0
+
+
+def test_next_delay_result_always_clamped() -> None:
+    handler = make_handler(backoff_base=1000.0, backoff_max=5.0)
+    for _ in range(50):
+        assert handler.next_delay(attempt=5) <= 5.0
```
ajtmccarty, line 115: should this `attempt` be incrementing?

tests/unit/sdk/test_rate_limit_retry.py
```diff
@@ -0,0 +1,783 @@
+"""Client-level tests for transparent HTTP 429 retry on the async and sync clients.
+
+Covers (in later implementation chunks): transparent 429->200 retry, honouring
+``Retry-After``, clean ``RateLimitError`` on exhaustion, the disabled path, async/sync
+parity, all-paths coverage (regular request, multipart, streaming init), and the E2/X1
+multipart body re-read regression. This module currently holds the shared imports/skeleton.
```
ajtmccarty: "later implementation chunks" looks leftover. not sure what this means "This module currently holds the shared imports/skeleton."

tests/unit/sdk/test_rate_limit_retry.py
```diff
@@ -0,0 +1,783 @@
...
+    from pytest_httpx import HTTPXMock
+
+__all__ = [
+    "Config",
+    "InfrahubClient",
+    "InfrahubClientSync",
+    "RateLimitError",
+    "httpx",
+    "pytest",
+]
```
ajtmccarty: this seems wrong. shouldn't each test file just import what it needs from where it actually exists?

tests/unit/sdk/test_rate_limit_retry.py
```diff
@@ -0,0 +1,783 @@
...
+
+    # The wait came from jittered exponential backoff for the first retry (attempt=0), not the
+    # header, so it lands within ``[0, compute_backoff(0)]`` of a handler built from Config defaults.
+    defaults = Config(address="http://mock")
+    handler = RateLimitRetryHandler(
+        max_retries=defaults.rate_limit_max_retries,
+        backoff_base=defaults.rate_limit_backoff_base,
+        backoff_max=defaults.rate_limit_backoff_max,
+    )
+    ceiling = handler.compute_backoff(attempt=0)
```
ajtmccarty, line 240: is this testing that `compute_backoff(0) <= compute_backoff(0)`?

tests/unit/sdk/test_rate_limit_retry.py
```diff
@@ -0,0 +1,783 @@
...
+@pytest.mark.parametrize("client_type", CLIENT_TYPES)
+async def test_request_disabled_surfaces_raw_429_without_retry(
+    client_type: str, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """With ``rate_limit_retry_enabled=False`` the driver does ONE send and returns the raw 429.
+
+    The ``_request`` path (the one the existing tests drive) returns the response untouched, so no
+    ``RateLimitError`` is raised and no wait occurs. A higher-level caller that later invokes
+    ``raise_for_status()`` would surface the underlying ``httpx.HTTPStatusError`` (never a
+    ``RateLimitError``); this asserts the driver behaviour directly.
```
ajtmccarty: I don't really understand what this means and it seems unnecessary. the first line captures what the test does

tests/unit/sdk/test_rate_limit_retry.py
ajtmccarty, line 1: this looks like it could be split into several test files. maybe 5 or 6. and I think there might be some overlap in here, but they're unit tests, so it doesn't really matter.

tests/unit/sdk/test_rate_limit_retry.py
```diff
@@ -0,0 +1,783 @@
...
+    assert recorded_sleeps == pytest.approx([1.0, 2.0, 4.0, 6.0])
+
+
+@pytest.mark.parametrize("client_type", CLIENT_TYPES)
+async def test_jitter_differs_between_instances(client_type: str, monkeypatch: pytest.MonkeyPatch) -> None:
+    """Two independent clients driven through the same header-less 429 sequence jitter differently.
+
+    With real full jitter (``jittered_delay`` NOT patched), the per-retry waits are random draws in
+    ``[0, compute_backoff(attempt)]``. Across four retries an exact match between two independent
+    instances is astronomically unlikely, so at least one position must differ (SC-003).
```
ajtmccarty: leftover "SC-003"

tests/unit/sdk/test_rate_limit_retry.py
```diff
@@ -0,0 +1,783 @@
...
+        runs.append(list(recorded_sleeps))
+
+    first, second = runs
+    # Both instances performed the same number of jittered waits ...
+    assert len(first) == len(second) == max_retries
+    # ... but real full jitter makes at least one recorded wait diverge between the two instances.
+    assert first != second
+
+
+# --- FR-006 / E2/X1: all-paths coverage and multipart body re-read ---------------------------
```
ajtmccarty: FR-006

tests/unit/sdk/test_rate_limit_retry.py
```diff
@@ -0,0 +1,783 @@
...
+    content_type = request.headers["content-type"]
+    _, _, boundary = content_type.partition("boundary=")
+    return request.content.replace(boundary.encode(), b"__BOUNDARY__")
+
+
+@pytest.mark.parametrize("client_type", CLIENT_TYPES)
+async def test_multipart_body_survives_retry(
+    client_type: str, httpx_mock: HTTPXMock, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """E2/X1: a retried multipart upload re-sends the FULL file body, not a consumed/empty stream.
```
ajtmccarty: is `E2/X1` a spec reference of some sort?

ajtmccarty (approved):
I like that this is trying to be SOLID with the `RateLimitRetryHandler`. I'd like it to be even more SOLID, but I'm not sure that's really possible without more refactoring of how the SDK client works.

infrahub_sdk/client.py
dgarros: Done — the retry loop now lives in `RateLimitRetryHandler` as `asend()` (async) and `send()` (sync), so each call site is just `await self._rate_limit_handler.asend(send=send, url=url)` / `self._rate_limit_handler.send(...)`. Both `_send_with_rate_limit_retry` methods on the client are gone.

Honest caveat, as you predicted: because we support both sync and async this **centralises** the logic without **de-duplicating** it — `asend` and `send` are still near-identical bodies (differing only in `await send()`/`asyncio.sleep` vs `send()`/`time.sleep`). The win is that they now sit side by side in one class, diffable by eye, instead of ~2,000 lines apart in the client. The pure decision methods (`next_delay`, `should_retry`, …) stay I/O-free and unit-tested in isolation. Existing tests already drive the real client `_request` path, so they cover `asend`/`send` end-to-end (64 pass).

infrahub_sdk/client.py
dgarros: I kept the callback signature rather than introducing a `RequestSenderInterface`, but moved the orchestration into `RateLimitRetryHandler.asend()/send()` so the loop is no longer inline in the client — which addresses most of the "hard to read/debug/test" concern.

I went back and forth on the full Sender interface. With the async/sync split it would need an async `Sender` and a sync `Sender` (different return types), and our 6 call sites are one-liner closures wrapping `client.stream`/`client.post`/`requester(...)` — turning each into a class felt like more boilerplate than the modularity bought us, since the substantive logic is already isolated and tested in the handler. Happy to revisit if you feel strongly, but I'd lean on doing it as a follow-up if we ever unify the sync/async client structure.

infrahub_sdk/client.py
dgarros: Good catch on the double construction. The handler is intentionally stateless — attempt count is passed in per call so a single instance is safe to share across concurrent requests — so I've moved construction into `BaseClient.__init__` (`self._rate_limit_handler = RateLimitRetryHandler(...)`). It's now built once per client and reused for every request, instead of once in each of the async/sync drivers.

I kept it as a per-instance attribute rather than a module-level global on purpose: the retry config (`rate_limit_max_retries`, `backoff_base/max`, `enabled`) comes from each client's `Config`, so different `InfrahubClient` instances can carry different settings — a shared global would break that.

Issue #1152, opened 2026-07-09
bug: assigning to a cardinality-many relationship is silently accepted and fails at save() with a misleading error


PhillSimonds:
### Component

Python SDK

### Infrahub SDK version

1.22.0 (behavior present at least since 1.17.0)

### Current Behavior

Setting a cardinality-**many** relationship via attribute assignment (`node.rel = value`) is silently accepted at assignment time but corrupts the node, and only fails later at `save()` with an error that names the wrong thing.

`InfrahubNode.__setattr__` (`infrahub_sdk/node/node.py:954`) only intercepts cardinality-**one** relationships — it checks `name in self._relationship_cardinality_one_data` and nothing else. Any assignment to a cardinality-many relationship falls through to `super().__setattr__(...)`, which stores the raw value as a normal instance attribute, shadowing the `RelationshipManager` held in `_relationship_cardinality_many_data`.

At `save()`, `_generate_input_data` (`node.py:480`) reads `rel.initialized` for every relationship. The shadowing value is now a raw object with no `.initialized`, so:

- assigning a single node (e.g. `acl.entries = entry`) → `'InfrahubNode' object has no attribute 'initialized'`
- assigning a list (e.g. `acl.entries = [entry]`) → `'list' object has no attribute 'initialized'`

The error names the object itself (`InfrahubNode`/`list`) as the cause, never the relationship or the real cause, so it's very hard to trace back to the assignment line.

There is an asymmetry that makes this particularly surprising: passing the same many-relationship as a **list inside the `data=` dict** at `create()` time works fine (that path builds the manager via `_init_relationships`). Only *attribute assignment after initial construction* is unsupported — and it fails silently rather than loudly.

### Expected Behavior

Assigning to a cardinality-many relationship should either work or fail immediately and clearly. Ideally the SDK would both:

1. **Support list assignment** — `node.rel = [peer1, peer2]` should populate the `RelationshipManager` (mirroring the `data=` behavior), removing the asymmetry.
2. **Reject non-list values at assignment time** with an actionable message, e.g. `"'entries' is a cardinality-many relationship; assign a list (or use .add()/.extend()), not a single node."`

Either way the failure must happen at the time of assignment, not be deferred to `save()` with a type-name error.

### Steps to Reproduce

```python
# Schema: node kind "Parent" with a cardinality-many relationship "children" -> "Child"
parent = await client.create(kind="Parent", data={"name": "p1"})
[...]

PR #1155, opened 2026-07-10
fix: Print clear error when fetching/editing a relationship on an id-less node

### LINKED ISSUE #1153: bug: RelationshipManager.fetch()/add() give misleading errors when the node has no id
### Component

Python SDK

### Infrahub SDK version

1.22.0

### Current Behavior

When a `RelationshipManager` belongs to a node that has no `id` (e.g. a node built with `client.create()` but never `save()`d, so it was never persisted), editing the relationship produces two confusing errors that don't point at the real cause:

- Error 1: `node.rel.fetch()` → `ValueError: At least one filter must be provided to get()`: `fetch()` (`infrahub_sdk/node/relationship.py:234`) re-fetches the parent node via `self.client.get(..., id=self.node.id, ...)` (`relationship.py:250`). With `id=None`, `get()` has zero filters and raises the generic message from `client.py:584`. Nothing tells the user the problem is the node's missing id, or gives an indication of *why* that id might be missing (e.g. the node was created but not saved).
- Error 2: `node.rel.add(...)` → `UninitializedError: Must call fetch() on RelationshipManager before editing members` (`relationship.py:296`). This tells the user to call `fetch()`, but for a node with no id `fetch()` can't succeed either — so the two errors point at each other and give the user no clear path forward.

### Expected Behavior

`fetch()` (and the `add()`/`remove()` guards) should detect the missing identifier and raise an actionable message instead of the generic filter error, e.g.:

> `Unable to fetch the 'members' relationship: this Parent has no ID, so it hasn't been saved to Infrahub yet (or wasn't loaded from it). Call .save() first, or fetch the node before accessing its relationships.`

It would be nice if the message is phrased around "no ID to look it up by *and* the node is *likely* not yet persisted in infrahub." Just raising that the ID doesn't exist requires the user to think about *why* the ID doesn't exist. That could be confusing for a user who doesn't yet understand that .create() creates locally but does not save data to infrahub and that a call to .save() is required to generate the id in infrahub. The user might also not understand that node.rel fetches anything from infrahub, which is another thing the user needs to think through to understand why there is no id. There are cases in which the ID doesn't exist for other reasons, but the id normally doesn't exist because the node has not been saved, so a nudge to save the node (a quick why they are hitting this error) in addition to explaining the node does not yet have an id (the technical reason we cant fetch the node) is probably the right move so we make the SDK easy for users to consume, understand, and troubleshoot.

### Steps to Reproduce

```python
# Parent has a cardinality-many relationship "children"
parent = await client.create(kind="Parent", data={"name": "p1"})  # NOT saved -> parent.id is None

await parent.children.fetch()
# -> ValueError: At least one filter must be provided to get()

parent.children.add(some_child)
# -> UninitializedError: Must call fetch() on RelationshipManager before editing members
#    (but fetch() can't succeed either, since parent has no id)
```

Fix on the user side is to `await parent.save()` first (which backpopulates `parent.id`); the request here is only that the error message say so.

### Additional Information

- `fetch()`: `relationship.py:234` → `get(id=self.node.id)` at `:250`; generic error at `client.py:584`.
- `add()`/`remove()` guards: `relationship.py:296` / `:340`.
- Related: sibling issue #1152 (assigning to a cardinality-many relationship) — same "silent/confusing failure in relationship editing" theme.

PhillSimonds:
## What

Fixes #1153.

Fetching or editing a cardinality-many relationship on a node with no ID (e.g. one built with `client.create()` but never `save()`d) produced confusing errors:

- `node.rel.fetch()` → `ValueError: At least one filter must be provided to get()` — because `fetch()` re-queries the parent via `client.get(id=self.node.id)` with `id=None`.
- `node.rel.add(...)` / `remove(...)` → `Must call fetch() on RelationshipManager before editing members` — but `fetch()` can't succeed either, so the two guards send the caller in a circle.

## Change

`fetch()`, `add()`, and `remove()` (both `RelationshipManager` and `RelationshipManagerSync`) now detect the missing node ID up front and raise a clear `UninitializedError` that nudges the user to `.save()` first, via a shared `_raise_missing_identifier()` helper. When the node **does** have an ID, the existing `"Must call fetch()"` guidance is preserved.

The message is deliberately phrased around "no ID to look it up by / likely not saved yet" rather than asserting the node doesn't exist in Infrahub — at that point only a local attribute has been inspected, no server call has been made.

## Tests

Added to `tests/unit/sdk/test_node.py` (async + sync):

- `test_fetch_relationship_without_node_id_raises_clear_error`
- `test_edit_relationship_without_node_id_raises_clear_error`
- `test_edit_relationship_with_node_id_still_requires_fetch` (regression: id-bearing node still gets "call fetch()")

`uv run invoke format lint-code` clean; `uv run pytest tests/unit/sdk/test_node.py` → 228 passed.

## Related

Sibling PR for #1152 (assigning to a cardinality-many relationship).

---
## Summary by cubic
Show a clear error when fetching or editing a relationship on a node without an ID, guiding users to save the node first. Prevents the generic `get()` error and the circular “call fetch()” guidance.

- **Bug Fixes**
  - In `RelationshipManager` and `RelationshipManagerSync`, `fetch()`, `add()`, and `remove()` now detect a missing node ID and raise `UninitializedError` with a hint to call `.save()` first.
  - Preserves the original “Must call fetch()” error when the node has an ID but the relationship isn’t loaded; added unit tests covering both cases.

<sup>Written for commit df30f2747b0974539d83c2d0ff40253d753ac7db. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1155?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/infrahub_sdk/node/relationship.py
+++ b/infrahub_sdk/node/relationship.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from collections import defaultdict
 from collections.abc import Iterable
-from typing import TYPE_CHECKING, Any, Generic, cast
+from typing import TYPE_CHECKING, Any, Generic, NoReturn, cast
 
 from ..exceptions import (
     Error,
@@ -17,6 +17,25 @@ if TYPE_CHECKING:
     from ..client import InfrahubClient, InfrahubClientSync
     from ..schema import RelationshipSchemaAPI
     from .node import InfrahubNode, InfrahubNodeSync
+
+
+def _raise_missing_identifier(node: InfrahubNode | InfrahubNodeSync, name: str) -> NoReturn:
+    """Raise a clear error for fetching/editing a relationship on a node with no ID.
+
+    A relationship can only be fetched or edited once the parent node has an ID to look it up
+    by. When it does not, the underlying ``client.get()`` call would otherwise fail with a
+    generic "At least one filter must be provided to get()" message that gives the caller no
+    hint about the real cause.
+
+    Raises:
+        UninitializedError: Always; the parent node has no ID to look the relationship up by.
+
+    """
+    raise UninitializedError(
+        f"Cannot access the '{name}' relationship because the {node._schema.kind} node has no ID to "
+        f"look it up by. This usually means the node was created locally but not saved yet — call "
+        f".save() on it first (or fetch it from Infrahub) before fetching or editing its relationships."
+    )
 
 
 class RelationshipManagerBase(Generic[PeerT]):
@@ -243,6 +262,8 @@ class RelationshipManager(RelationshipManagerBase[PeerT]):
 
         """
         if not self.initialized:
+            if not self.node.id:
+                _raise_missing_identifier(self.node, self.name)
             exclude = self.node._schema.relationship_names + self.node._schema.attribute_names
             exclude.remove(self.schema.name)
             node = await self.client.get(
@@ -293,6 +314,8 @@ class RelationshipManager(RelationshipManagerBase[PeerT]):
 
         """
         if not self.initialized:
+            if not self.node.id:
+                _raise_missing_identifier(self.node, self.name)
             raise UninitializedError("Must call fetch() on RelationshipManager before editing members")
         new_node = cast(
             "RelatedNode[PeerT]", RelatedNode(schema=self.schema, client=self.client, branch=self.branch, data=data)
@@ -337,6 +360,8 @@ class RelationshipManager(RelationshipManagerBase[PeerT]):
 
         """
         if not self.initialized:
+            if not self.node.id:
+                _raise_missing_identifier(self.node, self.name)
             raise UninitializedError("Must call fetch() on RelationshipManager before editing members")
         node_to_remove = RelatedNode(schema=self.schema, client=self.client, branch=self.branch, data=data)
 
@@ -440,6 +465,8 @@ class RelationshipManagerSync(RelationshipManagerBase[PeerTSync]):
 
         """
         if not self.initialized:
+            if not self.node.id:
+                _raise_missing_identifier(self.node, self.name)
             exclude = self.node._schema.relationship_names + self.node._schema.attribute_names
             exclude.remove(self.schema.name)
             node = self.client.get(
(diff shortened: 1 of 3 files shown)


infrahub_sdk/node/relationship.py
```diff
@@ -19,6 +19,25 @@
     from .node import InfrahubNode, InfrahubNodeSync
 
 
+def _raise_missing_identifier(node: InfrahubNode | InfrahubNodeSync, name: str) -> NoReturn:
+    """Raise a clear error for fetching/editing a relationship on a node with no ID.
+
+    A relationship can only be fetched or edited once the parent node has an ID to look it up
+    by. When it does not, the underlying ``client.get()`` call would otherwise fail with a
+    generic "At least one filter must be provided to get()" message that gives the caller no
+    hint about the real cause.
```
ajtmccarty, line 28: if this documentation is needed, it should go somewhere else. the first line of the docstring and the text of the error message cover the actual description of this function's responsibility

infrahub_sdk/node/relationship.py
```diff
@@ -19,6 +19,25 @@
     from .node import InfrahubNode, InfrahubNodeSync
 
 
+def _raise_missing_identifier(node: InfrahubNode | InfrahubNodeSync, name: str) -> NoReturn:
```
ajtmccarty, line 22: I don't think this function really needs to access the whole `node` object. I think passing in `object_kind: str` would be enough. and you can get it using `node.get_kind()` instead of accessing the private-ish `._schema`

infrahub_sdk/node/relationship.py
```diff
@@ -440,6 +465,8 @@ def fetch(self) -> None:
 
         """
         if not self.initialized:
```
ajtmccarty, line 467: haven't tested this, but it looks like if I create a Node with no ID AND relationship data, then `self.initialized` is set to `True`, which would skip your new error message below. not sure if that is a valid use case that should be covered by this change.

infrahub_sdk/node/relationship.py
ajtmccarty, line 1: what happens if a node is initialized with no UUID, but an HFID? would an error still be raised in that case? probably worth adding as a test to confirm whatever the expected behavior should be

PR #1167, opened 2026-07-12
feat: add X-Priority request header to the SDK

### LINKED ISSUE #1151: feature: send an X-Priority header, per-request with a client-wide default
### Component

Python SDK

The SDK should be able to attach an `X-Priority` header to the requests it
sends to Infrahub, so the server can classify traffic for priority-aware
backpressure and shed less-important requests first.

The header carries one of three values — `high`, `normal`, or `low` — where
`normal` is the implicit default when nothing is sent (opting *up* to `high`
or *down* to `low` is the only reason to set it explicitly).

Two ways of setting it are needed:

- **Per-request override** — a caller can set the priority for a single
  request, so the code path that *originates* the request (which knows why
  it's being made) can tag it honestly.
- **Client-wide default** — a client instance can be configured with a
  default priority that applies to every request it issues, unless overridden
  per request. This guarantees that once a default is configured, every use of
  that client honours it (e.g. a client dedicated to background work sends
  `low` on everything).

Both the async (`InfrahubClient`) and sync (`InfrahubClientSync`) clients
should share this behaviour.

### Describe the Use Case

Clients (including `infrahubctl` and the Infrahub Ansible collection, which
use this SDK internally) mix user-blocking work with deferrable background
work — batches, syncs, generators, prefetches. Without a way to signal
importance, the server can only shed load blindly and hurt everyone equally.
Letting callers tag requests — reserving `high` for work a user is actively
waiting on and pushing background/best-effort traffic to `low` — lets the
server protect what matters and shed the right thing first. A configurable
client-wide default means whole classes of traffic (e.g. all background jobs)
can be tagged in one place rather than at every call site.

### Additional Information

This is the client-side complement to #1124 (retry/backoff on HTTP 429). That
issue owns 429 handling; this one is scoped to *sending* the priority header
and configuring it.

Suggested notes (non-binding):

- A single request-priority concept (e.g. `Priority.HIGH/NORMAL/LOW`) that the
  HTTP layer maps to the `X-Priority` header, so call sites express intent
  rather than header strings.
- The client-wide default fits naturally in the existing pydantic `Config`.
- Server contract: values are case-insensitive; an unknown or absent value is
  treated as `normal`, so rollout is safe and can be adopted incrementally.
- Priority should reflect *what a request is*, not be escalated to dodge
  shedding.

Related: #1124

dgarros:
# Why

Infrahub's background systems (generators, artifacts, diffs, syncs, computed attributes) call back into the same API servers through this SDK, competing on equal footing with human/frontend traffic for a shared worker pool and database connections. Today the API layer cannot tell interactive traffic from background traffic, so under heavy background load it cannot preferentially protect the frontend.

**Goal:** let the originator of each request — the SDK caller — declare how important the request is, in a form the server can act on. This is the SDK's contribution to the server-side prioritization effort in INFP-636.

**Non-goals:** no server-side admission control, no 429/`Retry-After` handling (that is GitHub #1124), no traffic-classification guidance.

Closes #1151

## What changed

Behavioral changes (what callers can now do):

- **Client-wide default priority.** Set `Config.priority` (or `INFRAHUB_PRIORITY`, case-insensitive) and every request from that client carries `X-Priority: <value>` across all transports — GraphQL, multipart upload, and raw blob `_get`/`_post` — with no call-site changes. Ideal for a client dedicated to background work (`priority=Priority.LOW`).
- **Per-request override.** Pass `priority=Priority.HIGH` to `get`, `all`, `execute_graphql`, the diff methods, and node `save`/`create`/`update`/`delete` to override the default for exactly one request; the next un-annotated call reverts to the client default.
- **Typed contract.** A new `Priority` enum (`high`/`normal`/`low`) is exported from `infrahub_sdk`, so intent is expressed as a value rather than a raw header string; an invalid configured value is rejected at config-load time.
- **Zero behaviour change when unset.** With no priority configured and none passed per request, no `X-Priority` header is emitted — outgoing requests are unchanged from before this PR.

Implementation notes:

- The default is injected once into the client's base header set, so it rides every transport automatically (mirrors how `X-INFRAHUB-KEY` already works). The per-request override is applied only in the two GraphQL funnels; higher-level methods forward the kwarg and never re-resolve. Resolution is `per_request if per_request is not None else client_default`.
- Both `InfrahubClient` (async) and `InfrahubClientSync` (sync) behave identically.

What stayed the same:

- No new dependencies. No schema/API-contract changes beyond the additive `Priority` enum, `Config.priority` field, and optional `priority=` kwarg. Generated `protocols.py` untouched.
- An included fix keeps the automatic relogin token-refresh working for password-auth clients after the header-merge change (a per-request header snapshot must not shadow a freshly-refreshed `Authorization`).

## Related context

Built with spec-kit; the spec, plan, critique, task breakdown, and implementation report live under `dev/specs/ihs-259-sdk-x-priority-header/`. Server contract per **INFP-636** (header is exactly `X-Priority`; absent/unknown treated as `normal` server-side, which is why omitting the header when unset is a safe, incremental rollout). Complementary: GitHub #1124 (429 retry/backoff).

## How to review

Core change is small and funnels through a few points:

1. `infrahub_sdk/constants.py` — the `Priority` enum + case-insensitive `_missing_`.
2. `infrahub_sdk/config.py` — the `priority` field.
3. `infrahub_sdk/client.py` — base-header injection in `BaseClient.__init__`, the `priority=` handling in `execute_graphql`/`_execute_graphql_with_file`, and `_merge_request_headers` (per-request headers win, but live auth is re-asserted so relogin retries use the fresh token).
4. `infrahub_sdk/node/node.py` — `priority=` forwarding on `save`/`create`/`update`/`delete`.

The `docs/` `.mdx` changes are generated (`uv run invoke docs-generate`).

## How to test

```bash
uv run pytest tests/unit/sdk/test_priority.py tests/unit/sdk/test_config.py tests/unit/sdk/test_relogin_headers.py -q
```

Expected: green across both `standard` (async) and `sync` clients. The suite asserts the header on the wire per transport, no header when unconfigured, per-request override + revert, explicit `NORMAL` beating a `LOW` default, config validation, the full resolution truth table, and the relogin-token-refresh regression. Full SDK suite (`uv run pytest tests/unit/sdk/`) is green (1145 passed).

## Impact & rollout

- **Backward compatibility:** fully additive. Callers who set nothing see byte-identical outgoing requests (no `X-Priority`). Safe to deploy independently of the server work.
- **Performance:** negligible — one dict insertion per request, no new round-trips.
- **Config/env changes:** new optional `Config.priority` field / `INFRAHUB_PRIORITY` env var (default: unset).
- **Deployment notes:** safe to deploy; the header is inert until the server (INFP-636) acts on it.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added (`changelog/1151.added.md`)
- [x] External docs updated (regenerated SDK reference)
- [x] Internal .md docs updated (spec-kit artifacts under `dev/specs/`)

---
## Summary by cubic
Adds an `X-Priority` header to tag SDK requests as `high`, `normal`, or `low` for server-side prioritization. Header is omitted when unset; async and sync clients behave the same.

- **New Features**
  - `Priority` enum available as `infrahub_sdk.constants.Priority`.
  - Client-wide default via `Config.priority` or `INFRAHUB_PRIORITY`; applied to GraphQL, multipart uploads, blob `_get`/`_post`, and the `count` prefetch in `all(parallel=True)`.
  - Per-request override with `priority` on `get`, `all`, `count`, `execute_graphql`, diff APIs, and node `save`/`create`/`update`/`delete`; resource‑pool peer fetch after mutations uses the same priority.

- **Bug Fixes**
  - Per-request headers now layer over base headers via `_request_headers`/`_merge_request_headers`, allowing one-off header overrides and keeping refreshed `Authorization` on relogin retries.
  - Lazily import `pyarrow` in the line-delimited JSON importer so non-`object load` `infrahubctl` commands no longer require the `ctl` extra.

<sup>Written for commit 432a99d31e8bd21b4af71f20127b1db58e8f1eb1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1167?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


infrahub_sdk/__init__.py
```diff
@@ -4,11 +4,13 @@
 
 from .client import InfrahubClient, InfrahubClientSync
 from .config import Config
+from .constants import Priority
 
 __all__ = [
     "Config",
     "InfrahubClient",
     "InfrahubClientSync",
+    "Priority",
```
ajtmccarty: I'd say, don't add this here. it makes it so that importing `Priority` requires importing `Config` and the clients

infrahub_sdk/client.py
```diff
@@ -679,9 +706,15 @@ async def count(
...
         partial_match: bool = False,
         query_name: str | None = None,
+        priority: Priority | None = None,
         **kwargs: Any,
     ) -> int:
-        """Return the number of nodes of a given kind."""
+        """Return the number of nodes of a given kind.
+
+        Args:
+            priority: Override the client-wide request priority for this query. When None, the client default is used.
```
ajtmccarty: seems weird to document just 1 arg

infrahub_sdk/client.py
```diff
@@ -1198,7 +1246,7 @@ def clone(self, branch: str | None = None) -> InfrahubClient:
         """Return a cloned version of the client using the same configuration."""
         return InfrahubClient(config=self.config.clone(branch=branch))
 
-    async def execute_graphql(
+    async def execute_graphql(  # noqa: PLR0912
```
ajtmccarty: maybe this can be refactored to build the payload or headers or do the retry loop in a helper method to remove this typing ignore.

infrahub_sdk/client.py
```diff
@@ -3317,11 +3416,16 @@ def get_diff_tree(
...
         tracker: str | None = None,
+        priority: Priority | None = None,
     ) -> DiffTreeData | None:
         """Get complete diff tree with metadata and nodes.
 
         Returns None if no diff exists.
 
+        Args:
+            priority: Per-request priority emitted as the X-Priority header, overriding the client
+                default for this request only. When None, the client default (if any) is used.
```
ajtmccarty: same here

infrahub_sdk/constants.py
```diff
@@ -1,7 +1,30 @@
...
     TRACKING = "tracking"
     # IDEMPOTENT = "idempotent"
+
+
+class Priority(str, enum.Enum):
+    """Request priority emitted as the ``X-Priority`` header.
+
+    String-valued closed enum accepting values case-insensitively (e.g. "LOW",
+    "Low" and "low" all resolve to :attr:`Priority.LOW`). Unknown values raise
+    ``ValueError``, which surfaces as a ``pydantic.ValidationError`` at config load.
```
ajtmccarty: "Case-insensitive" covers it. it's obviously a string, we know what case-insensitive means, raising a ValueError is standard enum behavior, mentioning the "config" load is not the responsibility of this class

tests/unit/sdk/test_config.py
```diff
@@ -93,3 +97,73 @@ def test_password_auth_overrides_env_token_when_password_env_var_and_username_ex
...
+
+
+@dataclass
+class PriorityCase:
+    name: str
+    value: str | Priority
+    expected: Priority
+
+
+PRIORITY_CASES = [
```
ajtmccarty, line 123: very thorough

tests/unit/sdk/test_priority.py
```diff
@@ -0,0 +1,924 @@
...
+    else:
+        response = client.object_store.upload(content="any content")
+
+    assert response == {"checksum": "yyyyyyyyyyyyyy", "identifier": "xxxxxxxxxx"}
+    requests = [r for r in httpx_mock.get_requests() if r.method == "POST"]
+    assert len(requests) == 1
+    assert "x-priority" not in requests[0].headers
+
+
+@pytest.mark.parametrize("client_type", client_types)
```
ajtmccarty, line 242: I don't know if there's a good way to parametrize these test to handle both a defined and undefined `priority`, but it would probably reduce the size of the test file somewhat

tests/unit/sdk/test_priority.py
```diff
@@ -0,0 +1,924 @@
...
+        client.execute_graphql(query=query)
+
+    requests = [r for r in httpx_mock.get_requests() if r.method == "POST"]
+    assert len(requests) == 2
+    assert requests[0].headers["x-priority"] == "high"
+    assert requests[1].headers["x-priority"] == "low"
+
+
+@pytest.mark.parametrize("client_type", client_types)
+async def test_override_normal_beats_low_default(
```
ajtmccarty: this one looks like it is already handled by the above two

tests/unit/sdk/test_priority.py
```diff
@@ -0,0 +1,924 @@
...
+            "data": {"BuiltinLocationUpdate": {"ok": True, "object": {"id": "17aec828-9814-ce00-3f20-1a053670f1c8"}}}
+        },
+        is_reusable=True,
+    )
+
+    data = {"id": "17aec828-9814-ce00-3f20-1a053670f1c8", "name": {"value": "JFK1"}, "type": {"value": "SITE"}}
+    client = getattr(clients, client_type)
+    if client_type == "standard":
+        node = InfrahubNode(client=client, schema=location_schema, data=data)
+        node.name.value = "JFK2"  # type: ignore[union-attr]
```
ajtmccarty: I think `node._get_attribute("name").value` would let you remove the typing ignore

tests/unit/sdk/test_priority.py
```diff
@@ -0,0 +1,924 @@
...
+        node.save(priority=Priority.HIGH)
+
+    requests = [r for r in httpx_mock.get_requests() if r.method == "POST"]
+    assert len(requests) == 1
+    assert requests[0].headers["x-priority"] == "high"
+    assert requests[0].headers.get("content-type").startswith("multipart/form-data;")
+
+
+# ---------------------------------------------------------------------------
+# User Story 5: async / sync parity
```
ajtmccarty: mention of User Story here and above is unnecessary

tests/unit/sdk/test_priority.py
```diff
@@ -0,0 +1,924 @@
...
+    """
+
+    name: str
+    client_default: Priority | None
+    per_request: Priority | None
+    expected: str | None
+
+
+# The full resolution truth table from data-model.md. Each row must resolve
+# identically on both the async and sync clients (SC-005).
```
ajtmccarty: mentions spec stuff

should just check the whole file for SC-00X and User Story and .md references

tests/unit/sdk/test_priority.py
```diff
@@ -0,0 +1,924 @@
...
+def _client_with_default(client_type: str, default: Priority | None) -> InfrahubClient | InfrahubClientSync:
+    config = Config(address="http://mock", insert_tracker=True, pagination_size=3, priority=default)
+    if client_type == "standard":
+        return InfrahubClient(config=config)
+    return InfrahubClientSync(config=config)
+
+
+@pytest.mark.parametrize("client_type", client_types)
+@pytest.mark.parametrize("case", [pytest.param(tc, id=tc.name) for tc in RESOLUTION_TRUTH_TABLE])
+async def test_resolution_truth_table_parity(case: ResolutionCase, client_type: str, httpx_mock: HTTPXMock) -> None:
```
ajtmccarty, line 715: looks like this makes some of the `execute_graphql` tests above unnecessary, but I could be missing something

tests/unit/sdk/test_priority.py
```diff
@@ -0,0 +1,924 @@
...
+        json={"data": {"CoreRepository": {"count": 5}}},
+        match_headers={"X-Priority": "high"},
+        is_reusable=True,
+    )
+
+    client = getattr(clients, client_type)
+    if client_type == "standard":
+        result = await client.count(kind="CoreRepository", priority=Priority.HIGH)
+    else:
+        result = client.count(kind="CoreRepository", priority=Priority.HIGH)
```
ajtmccarty, line 759: does this need to assert the priority on the requests or is `match_headers` in the `httpx_mock` enough?

tests/unit/sdk/test_relogin_headers.py
```diff
@@ -0,0 +1,88 @@
...
+        url="http://mock/graphql/main",
+        json={"data": {"InfrahubInfo": {"version": "1.0"}}},
+    )
+
+    client = _build_password_client(client_type)
+    query = "query { InfrahubInfo { version }}"
+    if isinstance(client, InfrahubClient):
+        await client.execute_graphql(query=query, branch_name="main")
+    else:
+        client.execute_graphql(query=query, branch_name="main")
```
ajtmccarty: should this test include the `priority` override that seems to have semi-caused the issue? or the Authorization / X-INFRAHUB-KEY header?

CLAUDE.md
ajtmccarty, line 1: leftover

infrahub_sdk/client.py
```diff
@@ -239,6 +242,22 @@ def _echo(self, url: str, query: str, variables: dict | None = None) -> None:
...
+        re-asserted from ``self.headers`` so that a token refreshed mid-flight during the
+        automatic relogin retry always wins over a stale per-request snapshot.
+        """
+        merged = copy.copy(self.headers or {})
+        if headers:
+            merged.update(headers)
+        for auth_key in ("Authorization", "X-INFRAHUB-KEY"):
+            if auth_key in self.headers:
+                merged[auth_key] = self.headers[auth_key]
+        return merged
```
ajtmccarty, line 271: I'm not entirely sure I understand how this issue might arise. the `@handle_relogin` calls `Client.login`, which sets `self.headers["Authorization"]`. so it seems like this would only be an issue if someone was including their auth information in the `headers` dict for every individual request. in which case, I would think that we should respect that b/c the user would be affirmatively forcing themself to handle relogin on their own. preventing them from setting a per-request override for these two keys feels wrong

dgarros: Agreed — the re-assertion was the wrong fix and has been reworked.

Root cause: the per-request `priority` override has to win over the client default in `self.headers`, so the transport merge was flipped to let per-request headers win. But the GraphQL methods built the per-request dict as a full copy of `self.headers` (auth included), so on a relogin retry the stale snapshot's `Authorization` beat the freshly-refreshed token — hence the band-aid, which also clobbered any caller-supplied per-request auth.

Fixed in b7b11bb by not snapshotting at all: the GraphQL methods and `object_store` now pass only the request delta (`X-Infrahub-Tracker`, `X-Priority`) via a `_request_headers` helper, merged over a freshly-copied `self.headers`. So live auth always applies (relogin works with no special-casing), the delta still wins for its keys, and a per-request auth override is now respected. The auth loop and the `# noqa: PLR0912` are gone; tests cover the refreshed-token + priority-override retry and a per-request auth override.

infrahub_sdk/client.py
dgarros: Good catch, looking into how we can enforce this check in the future

ajtmccarty (approved):

Issue #855, opened 2026-02-26
task: Move from ujson to orjson


PhillSimonds:
### Component

Python SDK

### Task Description

The  infrahub-sdk relies on package ujson. I have an environment that's flagging the use of ujson as a security issue. The Github repo for ujson directly recommends moving to orjson instead of ujson because its "architecture is fundamentally ill-suited to making changes without risk of introducing new security vulnerabilities":

> [!WARNING]
> UltraJSON's architecture is fundamentally ill-suited to making changes without
> risk of introducing new security vulnerabilities. As a result, this library
> has been put into a *maintenance-only* mode. Support for new Python versions
> will be added and critical bugs and security issues will still be
> fixed but all other changes will be rejected. Users are encouraged to migrate
> to [orjson](https://pypi.org/project/orjson/) which is both much faster and
> less likely to introduce a surprise buffer overflow vulnerability in the
> future.

PR #1185, opened 2026-07-16
refactor: standardize SDK JSON serialization on orjson

### LINKED ISSUE #1165: tech-debt: standardize JSON library usage (json vs ujson) across the SDK
## Need

The SDK uses two JSON libraries interchangeably with no documented rule for
which to use when. Across `infrahub_sdk/`, ~16 modules use `ujson` (the apparent
house standard) and ~7 use stdlib `json`; `playback.py` and `utils.py` import
both. This makes the codebase harder to reason about and creates a latent
exception-handling hazard: `ujson` raises its own `ujson.JSONDecodeError` and
stdlib raises `json.JSONDecodeError`, and the two do not catch each other. Today
every decode/except pair happens to be internally consistent, but nothing
enforces that — a future `ujson.loads` guarded by `except json.JSONDecodeError`
(or the reverse) would silently fail to catch a malformed-input error.

## Context

Concrete smells found while surveying usage:

- **`playback.py`** serializes with `json.dumps` (:52) but reads the recorded
  files back with `ujson.load` (:57). This works today (both speak valid JSON)
  but the asymmetry is a smell and imports both libraries in one module.
- **`utils.py`** also imports both `json` and `ujson`.
- The decode/except pairs are currently consistent everywhere
  (e.g. pytest_plugin items pair `ujson.loads` with `except ujson.JSONDecodeError`;
  `infrahub_filters.py` and `parsers.py` pair stdlib `json.loads` with
  `except json.JSONDecodeError`), so this is a maintainability/consistency
  concern rather than an active bug.

Note: an earlier framing suspected `utils.py:97` (`decode_json`) of catching the
wrong exception for a `ujson` failure — that turned out to be a false alarm. That
line decodes via httpx's `response.json()` (stdlib-backed), so its
`except json.decoder.JSONDecodeError` is correct.

Affects: Python SDK.

### LINKED ISSUE #1171: tech-debt: migrate JSON serialization to orjson for performance
## Need

The SDK is a foundational library on the hot path of transforms, generators,
checks, exports/imports, and every GraphQL round-trip — JSON encode/decode
throughput directly affects how fast SDK-based automation runs at scale.
Today the SDK relies on `ujson` (and stdlib `json` in places). `orjson`, backed
by Rust (`serde_json`), is substantially faster than both and stricter on
correctness, so adopting it would give a measurable, no-behaviour-change speedup
for users running large syncs and pipelines.

## Context

- This is a deliberate follow-up to #1165 (JSON library standardization). That
  issue picks a single library across the codebase; this one proposes making
  that target `orjson` rather than settling on `ujson`. Sequencing/merging the
  two is up to whoever picks them up.
- Adding `orjson` is a new runtime dependency — flagged here because the repo
  asks to confirm dependency additions.
- Known migration constraints the implementer will hit (not a design, just
  gotchas): `orjson.dumps()` returns `bytes` rather than `str`, so `str(...)` /
  `.encode()` wrappers at call sites like `playback.py:52` and `utils.py:253`
  change; and `orjson` has no `sort_keys=` kwarg (it uses
  `option=orjson.OPT_SORT_KEYS`), which `dict_hash` (`utils.py:253`) depends on.

## References

- Related issue: #1165
- orjson: https://github.com/ijl/orjson

dgarros:
# Why

The SDK used two JSON libraries interchangeably — `ujson` (~16 modules) and stdlib `json` (~7; `utils.py` and `playback.py` imported both) — with no rule for which to use. This left a latent exception-handling hazard: `ujson.JSONDecodeError` and `json.JSONDecodeError` do not catch each other, so a future mismatched decode/except pair would silently fail to catch malformed input.

Goal: converge on a single JSON library — `orjson` (Rust-backed, faster, stricter) — for a no-behaviour-change speedup on the SDK's hot path (transforms, generators, checks, imports/exports, GraphQL round-trips).

Non-goals: no benchmark harness / CI perf gate; no change to the query-group naming scheme; no migration of JSON usage outside `infrahub_sdk/`.

Closes #1165
Closes #1171

## What changed

Behavioral:

- `orjson` is now the **sole** JSON library; `ujson` + `types-ujson` removed and all stdlib `json` usage eliminated in `infrahub_sdk/`.
- JSON output is preserved byte-for-byte at every consumed/compared site (indent=2 → `OPT_INDENT_2`; `infrahubctl` datetime rendering preserved via `OPT_PASSTHROUGH_DATETIME` + `default=str`).
- One documented behaviour change: the tracking-group name derived from query parameters containing **non-ASCII** characters shifts once on upgrade (orjson emits raw UTF-8 where the previous library emitted escaped sequences). Changelog entry added.

Implementation notes:

- All decode/except pairs unified on `orjson.JSONDecodeError` (decoding switched to `orjson.loads(response.content)` at httpx sites) — this incidentally fixes a latent bug in the pytest plugin where `except ujson.JSONDecodeError` never caught the stdlib error raised by `response.json()`.
- `orjson.dumps()` returns bytes: `.decode()` added at str-consuming sites; `OPT_NON_STR_KEYS` preserves prior int-key coercion at arbitrary-data sites; file-object I/O (`recorder`/`playback`) rewritten since orjson has no `load`/`dump`.

What stayed the same: `dict_hash` public signature and `str` return type; ASCII/int/float hashes are byte-identical; API contract unchanged.

### Suggested review order

1. `infrahub_sdk/utils.py` — `dict_hash` + `decode_json`.
2. Decode/except sites (`schema/__init__.py`, `ctl/parsers.py`, `template/infrahub_filters.py`, `pytest_plugin/items/*`).
3. Encode/`.decode()` sites and `pyproject.toml`.
4. Tests + `dev/specs/002-orjson-json-migration/` (decision trail).

## How to review

Focus scrutiny on except-clause correctness at decode sites and the non-ASCII `dict_hash` change. The rest is mechanical import swaps + `.decode()`.

## How to test

```bash
uv sync --all-groups --all-extras
uv run pytest tests/unit/
grep -rn "import ujson" infrahub_sdk/   # returns nothing
uv run invoke lint-code
```

Note: 8 pre-existing unit failures reproduce identically on `stable` (a worktree-path artifact where Rich strips `[...]` from console output) — unrelated to this change. Integration tests require a live Infrahub server (run in CI).

## Impact & rollout

- **Backward compatibility:** No public API change; `dict_hash` still returns `str`. One-time non-ASCII tracking-group rename (documented); orphaned-group cleanup is manual.
- **Performance:** ad-hoc benchmark — orjson `dumps` ~11.8x, round-trip ~2.1x faster than stdlib `json`.
- **Config/env changes:** new runtime dependency `orjson>=3.10`; `ujson` + `types-ujson` removed.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added (`changelog/+orjson-json-migration.changed.md`)
- [x] External docs updated (verified no user-facing doc referenced the JSON library)
- [x] Internal .md docs updated (spec artifacts under `dev/specs/`)

---
## Summary by cubic
Standardized all SDK JSON serialization on `orjson` to unify error handling and improve performance. Output remains functionally the same; non-ASCII JSON is now emitted as raw UTF‑8 and tracking‑group names with non‑ASCII query params shift once.

- **Refactors**
  - Replaced `ujson` and stdlib `json` with `orjson` across `infrahub_sdk/` (encode/decode, CLI, JSON formatter, Jinja filters, pytest plugin).
  - Unified decoding via `orjson.loads(response.content)` with `except orjson.JSONDecodeError`; fixes the pytest plugin mismatch on HTTP JSON errors.
  - Pretty-printing standardized to 2 spaces; preserved sorted keys and datetime rendering via `default=str` + `OPT_PASSTHROUGH_DATETIME`; added `.decode()` where strings are required.
  - Preserved non-string key coercion using `OPT_NON_STR_KEYS` at arbitrary-data sites (CLI transform/formatter, GraphQL multipart, exporter, playback request payload).
  - Updated recorder/playback and transfer import/export to `orjson`; GraphQL string escaping now uses `orjson.dumps`.

- **Migration**
  - No API changes. JSON with non-ASCII characters is now written as raw UTF‑8 (not `\uXXXX`); bytes differ but content is equivalent.
  - One-time shift for tracking-group names when query params contain non-ASCII; clean up old groups manually if needed.
  - Playback recordings made with older SDKs may not be found due to filename differences; re-record if needed.
  - Dependencies: add `orjson>=3.10`; remove `ujson` and `types-ujson`.

<sup>Written for commit b5d4e4bc7b1097bb7c1780b3fec4c34ac2586a83. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1185?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dgarros:
## Response to `cubic-dev-ai` review (16 findings)

Thanks — the findings collapse into three root causes (all inherent `orjson` vs `ujson`/stdlib differences). Addressed in `0eb537c` (code) and `51ff7a4` (docs).

### ✅ Incorporated — code (`0eb537c`)

**Missing `OPT_NON_STR_KEYS`** (a real regression: stdlib/ujson silently coerced non-string dict keys, orjson raises). Added the option at the remaining arbitrary-data sites, matching the exporter/multipart sites that already had it:

- `ctl/cli_commands.py:358` — transform output (strongest case: user transforms can return int-keyed dicts)
- `ctl/formatters/json.py` — both `format_list` and `format_detail`
- `playback.py:51` — request payload

Added a formatter unit test covering int-keyed values (also closes a previously-deferred coverage gap).

### ✅ Incorporated — docs (`51ff7a4`)

**Non-ASCII `\uXXXX` → raw UTF-8** (9 findings). This is correct and our contract overclaimed "byte-for-byte." orjson has no `ensure_ascii` option, and raw UTF-8 is valid JSON that round-trips through the SDK — re-escaping would add hot-path overhead for no functional gain. So we corrected the documentation instead of the behaviour: the changelog and spec now state non-ASCII is emitted as raw UTF-8 at **every** output site (CLI, `validate --out`, telemetry/transfer exports, check logs, pytest messages). Also broadened the quickstart import-check grep to catch `from json import`.

### 📝 Documented as known limitations (accepted, not fixed)

- **`from_json` large-int → float** (`infrahub_filters.py`): real but silent only for integers > 2⁶⁴ (Infrahub IDs are UUID strings); guarding needs a custom parser. Documented.
- **Playback recordings from older SDK versions** won't be located (request-body hash changed): recordings are regenerated test fixtures. Documented.
- **`orjson.loads` UTF-8-only** (`schema/__init__.py`, pytest items): the Infrahub API always responds in UTF-8; the pytest cases only affect pretty-printing of an error fallback (text still shown).

### ↩️ Declined (with reason)

- **Re-adding `\uXXXX` escaping** anywhere — contrary to adopting orjson; adds post-processing on the hot path; raw UTF-8 is standard and generally more readable.
- **`NaN`/`Infinity` handling** — orjson's stricter, spec-valid behaviour is preferable; not expected in API traffic.

ogenstad:
This also relates to #855

PR #1189, opened 2026-07-19
feat(ctl): add `infrahubctl schema format` command

petercrocker:
## What

Adds **`infrahubctl schema format`** — an opinionated, **offline** formatter for Infrahub schema YAML files. Its core job is normalising the *ordering of keys* (lines) within each node, generic, attribute, relationship, and dropdown choice, so hand-authored schemas read consistently and produce small diffs.

```bash
infrahubctl schema format schemas/                 # format in place
infrahubctl schema format schemas/dcim.yml --diff  # preview, no write
infrahubctl schema format schemas/ --check         # CI gate: exit 1 if any file would change
```

## Design

- **Canonical key order** derived from analysing the schema-library (52 files, 464 attributes, 259 relationships) plus the published JSON schema. `name`/`namespace` first, `attributes`→`relationships` last, `order_weight` last within each attribute/relationship, choices as `name`/`label`/`description`/`color`. Unknown keys are preserved (never dropped).
- **Core nodes only** — nodes/generics in a `RESTRICTED_NAMESPACES` namespace (Core, Builtin, Internal, Profile, Template, …) are left untouched.
- **Round-trip via `ruamel.yaml`** — comments (the `# yaml-language-server` header, standalone notes, and inline comments), quoting style, and flow-style sequences like `[manufacturer, name__value]` are all preserved. By default a format run produces a diff that is *purely* key reordering.
- **Semantics-preserving** — the formatter re-parses its own output and raises rather than write if the reloaded data differs from the input (beyond the opt-in transforms below). Verified idempotent across all 53 schema-library files.

### Opt-in transforms (off by default)

Three flags go further and change file content; each is neutralised in the safety check so only its intended effect is allowed, and the base command stays purely cosmetic:

- `--strip-defaults` — remove keys whose value equals the schema default, context-aware (e.g. `optional: true` is stripped on a relationship but kept on an attribute). Grounded in the published JSON-schema defaults; consequential/internal fields (`branch`, `state`, `inherited`, `display`) are intentionally excluded to avoid coupling a schema to a default that could later change.
- `--sort-by-order-weight` — sort attributes and relationships ascending by `order_weight`; items without one keep their authored order and go last.
- `--backfill-order-weight` — give attributes/relationships lacking an `order_weight` a single constant value (1000).

## Schema drift tracking

The canonical ordering is written against a known set of schema properties. A **warn-only** check compares the live JSON schema against a committed baseline so upstream additions/removals are surfaced without blocking releases:

- `infrahub_sdk/ctl/schema_drift.py` + baseline `schema_properties.json`; `invoke schema-drift-check` emits `::warning::` annotations and always exits 0; `invoke schema-drift-update` refreshes the baseline.
- `.github/workflows/schema-drift.yml` runs it on `release: published` and manual dispatch.

## Changes

- `infrahub_sdk/ctl/schema_format.py` (new) — formatting logic (ruamel round-trip, in-place key reordering, opt-in transforms, transform-aware safety guard).
- `infrahub_sdk/ctl/schema.py` — the `format` subcommand (`--check` / `--diff` / `--strip-defaults` / `--sort-by-order-weight` / `--backfill-order-weight`).
- `infrahub_sdk/ctl/schema_drift.py`, `schema_properties.json` (new) — drift detection + committed baseline.
- `tasks.py` — `schema-drift-check` / `schema-drift-update` invoke tasks.
- `.github/workflows/schema-drift.yml` (new) — warn-only drift check.
- `pyproject.toml` — add `ruamel.yaml` to the `ctl` and `all` dependency sets.
- `tests/unit/ctl/test_schema_format.py`, `test_schema_format_app.py`, `test_schema_drift.py` (new) — 36 tests, incl. comment/flow/quote preservation, malformed-input handling, the three opt-in transforms, and drift logic.
- Regenerated `docs/docs/infrahubctl/infrahubctl-schema.mdx` + changelog fragment.

## Verification

- `uv run pytest tests/unit/ctl/test_schema_format*.py tests/unit/ctl/test_schema_drift.py` — 36 passed.
- `uv run invoke format lint-code docs-validate` — clean.
- Across all 53 schema-library files and every flag combination: 0 guard aborts, idempotent on re-run.

A companion PR (opsmill/infrahub-skills#74) documents the command in the `infrahub-managing-schemas` skill.


petercrocker:
Added a **warn-only schema-drift check** so upstream changes to the published JSON schema get surfaced without gating releases.

- `infrahub_sdk/ctl/schema_drift.py` compares the live schema's property sets against a committed baseline (`infrahub_sdk/ctl/schema_properties.json`) and reports added/removed properties. It never raises on drift.
- `invoke schema-drift-check` emits GitHub `::warning::` annotations and always exits 0; `invoke schema-drift-update` refreshes the baseline.
- `.github/workflows/schema-drift.yml` runs it on `release: published` and `workflow_dispatch` — **warn only**, never fails the run.

Rationale for baseline-vs-live (rather than diffing the formatter's key lists directly): the schema has internal fields the formatter deliberately doesn't order (`id`, `state`, `inherited`, `display`, `hierarchy`), so a direct comparison would produce day-one noise. The baseline snapshot makes day one clean and warns only on genuine future drift. When a warning fires, a maintainer updates `schema_format.py` if needed and runs `invoke schema-drift-update`.

Note: because it's release-gated + `workflow_dispatch`, the warning shows up in the release run's annotations/summary. If you'd like it visible earlier, adding a `pull_request` or `schedule` trigger is a one-line change.

infrahub_sdk/ctl/schema.py
```diff
@@ -414,3 +422,153 @@ async def schema_show(
...
+    Args:
+        location: Path of the file on disk.
+        entries: SchemaFile entries parsed for this location (more than one means
+            a genuine multi-document file, which is not supported).
+        check: Report changes without writing.
+        diff: Print a diff instead of writing.
+        options: Opt-in transforms to apply.
+
+    Returns:
+        One of ``"error"``, ``"skipped"``, ``"unchanged"`` or ``"changed"``.
```
ogenstad: It would be cleaner if this returned an enum instead, now there's a loose contract between this function and the if-statement in schema_format()

petercrocker: Good call — done in b787d03. `_format_one_schema_file` now returns a `FormatOutcome` enum (ERROR/SKIPPED/UNCHANGED/CHANGED) and the command loop matches on it with `is`, so the contract is no longer a loose string.

infrahub_sdk/ctl/schema_format.py
```diff
@@ -0,0 +1,431 @@
...
+    normalized = dict(entity)
+    if options.strip_defaults:
+        for key, default in ENTITY_DEFAULTS.items():
+            normalized.setdefault(key, default)
+    for key, defaults in (("attributes", ATTRIBUTE_DEFAULTS), ("relationships", RELATIONSHIP_DEFAULTS)):
+        items = normalized.get(key)
+        if isinstance(items, list):
+            items = [_normalize_item(item, defaults, options) for item in items]
+            if options.sort_by_order_weight:
+                items = sorted(items, key=lambda it: it.get("name", "") if isinstance(it, dict) else "")
```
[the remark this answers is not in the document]
petercrocker: Fixed in b787d03. The guard now neutralises the reorder by sorting each list on a total, content-based key (`json.dumps(item, sort_keys=True)`) rather than by `name`. I went with full content instead of `_order_weight_sort_key` because weight isn't unique either (many items share a weight, and missing ones map to the same sentinel), so a weight key would have the same duplicate-key failure mode. Content sorting permits any intended reorder while still catching a dropped or corrupted item. Added a regression test (`test_sort_permits_same_named_items_with_different_weights`).

infrahub_sdk/ctl/schema_format.py
```diff
@@ -0,0 +1,434 @@
...
+        The formatted YAML text, with comments and quoting preserved.
+
+    Raises:
+        FormatError: If formatting would change the file's meaning beyond the
+            transforms requested via ``options``.
+
+    """
+    options = options or FormatOptions()
+    yaml_handler = _build_yaml()
+    data = yaml_handler.load(raw_text)
```
gmazoyer: A file with a duplicate key kills the whole run here, not just that file.

Round-trip `YAML()` has `allow_duplicate_keys=False`, so it raises ruamel's `DuplicateKeyError`, not a `FormatError`. That slips past the per-file catch and hits `@catch_exception`, which prints a traceback and exits 1. `SchemaFile` uses PyYAML `safe_load`, which allows dup keys (last wins), so a file that `schema load` accepts will crash here. On a folder run, files before it are already written and the ones after never run.

Can we catch the ruamel `YAMLError` and raise `FormatError` instead, so it stays a normal per-file error?

PhillSimonds (approved):
Nice! LGTM

infrahub_sdk/ctl/schema_format.py
petercrocker: Good catch — fixed in 31dbf19. `format_schema_text` now wraps the round-trip `yaml_handler.load()` in `except YAMLError` and re-raises as `FormatError`, so a duplicate-key file (or anything else ruamel parses more strictly than PyYAML) is reported as a normal per-file error and the rest of the run continues. Added a unit test and a CLI test asserting a dup-key file in a folder is reported per-file while the other files still format.

PR #1207, opened 2026-07-23
perf(client): pass pagination offset and limit as GraphQL variables

fatih-acar:
## Summary

Paginated queries generated by the SDK (`all()`, `filters()`, `get()`, and resource pool allocation lookups) now pass `offset` and `limit` as GraphQL variables instead of inlining them into the query text. The query document stays byte-identical across pages, so repeated pages of the same fetch can hit Infrahub's server-side GraphQL query cache (opsmill/infrahub#8102) — the same optimization the frontend adopted in opsmill/infrahub#8245.

## Key Changes

- `client.filters()` (async and sync) renders the query once per call with `$offset`/`$limit` placeholders and sends the per-page values in the GraphQL `variables` payload — previously the query was re-generated and re-rendered for every page with literal values baked in.
- `get_pool_allocated_resources()` (async and sync) does the same for its pagination loop, extending the `$pool_id`/`$resource_id` variables it already used.
- `generate_query_data()` / `generate_query_data_init()` accept variable placeholder strings (e.g. `"$offset"`) for `offset` and `limit`; passing literal ints keeps working unchanged.
- `count()` needed no change: it takes no pagination arguments, so its query text was already stable per filter set.

## Related Context

- opsmill/infrahub#8245 — frontend counterpart of this change
- opsmill/infrahub#8102 — the backend GraphQL query caching this leverages

## Documentation Updates

- Regenerated SDK reference (`docs/docs/python-sdk/sdk_ref/infrahub_sdk/node/node.mdx`) for the widened signatures.
- Changelog fragment `changelog/+graphql-pagination-variables.changed.md`.

## Test Plan

- `uv run pytest tests/unit/` — updated the six operation-name assertions to expect the `($offset: Int!, $limit: Int!)` signature; added `test_method_all_pagination_uses_graphql_variables`, which fetches two pages and asserts the query text is identical across pages while only the variables differ (`{"offset": 0, "limit": 3}` → `{"offset": 3, "limit": 3}`).
- `uv run invoke lint-code` (ruff, ty, mypy) passes; `uv run invoke docs-generate` output is committed.

---
## Summary by cubic
SDK paginated queries now send `offset` and `limit` as GraphQL variables, keeping the query text identical across pages and enabling server-side cache hits. Queries are rendered once per call, and `generate_query_data*` now accept variable placeholders.

- **Refactors**
  - Use GraphQL variables for `offset`/`limit` in `all()`, `filters()`, `get()`, and pool allocation lookups. One rendered document per call; per-page values go in `variables`.
  - `generate_query_data()` and `generate_query_data_init()` accept `int` or variable placeholders (e.g. `"$offset"`, `"$limit"`). Backward compatible.
  - `count()` unchanged.

<sup>Written for commit ac9e0f73b5a9a723fd9c792d1d94a71a608c57bf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1207?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dgarros (approved):
Looks good, ideally I think we should target `infrahub-develop` directly because there are a lot of changes in this branch right now compared to `develop`

PR #1220, opened 2026-07-29
feat(node): return bare IP addresses for attributes declaring allow_prefix [INFP-551]

ajtmccarty:
Counterpart to opsmill/infrahub#10081, the first PR in the Infrahub-side stack for INFP-551.

# Why

Infrahub is gaining a `parameters.allow_prefix` declaration on `IPHost` attributes. When it is `false`, the
attribute holds a **bare** IP address — `10.0.0.1` rather than the `10.0.0.1/32` the server previously
invented — and a real subnet prefix is refused.

Without this change the SDK re-attaches the mask the server just removed: every `IPHost` response is parsed
with `ipaddress.ip_interface`, so a bare `10.0.0.1` becomes `IPv4Interface('10.0.0.1/32')` and the mask
reappears at the SDK boundary. That directly contradicts the feature's contract, which is why this ships
alongside the server change rather than after it.

## What changed

- **Value coercion** (`node/attribute.py`): for `IPHost`, select `ipaddress.ip_address` when `allow_prefix`
  is `False` and `ipaddress.ip_interface` otherwise. The flag is read **tolerantly** as
  `(self._schema.parameters or {}).get("allow_prefix", True)`, because `AttributeSchemaAPI.parameters` is
  `dict[str, Any] | None` — an SDK talking to an older server that does not send the key behaves exactly as
  it does today.
- **Removed the dead `"IPAddress"` kind key.** No such attribute kind exists server-side: the generated
  `AttributeKind` enum has no `IPADDRESS` member, and two tests exercising it were already skipped citing
  that. The `ip_address` callable it referenced becomes the declared-`IPHost` branch.
- **Protocol generation** (`protocols_generator/generator.py`): emit `IPAddress` / `IPAddressOptional` for a
  declared attribute, `IPHost` / `IPHostOptional` otherwise, composing with the existing
  `optional and default_value is None` suffix rule. All four protocol names were already registered and
  imported, so no change was needed to `constants.py`, `protocols_base.py`, or `template.j2`.
- **Regenerated** `schema/generated/{read,write}.py`, which gain `IPHostAttributeParameters{Read,Write}` and
  `IPHostAttribute{Read,Write}`, with `AttributeKind.IPHOST` moving out of the generic variant.
- Documented the version floor in the compatibility matrix.

## Why the generated models matter to the server

More than a type nicety: `SchemaLoadAPI` on the server inherits these write models. Without
`IPHostAttributeWrite`, an `IPHost` attribute matches `GenericAttributeWrite`, whose `parameters` is
field-less with `extra="ignore"` — so `allow_prefix` is **silently dropped from every schema-load payload**
and the server-side feature is inert. The Infrahub PR cannot pass CI until this branch is available.

## How to review

`node/attribute.py` and `protocols_generator/generator.py` are the whole change; everything else is
generated output or tests.

Coverage follows a three-way pattern rather than two: **declared**, **undeclared**, and **schema payload
with no `allow_prefix` key at all** — the last being the old-server tolerance case, which is the one most
likely to regress silently.

- `test_node_IPHost_deserialization_honours_allow_prefix` — 7 cases × 2 client types
- `test_filter_render_iphost_attribute` — 8 cases
- Two previously-skipped `IPAddress` tests repurposed onto declared `IPHost` rather than deleted

## Merge order

This should merge **before** opsmill/infrahub#10081. That PR pins this branch's commit provisionally so
both can be reviewed in parallel; once this merges, the Infrahub submodule pointer is re-pinned to the
merged commit and confirmed an ancestor of `infrahub-develop`.

---
## Summary by cubic
Return bare IP addresses for `IPHost` attributes that set `parameters.allow_prefix: false`, and mark them as `IPAddress` in generated protocols. Addresses INFP-551 by delivering bare IPs without a new attribute kind, while keeping old-server behavior.

- **New Features**
  - Coerce `IPHost` with `ipaddress.ip_address` when `allow_prefix` is false; otherwise use `ipaddress.ip_interface`.
  - Generate `IPAddress`/`IPAddressOptional` types for flagged `IPHost`; otherwise keep `IPHost`/`IPHostOptional`.
  - Add `IPHostAttribute{Read,Write}` with `parameters.allow_prefix` (regenerated after description tweak) and update compatibility docs with min versions (SDK ≥ 1.23.0, Infrahub ≥ 1.11).

- **Migration**
  - Behavior change: flagged attributes return `IPv4Address`/`IPv6Address` (no mask) instead of interfaces.
  - Compatibility: older servers are safe; when `allow_prefix` is absent, values remain interfaces. The server must be ≥ 1.11 to publish the flag.

<sup>Written for commit 525b28fcc2253e2d1442e24d4ad8e68500d828ed. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1220?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/changelog/+bare-iphost-attribute.added.md
+++ b/changelog/+bare-iphost-attribute.added.md
@@ -1,0 +1,8 @@
+Added support for `IPHost` attributes declaring `allow_prefix: false`. Their values are exposed as bare
+`ipaddress.IPv4Address`/`IPv6Address` objects (no prefix), serialized back as a bare-address string, and
+typed as `IPAddress` in generated protocols. `IPHost` attributes that do not declare the parameter keep
+returning `ipaddress.IPv4Interface`/`IPv6Interface`, as does any attribute read from a server that does
+not publish the parameter.
+
+This replaces the `IPAddress` attribute kind announced earlier in this release cycle: no such attribute
+kind exists on the Infrahub server, so the code paths handling it were unreachable and have been removed.
--- a/changelog/+ipaddress-attribute-kind.added.md
+++ b/changelog/+ipaddress-attribute-kind.added.md
@@ -1,1 +1,0 @@
-Added support for the new `IPAddress` attribute kind. Values are exposed as bare `ipaddress.IPv4Address`/`IPv6Address` objects (no prefix) and serialized to a bare-address string when writing, alongside the existing `IPHost` and `IPNetwork` kinds.
--- a/docs/_templates/sdk_compatibility.j2
+++ b/docs/_templates/sdk_compatibility.j2
@@ -57,7 +57,7 @@ Some SDK features require a minimum Infrahub version:
 
 ### Bare IP address attributes
 
-An `IPHost` attribute whose schema sets `allow_prefix: false` holds a bare address such as `10.0.0.1` rather than `10.0.0.1/32`. From SDK 1.23.0 onwards, reading such an attribute returns an `ipaddress.IPv4Address` or `ipaddress.IPv6Address`, and generated protocols type it as `IPAddress` instead of `IPHost`. An `IPHost` attribute that does not set `allow_prefix` is unaffected and still returns an `ipaddress.IPv4Interface` or `ipaddress.IPv6Interface`.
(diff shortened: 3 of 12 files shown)


ajtmccarty:
replaced by #1222

PR #1223, opened 2026-07-31
feat: report the schema fields the write contract does not apply

dgarros:
## What changed

The generated write models set `extra="ignore"`, so a field the user may not set never reaches the server. That settled whether such a field has an *effect*, but not whether the author hears about it — a misspelled key produced a schema quietly different from the one they wrote.

Every extra key in a submitted payload is now classified:

- a name the contract knows at that location but the user may not set is **dropped and reported as a warning**, so a schema read back from Infrahub, edited and loaded again keeps working;
- any other name is an **error**, since the only ways to produce one are a typo and a field that no longer exists.

```text
nodes[0].attributes[0].inherited: Read-only field, the submitted value is ignored (received: True)
nodes[0].attributes[0].optionl: Unknown field, it is not part of the schema (received: True)
```

## Implementation notes

**A generated table, not a third model family.** `infrahub_sdk/schema/generated/contract.py` holds the non-settable field names of each write class (17 entries). It is produced by the Infrahub-side generator from two diffs: the read variant of a class against its write variant, and the internal pydantic counterpart of a value model against its generated write model. The second diff matters — the parameters/choice/computed-attribute/extension models are hand-declared in the generator, so they omit the `id`/`state` bookkeeping every internal schema model carries, and each computed-attribute variant omits its siblings' fields. Those names all appear in a schema dumped from Infrahub's own models, so treating them as unknown would have broken the round trip.

**Applying the table needs to know which model governs each place in the payload.** `_collect_extra_fields` walks the raw payload alongside the *validated* write document: the document resolves the model at every location, including which member of a discriminated union an attribute matched, so raw keys are compared against the fields that location actually accepts. This is why the walk gives exact dotted paths and the owning kind/element for free, and it avoids both a second generated model family and a hand-rolled discriminator resolver.

One consequence, documented in the docstring: extra fields surface only once the payload is otherwise valid, because the validated document is what resolves the contract. A payload rejected for another reason names its extra fields on the next run.

## API changes

- `SchemaValidationResult` gains `warnings` (`list[SchemaValidationWarningDetail]`) and `warning_messages`. Each warning carries the dotted path, the bare field name, and the schema kind / element that set it.
- `client.schema.validate()` returns the verdict instead of `None`, and raises `ValueError` rather than a pydantic `ValidationError`. `ValidationError` subclasses `ValueError`, so `except ValueError` callers are unaffected; a caller catching `ValidationError` specifically needs updating.
- `validate_schema_content_and_exit()` no longer takes `client` — it never needed a server.
- `infrahubctl validate schema` no longer builds a client, so it works with no server configured. It reports warnings and errors; `infrahubctl schema load` / `schema check` report errors locally and leave warnings to the server response, which already carries them, to avoid printing each one twice.

## Two shapes that previously passed and now fail

Both were silently dropped before, so the loaded schema differed from the authored one:

1. a misspelled field name;
2. attribute `parameters` belonging to a different attribute `kind` — for example `start_range` on a `Number` attribute, which configured nothing.

## Testing

- `tests/unit/test_schema_offline_validation.py` restructured: read-only cases assert the exact set of warning paths, unknown-field cases assert the exact set of error paths. Covers every nesting level (root, node, generic, attribute, relationship, `parameters`, `choices`, `computed_attribute`, `extensions.nodes[*].attributes`), the value-model bookkeeping fields, the sibling-variant field, and the reporting-order consequence above.
- Full unit suite: 1735 passed. Three pre-existing failures in `test_repository_app.py` / `test_task_app.py` (rich table width mismatches) fail identically on a clean tree.
- `invoke format` and `invoke lint-code` (ruff, ty, mypy) clean.

## Companion PR

`contract.py` is generated by `invoke backend.generate` in the Infrahub repo. The Infrahub side, which generates it and consumes the warnings on `POST /api/schema/load`, is opsmill/infrahub#10095.

Note this branch is based on the commit the Infrahub `release-1.11` submodule currently pins, so it sits behind `infrahub-develop`; the changes to `ctl/schema.py` on develop are additive, in a different part of the file.


infrahub_sdk/ctl/validate.py
```diff
@@ -36,16 +37,16 @@ async def validate_schema(schema: Path, _: str = CONFIG_PARAM) -> None:
         console.print(f"[red]Unable to find {schema}")
         raise typer.Exit(1)
 
-    client = initialize_client()
```
ajtmccarty, line 39: nice to see the client removed here

infrahub_sdk/schema/validate.py
```diff
@@ -79,6 +111,112 @@ def _collect_validation_errors(
...
+) -> None:
+    """Report every payload key the write contract does not declare, walking the validated model.
+
+    The validated instance resolves the model that applies at each location -- including which
+    member of a discriminated union an attribute matched -- so the raw payload can be compared
+    against the fields that location actually accepts.
+    """
+    model = type(instance)
+    fields = model.model_fields
+    read_only = _read_only_fields(model)
```
ajtmccarty: I would think that `model.model_fields` would include all fields defined on the model including those inherited, in which case passing `fields` into `_read_only_fields` would let you skip the `__mro__` loop. but I could definitely be missing something

infrahub_sdk/schema/validate.py
```diff
@@ -87,23 +225,31 @@ def validate_schema(schema: dict[str, Any], *, raise_on_error: bool = False) ->
...
     errors: list[SchemaValidationErrorDetail] = []
+    warnings: list[SchemaValidationWarningDetail] = []
 
     try:
-        InfrahubSchemaWrite.model_validate(schema)
+        validated = InfrahubSchemaWrite.model_validate(schema)
     except PydanticValidationError as exc:
         _collect_validation_errors(exc=exc, errors=errors)
+    else:
+        _collect_extra_fields(payload=schema, instance=validated, errors=errors, warnings=warnings)
```
ajtmccarty, line 255: this looks like it works, but it is hard to follow and uses some patterns that aren't quite code smells, but are a little suspicious
- `__mro__` access
- a lot of `isinstance`, `getattr`, and `.get()` that are brittle

I have an idea for another approach that might be worth. I haven't tested it and could definitely be wrong about it.
1. first deserialize the request into the `Read` model, capturing any errors. I think this would cover the removed field and typo cases
2. transfer the `Read` instance to the corresponding `Write` model, capturing anything dropped/rejected as warnings

again, not sure if this would work, but it would be nice to let Pydantic do this work for us

ajtmccarty (commented):
one idea for a different approach while I review the tests and the other PR. there's also a cubic comment that I'm not really sure about

infrahub_sdk/schema/validate.py
dgarros: Fair on all three, and two of them are pushed in `c257f7c`.

`__mro__` is gone — the generator now resolves the parent chain when it emits the table, so each entry is complete per class and the lookup is a plain dict access. The defensive `isinstance` checks on the raw payload have collapsed into a single contract guard at the top of the walk; the ones that remain dispatch on the validated value's shape rather than second-guessing the input. Doing that surfaced a latent bug I'd otherwise have shipped: `_descend_context` reads keys off the raw value, so it had to move inside the guard instead of running before it.

I did try the read→write idea. It has a hole worth knowing about: the read root only declares `nodes` and `generics`, so with `extra="ignore"` step 1 silently eats `version` and the entire `extensions` block, and step 2 then fails `version: Field required` on every payload — with extension attributes never validated or loaded. Step 1 also only catches typos if the read models are `extra="forbid"`, and that breaks the read path, since `kind` is a *computed* field on read and a forbid-read model rejects the exact body `GET /api/schema` emits (`[('kind',)] ['extra_forbidden']`).

```
read root fields : ['generics', 'nodes']
write root fields: ['extensions', 'generics', 'nodes', 'version']
```

infrahub_sdk/schema/validate.py
```diff
@@ -79,6 +111,117 @@ def _collect_validation_errors(
...
+        # undeclared key, so there is nothing to compare and nothing below worth walking.
+        return
+
+    if field is not None:
+        kind, element, qualifier = _descend_context(
+            field=field, item=payload, kind=kind, element=element, qualifier=qualifier
+        )
+
+    fields = type(instance).model_fields
+    read_only = READ_ONLY_FIELDS.get(type(instance).__name__, frozenset())
```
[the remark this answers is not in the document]
dgarros: False positive — the premise is wrong. `TextAttributeParametersWrite` and `ListAttributeParametersWrite` are both in `READ_ONLY_FIELDS`, and were before this change too. From `contract.py` at the reviewed commit:

```
"AttributeParametersWrite": frozenset({"id", "state"}),
"ListAttributeParametersWrite": frozenset({"id", "state"}),
"NumberAttributeParametersWrite": frozenset({"id", "state"}),
"NumberPoolParametersWrite": frozenset({"id", "state"}),
"TextAttributeParametersWrite": frozenset({"id", "state"}),
```

Each parameters class is mapped to its own internal counterpart when the table is built, so it gets its own entry rather than relying on the base. The behaviour predicted to break does not:

```
Text        valid=True  warnings=['parameters.id', 'parameters.state']  errors=0
Number      valid=True  warnings=['parameters.id', 'parameters.state']  errors=0
List        valid=True  warnings=['parameters.id', 'parameters.state']  errors=0
NumberPool  valid=True  warnings=['parameters.id', 'parameters.state']  errors=0
Dropdown    valid=True  warnings=['parameters.id', 'parameters.state']  errors=0
```

The general concern behind it is reasonable, though: with a lookup by exact class name there is nothing to catch a generator that forgets a class. That is now covered by a test asserting no generated write class inherits a read-only field its own entry omits — it passes on the current table, and fails with `TextAttributeWrite: ['inherited']`, `NodeSchemaWrite: ['hash', 'kind']` and five more if the generator regresses to emitting own-fields-only.

Not taking the suggested MRO walk: it was removed deliberately in this commit, and resolving the inheritance once in the generator is what makes the lookup a plain dict access.

PR #1273, opened 2026-08-24
feat(protocols): support generating the SDK core protocols module [IFC-3054]

ogenstad:
# Why

## The short story

We generate the infrahub_sdk/protocols.py file from the Infrahub repo, the way we are generating this file is lacking and falling behind compared to how things are done when we generate these protocol type files from the SDK. The problem is that for our core schema we go blind whenever an internal kind has a relationship, i.e. we can't see what's behind that relationship. This PR is the first part of an effort to correct that so that it aligns with what the SDK itself does. The idea is to update the SDK generator so that it also works with the generation that needs to happen for our internal kinds, then a follow up PR in the Infrahub repository will replace the current generator and use this code instead. During that phase the infrahub_sdk/protocols.py file will also be updated.

As it is now the lack of this generates a lot of typing ignores within the Infrahub code base.

## Longer version (AI)

The code generator behind `infrahubctl protocols` can only render protocols for a user's own
schema, importing the core kinds it references from `infrahub_sdk.protocols`. It cannot render
that module itself, which is why `infrahub_sdk/protocols.py` is instead generated from a separate
Jinja template that lives in the Infrahub repository. That second template has drifted: it emits
bare `RelatedNode` / `RelationshipManager` with no peer type, so the protocols we ship lose the
peer of every relationship, while this generator has emitted `RelationshipManager[Peer]` and
`RelationshipAttribute[Peer]` for a while now.

Goal: teach this generator to render the core module too, so there is one renderer and the two
cannot drift again.

Non-goals: this PR does not regenerate `infrahub_sdk/protocols.py`. Producing it needs the
matching change on the Infrahub side, and it is a ~2,300-line generated diff that deserves its own
review. It follows in a second PR, with its own changelog entry.

Part of IFC-3054.

## What changed

Behavioral changes, for users:

- A hierarchical kind no longer declares `parent` and `children` twice. When the schema already
  exposes them as relationships, which is the case for anything read from the API, they were
  emitted once from the relationship list and again from the hierarchy, with the second pair
  overriding the first. This is the only change to what `infrahubctl protocols` writes.

Implementation notes:

- `ProtocolTarget` selects between `USER_SCHEMA` (the default, so every existing caller is
  unaffected) and `SDK_CORE`.
- The per-file header moved out of `template.j2` into `header.j2`, so the core module can render
  one header followed by two bodies. The user branch of that header is byte-identical to before.
- Two name sets decide which referenced classes switch to a `Sync` counterpart. They are separate
  because the two positions already resolved names differently for a user schema: a peer may be
  any core kind, while an inheritance list only ever switches the three names in
  `CORE_BASE_CLASS_TO_SYNCIFY`. Generating the core module makes both sets the same, since every
  class in it is local. That asymmetry is preserved rather than fixed, to keep user output stable.

What stayed the same:

- `infrahubctl protocols` output, apart from the duplicate lines above. This is enforced, not
  claimed: `tests/fixtures/protocols_generator/` pins the full rendered file for both variants.
  The fixtures were captured from the generator *before* any change and the test confirmed green
  on the untouched code, so the two removed lines are the entire diff.
- The public API. `CodeGenerator(schema=...)` and `render(sync=...)` are unchanged.

## How to review

Suggested order:

1. `target.py` and `generator.py` for the mechanism.
2. `template.j2`, which is only the header being removed, `{{ suffix }}` on the four class-name
   lines, and the hierarchical guard.
3. `header.j2`, where the user branch should match the block deleted from `template.j2`.
4. The tests, with the golden fixtures being generated output rather than hand-written.

The part worth extra scrutiny is the hierarchical guard, since it is the one behavioural change.
I chose to fix it here rather than leave it because the regenerated core module would otherwise
ship duplicated annotations for `CoreGroup`, `BuiltinIPPrefix`, `CoreMenu` and others.

## How to test

```bash
uv run pytest tests/unit/sdk/test_protocols_generator.py
```

If `test_repository_app`, `test_task_app::test_task_list_command` or
`test_config::test_missing_password` fail for you locally, that is environment leakage rather than
this branch: they fail the same way on a clean `origin/infrahub-develop` checkout, and CI is green
on all five Python versions.

## Impact & rollout

- **Backward compatibility:** no API change. Generated output changes only by dropping the
  duplicate `parent`/`children` pair described above.
- **Deployment notes:** safe to merge on its own. The submodule pointer bump in Infrahub waits on
  the follow-up PR that regenerates `infrahub_sdk/protocols.py`.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added


ogenstad:
Thanks, the finding is valid and is fixed in 516f7352.

The guard checked only `parent` before deciding to emit the pair, so a schema declaring `children` but not `parent` would have got a duplicate `children`, and one declaring `parent` but not `children` would have lost `children` entirely. Each name is now checked independently.

One clarification for anyone reading later: only generics carry `hierarchical` (`NodeSchemaAPI` has `hierarchy` instead, and profiles and templates have neither), so the equivalent blocks in the node and profile loops cannot fire today. I applied the same guard to all three copies anyway rather than making a reader work out which one is live.

Added `test_hierarchy_members_are_declared_once`, which covers all four combinations of the pair being declared as relationships and asserts each is emitted exactly once.

gmazoyer (approved):

PR #1277, opened 2026-08-25
fix(packaging): correct declared dependencies and publish the tests extra

ogenstad:
# Why

The unit-test matrix runs five Python versions but every job installs exactly what `uv.lock` pins, so the version ranges declared in `pyproject.toml` are never exercised. Checking them by hand turned up several that were wrong, two of which break users outright.

**Goal:** make the declared dependency surface match what the SDK actually needs, and add tests so it stays that way.

**Non-goals:** no CI jobs that resolve at the declared lower and upper bounds. That is the follow-up work, and it is what will genuinely test the ranges end to end. This PR only covers what can be checked without extra CI time.

Ref IHS-224 (part 1 of 3).

## What changed

Behavioral changes:

- `pip install 'infrahub-sdk[tests]'` works. The extra is documented in the installation guide but was never published, so the command warned that no such extra existed and installed nothing beyond the base package.
- Installing against pydantic 2.0 or 2.0.2 now fails while resolving instead of succeeding and then raising `SchemaError` on `import infrahub_sdk`. Those versions reject the `\_` escape in the generated schema model patterns, so they never worked.
- `anyio`, `typing-extensions` and `packaging` are declared directly. All three were imported by the shipped package but only arrived as transitive dependencies, so a constrained resolution could install the SDK unusable.
- `infrahub-sdk[ctl]` no longer installs `numpy` or `mdxify`, neither of which the SDK imports.
- `infrahub-sdk[all]` now covers `ctl` and `tests` together, so it pulls in considerably more than before.

Implementation notes:

- Every floor was established by installing the candidate version and exercising the import, not read off release notes. Two initial guesses were wrong: `anyio.Path` arrives in 3.3.0 rather than 3.0, and `typing-extensions` needs 4.4.0 for PEP 696 `TypeVar` defaults.
- `all` is now `["infrahub-sdk[ctl,tests]"]`. Hand-duplicating the lists is how it had already lost `mdxify`, and a self-reference cannot drift.
- `mdxify` moves to a new `docs` dependency group, included in `dev`, so `uv sync --all-groups` still provides it for `invoke docs-generate`.

What stayed the same: no runtime code changed. This is packaging metadata, a regenerated lock, one docs note, and a new test module.

## How to review

Start with the `pyproject.toml` diff, which is the whole substance of the change. Then `tests/unit/test_packaging_metadata.py`. The `uv.lock` diff is mechanical, and mostly shrinkage from dropping `numpy` and `mdxify`.

Two things worth extra scrutiny:

- **The `tests` extra shares its name with the `tests` dependency group** while holding different contents. The name was chosen to match what the docs and README already tell people to type. uv handles the two namespaces without complaint, but the follow-up PR that splits the test groups is the natural place to rename the group.
- **The `tests` extra is heavy**, adding roughly 66 packages including Docker, FastAPI, uvicorn and Prefect client libraries, all via `infrahub-testcontainers`. Someone who only wants the pytest plugin gets all of it. Splitting the container-based helpers into their own extra is worth considering in the follow-up.

## How to test

```bash
uv run pytest tests/unit/test_packaging_metadata.py
uv run invoke lint-code
uv lock --locked --offline
```

The new tests are non-vacuous: pointing them at the previous `pyproject.toml` fails four of the five checks, each naming a real defect. The `requires-python` check passes on the old metadata too, so it is a regression guard rather than a bug finder.

Verified locally: `ruff`, `ty` and `mypy` clean; `docs-validate` exits 0 with no committed docs changed; `lint-docs` byte-identical to the base branch. The unit suite is 1840 passed with 2 failures, both of which reproduce identically in a pristine worktree of the base commit (macOS-only Rich wrapping of long `/private/var/folders/...` temp paths, in files this PR does not touch).

## Impact & rollout

- **Backward compatibility:** raising the pydantic floor and dropping `numpy`/`mdxify` from the extras are resolver-visible narrowings. Nothing that worked before stops working, since the removed versions could not import the SDK and the removed packages were never used, but the install-time behavior changes. This targets `infrahub-develop` deliberately so it ships with the next Infrahub version rather than as a patch to the current SDK line.
- **Performance:** no runtime impact.
- **Deployment notes:** safe to merge independently. The follow-up CI work depends on this landing first.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added
- [x] External docs updated (if user-facing or ops-facing change)

---
## Summary by cubic
Fixes the SDK packaging metadata so plain installs include the dependencies required by shipped modules and `infrahub-sdk[tests]` installs its documented test tools. It also rejects unsupported `pydantic` 2.0.x versions during resolution instead of failing on import. Addresses IHS-224.

**Packaging**
- Moves Jinja2, PyYAML, rich, anyio, and typing-extensions into core dependencies and removes rich’s upper bound.
- Adds the `tests` extra with `infrahub-testcontainers`, pytest, and packaging.
- Makes `all` aggregate `ctl` and `tests`, removes unused `numpy` and `mdxify` from `ctl`, and moves `mdxify` to the `docs` dependency group.
- Plain installs grow from 19 to 26 packages; `tests` adds roughly 61 packages, and projects pinning `pydantic` 2.0–2.0.2 must upgrade.

**Validation**
- Adds packaging tests for dependency lower bounds, extras aggregation, import coverage, and Python classifiers.
- The import checks distinguish load-time, deferred, and type-checking imports and handle `tomllib` on Python 3.10.
- Updates the installation guide, changelog, and lockfile.

<sup>Written for commit aa5dcc79ab369df30dd8e1ca29d7d9077f992cdf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1277?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ogenstad:
## Notes on the latest findings

Recording two open items against `78faca97` so they are not lost. Neither is fixed yet.

### 1. The import scan counts `if TYPE_CHECKING:` blocks as import-time

`_import_time_nodes` stops descending at function bodies, but it unwinds every module-level `if` without looking at the guard. So an import under `if TYPE_CHECKING:` is treated as executing at load time, when under `from __future__ import annotations` it never runs at all.

Confirmed **latent, not currently firing**: no base module type-checks against an extra-only package today, which is why the suite is green. The exposure is real though, and it is a trap rather than a cosmetic issue: the natural way to type-annotate against a `ctl`-only package in a base module is exactly a `TYPE_CHECKING` import, and `test_shipped_modules_only_import_what_their_install_provides` would reject it as a runtime dependency. A contributor doing the correct thing gets a confusing failure.

Fix is to skip descending into `If` nodes whose test is the `TYPE_CHECKING` name. Worth doing before this merges, since the whole point of the check is to guide people to the right pattern.

### 2. The `rich<14` cap is probably unjustified, and this PR widens its reach

Moving `rich` into the core dependencies also promotes its `<14` cap to every plain install, not just `ctl`/`all` users. Anyone already on rich 14 would hit a resolution conflict. That is a genuine consequence of the change and worth being deliberate about.

What I checked:

- **The private API is not what the cap protects.** `Traceback._guess_lexer`, used by `jinja2.py:17` and `template/__init__.py:228`, works unchanged on rich 13.9.4, 14.0.0 and 14.2.0.
- **The suite does not care.** With `rich==14.2.0` overlaid on the project environment, the full unit suite is byte-identical to the current result: 1841 passed, with only the two pre-existing macOS temp-path failures.
- **The cap has no recorded cause.** It entered in 18677645 ("Convert project to UV"), a mechanical conversion, and rich 14.0 predates that commit. It reads as carried-over pinning rather than a response to a known break.

So the evidence points to the cap being safe to drop, and dropping it would remove the conflict this PR would otherwise impose on base installs. Holding off because lifting a version cap is a separate judgement call from fixing the dependency surface, and it deserves its own decision rather than riding along here.

Two options, in preference order: drop `<14` in this PR since this PR is what makes the cap load-bearing for every user; or keep it and lift it deliberately in the follow-up upper-bound CI work, which is designed to report exactly this ("a cap that can now be lifted").

### Related, still unticketed

Both `rich` call sites depend on `Traceback._guess_lexer`, a private API, now sitting under core rather than optional functionality. It works across 13.x and 14.x, but nothing stops it moving in any minor release. Worth replacing with a supported call independently of the cap question.

gmazoyer (approved):
Looks good, just wonder how all those multi-line changelog record will look like once combined in a release note.

PR #1291, opened 2026-08-27
fix: run infrahubctl generator on target kinds without a name attribute

### LINKED ISSUE #1289: bug: infrahubctl generator fails on target nodes that have no name attribute
### Component

infrahubctl

### Infrahub SDK version

1.23.0 (current latest on PyPI)

### Current Behavior

`infrahubctl generator <name>` exits immediately with:

```
Error: The node doesn't have an attribute for name
```

when the generator definition uses `targets` and the members of that group are of a kind that has no `name` attribute. The generator itself never runs, `generate()` is not called.

The CLI iterates over the group members and builds the params dict with a hardcoded `name` attribute, `infrahub_sdk/ctl/generator.py:91`:

```python
for member in targets._get_relationship_many(name="members").peers:
    check_parameter = {}
    if identifier:
        attribute = getattr(member.peer, identifier)
        check_parameter = {identifier: attribute.value}
    params = {"name": member.peer._get_attribute(name="name").value}
```

`_get_attribute` raises `ResourceNotDefinedError` when the attribute does not exist, `infrahub_sdk/node/node.py:748`. Not every kind has a `name`. In the customer schema the target kind is a firewall rule whose `human_friendly_id` and `display_label` are both `rule_id__value`, and its only attributes are `rule_id`, `duration`, `changeid` and `status`.

The query variables are not affected by this, they come from `check_parameter`, which is built correctly from the generator definition parameters. Only `params` on the generator instance needs the value.

Server side execution does not go through this code path, so the same generator works in a proposed change and fails only from the CLI.

### Expected Behavior

`infrahubctl generator <name>` runs the generator for each member of the target group, whatever attributes the target kind happens to have.

### Steps to Reproduce

1. Define a kind with no `name` attribute, for example:

```yaml
- name: Rules
  namespace: Firewall
  human_friendly_id: ["rule_id__value"]
  display_label: "{{ rule_id__value }}"
  attributes:
    - name: rule_id
      kind: Number
    - name: status
      kind: Text
```

2. Create a few nodes of that kind and add them to a `CoreStandardGroup` called `pending-firewall-rules`
3. Define a generator in `.infrahub.yml` targeting that group:

```yaml
generator_definitions:
  - name: process_pending_rules
    class_name: ProcessPendingRules
    file_path: generators/pending_rules/process_pending.py
    query: pending_rules_context
    targets: pending-firewall-rules
    parameters:
      rule_id: rule_id__value
```

4. Run `infrahubctl generator process_pending_rules`

### Additional Information

Workaround for now is to pass the variable explicitly, which takes the other branch in `run()` and skips the group iteration:

```
infrahubctl generator process_pending_rules rule_id=12345
```

That only runs one target per invocation, so it does not replace the group behaviour.

Reported by a customer running SDK 1.23.0 on Python 3.12.13.

pthmas:
# Why

`infrahubctl generator <generator>` exits immediately with `Error: The node doesn't have an attribute for name` when the generator's `targets` group holds members of a kind that has no `name` attribute. The generator never runs. The same generator works server-side in a proposed change, because the CLI is the only path that builds its per-target values this way. Reported by a customer whose target kind is a firewall rule identified by `rule_id`.

Goal: the CLI runs the generator once per group member whatever attributes the target kind happens to have, deriving the per-target values the same way Infrahub does.

Closes #1289

## What changed

Behavioral changes:

- Target kinds without a `name` attribute no longer abort the run.
- Every declared parameter now reaches the query, not just the first one.
- A CLI run and a server run of the same generator against the same target now share a tracking group, so each cleans up the other's stale nodes.

Implementation notes: the per-target values now come from `member.peer.extract(params=generator_config.parameters)` — the same call Infrahub makes at `backend/infrahub/generators/tasks.py:246`. That resolves each parameter's declared path (`rule_id: rule_id__value`) rather than reading the parameter key as an attribute name, which only worked when the two coincided. For `parameters: {name: "name__value"}` the result is identical to before, so existing tracking groups keep their names.

What stayed the same: no CLI signature change, no schema change, server-side execution untouched.

## How to review

`infrahub_sdk/ctl/generator.py` — the group loop now resolves the parameters mapping in one call, and the hand-rolled first-key lookup above it is gone. The resolved dict serves as both the query variables and the generator params, as it does server-side.

## How to test

```bash
uv run pytest tests/unit/ctl/test_generator.py
```

Manual: define a kind with no `name` attribute, add nodes to a `CoreStandardGroup`, point a generator definition's `targets` at it with a parameter on one of that kind's own attributes, and run `infrahubctl generator <generator>` with no variables. Before: exits with `The node doesn't have an attribute for name`. After: one run per member.

## Impact & rollout

- **Backward compatibility:** for `parameters: {name: "name__value"}`, the resolved values are identical to before — same tracking group names.
- **Deployment notes:** safe to release.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added
- [ ] External docs updated — not needed, no interface change
- [ ] Internal .md docs updated — not needed


ogenstad (commented):
For a change like this we also need to ensure that we have the same behavior within Infrahub itself. I.e. when a generator runs within Infrahub. Does this already work as expected when the generator runs inside Infrahub or was this validated?

pthmas:
@ogenstad  the server never had this bug: it resolves each target’s variables with `member.extract(params=...)` and never touches a name attribute. Comparing the two did show a drift: the CLI hardcoded {"name": ...} as the tracking-group key and only used the first parameter, so for {rule_id: rule_id__value} a CLI run and a server run landed in different tracking groups. Reworked the PR to use extract() in the CLI too. Should be same behavior on both paths now.

ogenstad (approved):

Issue #1275, opened 2026-08-25
bug: object get silently omits Generic-kind many relationships from its output


lancamat1:
### Component

infrahubctl

### Infrahub SDK version

1.23.0

### Current Behavior

`infrahubctl object get <kind>` does not include cardinality-many relationships of kind `Generic` in its output — in any output format, even though list mode queries with `prefetch_relationships=True`. The SDK's query generation only includes many relationships of kind `Attribute` or `Parent` by default, and the `object get` command never passes `include` for the others.

Example with a `TestingPerson` node whose `animals` relationship (cardinality many, kind `Generic`) has three peers:

```yaml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: TestingPerson
  data:
  - name: Ethan Carter
    height: 185
```

The `animals` relationship is missing entirely — there is no indication in the output or on stderr that data was skipped.

Note: this is about the relationship *kind* `Generic` (the default kind for a relationship in the schema), not about the peer being a generic schema. The peers here are real, instantiated nodes (`TestingCat`/`TestingDog` inheriting `TestingAnimal`).

### Expected Behavior

An export intended to be reloadable (`object get --output yaml > backup.yml`, as shown in the command's own help text) should carry the object's relationships, or at minimum make it visible that some relationships were not exported. Silent omission means a get → load round-trip quietly loses relationship data.

### Steps to Reproduce

1. Load a schema where a node kind has a cardinality-many relationship of kind `Generic` (e.g. `TestingPerson.animals` from the SDK's testing schemas).
2. Create a person and several animals owned by that person.
3. `infrahubctl object get TestingPerson --output yaml`
4. The output contains only attributes; `animals` is absent.

Reproduced end-to-end with `infrahub-testcontainers`.

### Additional Information

- Related: #1274 (many relationships that *are* exported use a format `object load` rejects). Together these two issues currently prevent a faithful get → load round-trip.
- Origin of the behavior: `generate_query_data_node` (`infrahub_sdk/node/node.py`) skips many relationships whose kind is not `Attribute`/`Parent` unless they are explicitly listed in `include`; `object get` (`infrahub_sdk/ctl/object/get.py`) doesn't pass `include`. The SDK default is likely a deliberate performance choice (many-rels can be huge), so the gap is in how `object get` uses it.

PR #1306, opened 2026-08-31
fix: emit loadable YAML for cardinality-many relationships

### LINKED ISSUE #1274: bug: object get YAML output is not loadable by object load (many relationships exported as invalid nested data dict)
### Component

infrahubctl

### Infrahub SDK version

1.23.0

### Current Behavior

The YAML produced by `infrahubctl object get <kind> --output yaml` cannot be fed back into `infrahubctl object load`.

For cardinality-many relationships, the YAML formatter exports peer references wrapped in a nested dict:

```yaml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: TestingGadget
  data:
  - name: multitool
    tags:
      data:
      - red
      - blue
```

`object load` accepts two formats for a many relationship: a plain list of HFIDs, or a `data:` list of dicts describing nested objects. A `data:` list of HFID strings matches neither, so loading fails validation:

```text
1.tags: Relationship tags doesn't have the right format many / <class 'dict'>
```

Any kind with a populated `Attribute`-kind many relationship (for example `tags`) is affected, because those relationships are fetched by default by `object get` and always end up in the exported YAML.

### Expected Behavior

The output of `object get --output yaml` should be directly loadable by `object load`, e.g. `infrahubctl object get InfraDevice --output yaml > backup.yml && infrahubctl object load backup.yml` should work. The `object get` help text even shows the `> backup.yml` workflow, and the formatter docstring states the output is round-trippable.

### Steps to Reproduce

1. Load a schema with a many relationship of kind `Attribute` (e.g. a node with `tags` pointing to `BuiltinTag`).
2. Create two tags and one object referencing them.
3. `infrahubctl object get TestingGadget --output yaml > gadgets.yml`
4. `infrahubctl object load gadgets.yml` → fails with the error above.

Reproduced end-to-end with `infrahub-testcontainers`; the failing round-trip is covered by an integration test (`object get` → file → `object load`).

### Additional Information

- Origin: `YamlFormatter._node_to_data_entry` (`infrahub_sdk/ctl/formatters/yaml.py`, the many-relationship branch) emits `{"data": [<hfids>]}`; `get_relationship_info` (`infrahub_sdk/spec/object.py`) leaves such a value's format `UNKNOWN` and `validate_related_nodes` rejects it.
- Related but distinct: many relationships of kind `Generic` are silently omitted from `object get` output entirely (the SDK only fetches `Attribute`/`Parent`-kind many relationships by default), so those never round-trip. Filed separately.

rootsec1:
Fixes #1274.

`object get --output yaml` wraps populated cardinality-many HFID references in a `data` dictionary, which the loader reserves for nested objects and therefore rejects when it contains a string list. Emit the already-supported plain HFID list instead and cover the formatter-to-loader path with a regression test.

Cardinality-one output, empty-relationship omission, and multi-component HFIDs are unchanged. Generic-kind relationship omission is excluded and remains tracked in #1275.

Tested with:
- `uv run pytest -q tests/unit/ctl/formatters/test_yaml.py tests/unit/sdk/spec/test_object.py` (49 passed)
- `COLUMNS=140 uv run pytest -q tests/unit` (1819 passed, 1 xfailed)
- `uv run invoke format`
- `uv run invoke lint-code`
- `uv run invoke docs-validate`
- `uv run invoke lint-docs`
- `uv run towncrier build --draft --version 1.23.2`

### DIFF
--- a/infrahub_sdk/ctl/formatters/yaml.py
+++ b/infrahub_sdk/ctl/formatters/yaml.py
@@ -100,7 +100,7 @@ class YamlFormatter:
                 peers = getattr(rel, "peers", None) or []
                 refs = [r for p in peers if (r := _related_node_ref(p)) is not None]
                 if refs:
-                    entry[rel_name] = {"data": refs}
+                    entry[rel_name] = refs
 
         return entry
 
--- a/tests/integration/test_enduser_cli.py
+++ b/tests/integration/test_enduser_cli.py
@@ -7,9 +7,11 @@ the existing ``test_infrahubctl.py`` integration tests.
 
 from __future__ import annotations
 
+import asyncio
 import json
 import os
 from typing import TYPE_CHECKING
+from uuid import uuid4
 
 import pytest
 import yaml
@@ -23,6 +25,8 @@ from infrahub_sdk.testing.schemas.animal import SchemaAnimal
 
 if TYPE_CHECKING:
     from collections.abc import Generator
+    from pathlib import Path
+    from typing import Any
 
     from infrahub_sdk import InfrahubClient
     from infrahub_sdk.node import InfrahubNode
@@ -207,6 +211,46 @@ class TestEnduserCliWrite(_EnduserCliBase):
         """Create without --set or --file fails."""
         result = runner.invoke(app, ["object", "create", "TestingPerson"])
         assert result.exit_code != 0
+
+    async def test_get_yaml_round_trips_attribute_many_relationship(
+        self,
+        base_dataset: None,
+        client: InfrahubClient,
+        schema_extension_01: dict[str, Any],
+        tmp_path: Path,
+    ) -> None:
+        """Round-trip cardinality-many attribute relationships through the CLI."""
+        response = await client.schema.load(schemas=[schema_extension_01], wait_until_converged=True)
+        assert not response.errors
+
+        suffix = uuid4().hex
+        tag_names = [f"yaml-round-trip-{suffix}-one", f"yaml-round-trip-{suffix}-two"]
+        tags = []
+        for tag_name in tag_names:
+            tag = await client.create(kind="BuiltinTag", name=tag_name)
+            await tag.save()
+            tags.append(tag)
+
+        rack_name = f"yaml-round-trip-rack-{suffix}"
+        rack = await client.create(kind="InfraRack", name=rack_name, tags=tags)
+        await rack.save()
+
+        get_result = await asyncio.to_thread(
+            runner.invoke,
+            app,
+            ["object", "get", "InfraRack", "--filter", f"name__value={rack_name}", "--output", "yaml"],
+        )
+        assert get_result.exit_code == 0, f"object get failed: {get_result.output}"
+
+        object_file = tmp_path / "infra-rack.yaml"
+        object_file.write_text(get_result.stdout, encoding="utf-8")
+        load_result = await asyncio.to_thread(runner.invoke, app, ["object", "load", str(object_file)])
+        assert load_result.exit_code == 0, f"object load failed: {load_result.output}"
+
+        fetched_rack = await client.get(kind="InfraRack", id=rack.id)
+        fetched_tags = fetched_rack._get_relationship_many(name="tags")
+        await fetched_tags.fetch()
+        assert sorted(peer.hfid or [] for peer in fetched_tags.peers) == sorted([[name] for name in tag_names])
 
     def test_update_inline(self, base_dataset: None) -> None:
         """Update a person's height using --set."""
--- a/tests/integration/test_enduser_cli.py
+++ b/tests/integration/test_enduser_cli.py
@@ -220,37 +220,45 @@ class TestEnduserCliWrite(_EnduserCliBase):
         tmp_path: Path,
     ) -> None:
         """Round-trip cardinality-many attribute relationships through the CLI."""
-        response = await client.schema.load(schemas=[schema_extension_01], wait_until_converged=True)
-        assert not response.errors
+        tags: list[InfrahubNode] = []
+        rack: InfrahubNode | None = None
+        try:
+            response = await client.schema.load(schemas=[schema_extension_01], wait_until_converged=True)
+            assert not response.errors
 
-        suffix = uuid4().hex
-        tag_names = [f"yaml-round-trip-{suffix}-one", f"yaml-round-trip-{suffix}-two"]
-        tags = []
-        for tag_name in tag_names:
-            tag = await client.create(kind="BuiltinTag", name=tag_name)
-            await tag.save()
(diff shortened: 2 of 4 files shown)


infrahub_sdk/ctl/formatters/yaml.py
```diff
@@ -100,7 +100,7 @@ def _node_to_data_entry(
                 peers = getattr(rel, "peers", None) or []
                 refs = [r for p in peers if (r := _related_node_ref(p)) is not None]
                 if refs:
-                    entry[rel_name] = {"data": refs}
+                    entry[rel_name] = refs
```
gmazoyer, line 103: Seems like a correct a minimal fix that describe what our documentation actually says.

infrahub_sdk/ctl/formatters/yaml.py
gmazoyer, line 5: The module docstring says the output is round-trippable with `infrahubctl object load`. With #1275 open, many relationships of kind `Generic` are still dropped from the output (infrahub_sdk/node/node.py:1507-1508 skips them). A loaded document silently loses those. Not introduced by this PR though.

gmazoyer (commented):
The unit test proves the formatter output reaches MANY_REF and passes validation. It stops there. Nothing runs the real object get `--output yaml` then object load path against a server. `test_get_list_yaml` in t`ests/integration/test_enduser_cli.py` only parses the YAML, and I found no object load call anywhere in `tests/integration/`. Issue #1274 says the failing round trip is covered by an integration test, but I could not find one in this repo. Maybe we should consider to add one.

tests/unit/ctl/formatters/test_yaml.py
```diff
@@ -327,7 +329,48 @@ def test_rel_cardinality_many_with_peers_uses_hfid(self) -> None:
...
+        errors = await InfrahubObjectFileData.validate_related_nodes(
+            client=client,
+            position=[1, "tags"],
+            rel_info=rel_info,
+            data=relationship_value,
+        )
+
+        assert rel_info.format == RelationshipDataFormat.MANY_REF
+        assert rel_info.is_reference
+        assert errors == []
```
polmichel, line 367: We try to avoid `MagicMock()` and similar mocking libraries in our unit tests. We recently added this rule to our LLM coding rules: `.agents/rules/python-testing.md.`

The reason is that we'd rather have a less isolated and slower test than one that looks like it protects against regressions but doesn't, because the mocks no longer match the real code.

Existing tests such `tests.unit.sdk.spec.test_object.test_validate_object` use `InfrahubClient()` and the `client.schema.set_cache()` method to do so.

polmichel (commented):
This mocking pattern does exist in other tests in the project, but we're trying to move away from it. Let me know if it makes sense! 

Thanks for the contribution

tests/unit/ctl/formatters/test_yaml.py
rootsec1: Updated in 49feedc: I replaced the mock-based regression setup with a real `InfrahubClient` and `client.schema.set_cache()`, following the existing project pattern. The formatter tests pass 21/21, and format, Ruff, ty, and mypy are all clean. Thanks for the guidance!

tests/integration/test_enduser_cli.py
```diff
@@ -208,6 +212,46 @@ def test_create_missing_args(self, base_dataset: None) -> None:
...
+        suffix = uuid4().hex
+        tag_names = [f"yaml-round-trip-{suffix}-one", f"yaml-round-trip-{suffix}-two"]
+        tags = []
+        for tag_name in tag_names:
+            tag = await client.create(kind="BuiltinTag", name=tag_name)
+            await tag.save()
+            tags.append(tag)
+
+        rack_name = f"yaml-round-trip-rack-{suffix}"
+        rack = await client.create(kind="InfraRack", name=rack_name, tags=tags)
```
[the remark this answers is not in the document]
rootsec1: Good catch — the test now deletes the rack and both tags in a `finally` block, so cleanup also runs when the get/load round trip fails. The schema extension remains scoped to the disposable class-owned Infrahub instance.

tests/integration/test_enduser_cli.py
```diff
@@ -208,6 +212,54 @@ def test_create_missing_args(self, base_dataset: None) -> None:
...
+            load_result = await asyncio.to_thread(runner.invoke, app, ["object", "load", str(object_file)])
+            assert load_result.exit_code == 0, f"object load failed: {load_result.output}"
+
+            fetched_rack = await client.get(kind="InfraRack", id=rack.id)
+            fetched_tags = fetched_rack._get_relationship_many(name="tags")
+            await fetched_tags.fetch()
+            assert sorted(peer.hfid or [] for peer in fetched_tags.peers) == sorted([[name] for name in tag_names])
+        finally:
+            if rack is not None:
+                await rack.delete()
```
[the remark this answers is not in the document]
rootsec1: Fixed in 8dc0961. Each resource deletion is now attempted independently, cleanup failures are suppressed when the test body already failed so the primary regression remains visible, and the first cleanup failure is raised only when the test body succeeded. Format, Ruff, ty, mypy, and all 21 YAML formatter tests pass.

tests/integration/test_enduser_cli.py
```diff
@@ -208,6 +212,65 @@ def test_create_missing_args(self, base_dataset: None) -> None:
...
+        base_dataset: None,
+        client: InfrahubClient,
+        schema_extension_01: dict[str, Any],
+        tmp_path: Path,
+    ) -> None:
+        """Round-trip cardinality-many attribute relationships through the CLI."""
+        tags: list[InfrahubNode] = []
+        rack: InfrahubNode | None = None
+        body_succeeded = False
+        try:
```
polmichel, line 226: The try/finally block in this test hurts readability. The test should only be about what we expect from the system. I think using fixtures with a `yield` structure would let us set up and clean up the data without making the test hard to read.
Let me know if that makes sense!

polmichel: For instance something like this could to the trick. Fyi I have not tested it locally.

```
     @pytest.fixture(scope="class")
      async def rack_with_tags(self, client):
          await client.schema.load(schemas=[schema_extension_01], wait_until_converged=True)
          suffix = uuid4().hex
          tag_names = [f"yaml-round-trip-{suffix}-one", f"yaml-round-trip-{suffix}-two"]
          tags = []
          for name in tag_names:
              tag = await client.create(kind="BuiltinTag", name=name)
              await tag.save()
              tags.append(tag)
          rack = await client.create(kind="InfraRack", name=f"rack-{suffix}", tags=tags)
          await rack.save()
          yield rack, tag_names
          await rack.delete()
          for tag in reversed(tags):
              await tag.delete()
```

PR #1307, opened 2026-09-01
Add the refresh tokens relationship to the account protocols

### LINKED ISSUE #10304: bug: deleting an SSO user's account leaves its external identity behind, permanently locking the user out of SSO login
### Component

API Server / GraphQL

### Infrahub version

1.10.x (reproduced on `stable` @ `982a6c018`)

### Current Behavior

An SSO user (reported with Keycloak / OIDC) whose Infrahub account is deleted can never sign in again. Every subsequent login attempt fails with `(0) Failed to complete sign-in. Please try again.` in the UI and an unhandled 500 on the server.

Deleting the `CoreAccount` does not remove the `InternalExternalIdentity` node that was created alongside it, so the identity survives with a dangling `account` relationship. The next login still matches that orphan on `(sub, provider_name, protocol)`, takes the "returning user" path, and dereferences the missing peer:

```
File "/source/backend/infrahub/auth/auth.py", line 238, in _account_from_existing_identity
    account = await identity_node.get_relationship(name="account").get_peer(
        db=db, peer_type=CoreAccount, raise_on_error=True
    )
File "/source/backend/infrahub/core/relationship/model.py", line 1029, in get_peer
    raise LookupError("Unable to find the peer")
LookupError: Unable to find the peer
```

`LookupError` is not an `InfrahubError`, so nothing converts it into an API error — it escapes the endpoint as a 500.

The user is locked out permanently. The only workaround found by the reporter is to delete the user in the identity provider and recreate them, which issues a fresh `sub` that no longer matches the orphaned identity.

### Expected Behavior

Deleting an SSO user's account in Infrahub should return them to a clean slate: the next SSO login creates a new account, exactly as a first-time login does. Whatever the state of the data, a login must never surface as an unhandled 500.

### Steps to Reproduce

1. Configure SSO (OIDC, e.g. Keycloak) and sign in as a user, letting Infrahub create the account.
2. Delete that user's account in Infrahub.
3. Sign in again through SSO as the same provider user.
4. The UI shows `(0) Failed to complete sign-in. Please try again.` and the server logs the `LookupError: Unable to find the peer` traceback above.

### Additional Information

Reproduced against a live Neo4j:

```
DELETED: [('CoreAccount', '18cc9327-…')]     # succeeds, identity not removed
REMAINING IDENTITIES: 1
SECOND LOGIN RAISED: LookupError Unable to find the peer
```

Two separate defects meet here:

- **The account delete should never have left the identity behind.** `InternalExternalIdentity.account` is `optional=False`, so `NodeDeleteValidator` classifies it as a mandatory dependent and ought to refuse the delete. It does not, because the query it relies on unconditionally excludes the `Internal` namespace (`backend/infrahub/core/query/relationship.py:1156`, `NOT src.namespace IN $excluded_namespaces`), so `_analyze_delete_dependencies` (`backend/infrahub/core/node/delete_validator.py:145`) sees no dependents. Disabling that exclusion and repeating the delete raises the expected error:

  ```
  ValidationError: Cannot delete CoreAccount '...'. It is linked to mandatory relationship account
  on node InternalRefreshToken '...', Cannot delete CoreAccount '...'. It is linked to mandatory
  relationship account on node InternalExternalIdentity '...'
  ```

  The same gap orphans `InternalAccountToken` and `InternalRefreshToken` rows on any account delete. Note it cannot be closed by lifting the exclusion alone: the guard would then make any account that has ever logged in undeletable, since its refresh token is also a mandatory dependent.

- **Independently, `signin_sso_account` assumes an identity node always has a live account peer** (`backend/infrahub/auth/auth.py:192-242`). Deployments that already contain orphaned identities stay broken even once the delete side is corrected, so this path has to cope on its own.

### Related issues

The delete side of this bug has surfaced twice before under different symptoms, both closed as not planned:

- #7202 — deleting an account does not invalidate active sessions/tokens (closed 2025-09-11, not planned).
- #8955 — deleting the admin account breaks worker communication (closed 2026-05-20, not planned); also asks for a delete guard on the built-in service accounts and for self-deletion prevention.

The same underlying fact drives all three: deleting a `CoreAccount` cleans up none of its Internal-namespace children, because the delete-dependency query cannot see them.

Related but a distinct defect in the same component — a race rather than a visibility gap, so not a duplicate:

- #6849 — cascade delete can miss related nodes added while the delete logic is running.

gmazoyer:
# Why

Infrahub declares a `refresh_tokens` relationship on `CoreGenericAccount`, so that
the delete of an account also removes its refresh tokens. This file is generated
from that schema, so it has to carry the same relationship. Without it,
`invoke backend.generate` in the Infrahub repository leaves this file dirty and the
`validate-generated` job stays red.

Needed by opsmill/infrahub#10435, which fixes opsmill/infrahub#10304.

## What changed

- `refresh_tokens` is added to `CoreGenericAccount` and `CoreGenericAccountSync`.

Nothing else. The file is rendered by `invoke backend.generate` in the Infrahub
repository, not written by hand.

## How to review

One commit on top of `stable`, two lines. The Infrahub submodule already pins the
`stable` head, so the pointer bump on that side moves by this commit alone.

### DIFF
--- a/infrahub_sdk/protocols.py
+++ b/infrahub_sdk/protocols.py
@@ -121,6 +121,7 @@ class CoreGenericAccount(CoreNode):
     account_type: Enum
     status: Dropdown
     tokens: RelationshipManager
+    refresh_tokens: RelationshipManager
     external_identities: RelationshipManager
 
 
@@ -726,6 +727,7 @@ class CoreGenericAccountSync(CoreNodeSync):
     account_type: Enum
     status: Dropdown
     tokens: RelationshipManagerSync
+    refresh_tokens: RelationshipManagerSync
     external_identities: RelationshipManagerSync
 
 


polmichel (approved):

ogenstad (changes requested):
Sanity check for this one. These are in the `Internal` namespace. I don't think they are exposed through the GraphQL API? 

Are we sure this is needed?

gmazoyer:
> Sanity check for this one. These are in the `Internal` namespace. I don't think they are exposed through the GraphQL API?
> 
> Are we sure this is needed?

I guess we have a code generator bug for Python SDK protocols then, because we have internal stuff in there already (even though we do not expose them over the graphql API).

polmichel:
> Sanity check for this one. These are in the `Internal` namespace. I don't think they are exposed through the GraphQL API?
> 
> Are we sure this is needed?

External identities are also exposed in the SDK and they also belong to the Internal namespace `backend/infrahub/core/schema/definitions/core/account.py:183`. This is why I thought it was expected.

ogenstad (approved):
Approving instead since this is just a temporary fix to resolve a CI validation issue and no consumers will attempt to use these. The target will then be to avoid these to be generated within the 1.12.0 cycle for Infrahub.

PR #1319, opened 2026-09-03
ci: test the declared dependency lower bounds and the install surface

ogenstad:
# Why

The `unit-tests` matrix runs five Python versions but every job installs exactly what `uv.lock` pins, so the version ranges declared in `pyproject.toml` have never been exercised by anything. This adds the jobs that hold a pull request to them.

**Goal:** the declared lower bounds are correct, and the published package installs on every supported interpreter.

**Non-goals:** upper bounds. Any upstream release could turn such a job red without a line of this code changing, and blocking an unrelated pull request on that is worse than not knowing. Tracked separately.

Ref IHS-224.

## What changed

Two jobs:

- **`dependency-lower-bounds`** re-resolves at the declared floors and runs the unit tests. Gates pull requests, because a floor only moves when `pyproject.toml` does.
- **`install-matrix`** builds the wheel and confirms it installs, imports every shipped module and runs the CLI on all five supported interpreters, resolving fresh rather than from the lock. This is the half the existing matrix cannot see: it proves the pinned set works, not what a user actually gets from a plain install.

Writing the first job immediately found two wrong floors:

- **`typer>=0.15.0` was impossible.** Combined with the `click>=8.3` this package already required, typer 0.15 crashes on any `infrahubctl --help` with `TypeError: Parameter.make_metavar() missing 1 required positional argument: 'ctx'`. click 8.3 made that argument mandatory and typer only followed in 0.16. Verified by holding typer at 0.15.0 and moving click:

  | typer | click | `--help` |
  | --- | --- | --- |
  | 0.15.0 | 8.3.0 | crashes |
  | 0.15.0 | 8.1.8 | exit 0 |
  | 0.16.0 | 8.3.0 | exit 0 |

  So `>=0.16.0` is not a narrowing, it states a constraint that already existed implicitly. Note that widening typer support means lowering the `click` floor, not this one; typer 0.15.4 additionally caps `click<8.2`, so the compatible set is not a simple range.

- **`Jinja2>=3` was wrong.** Before 3.1.5, template error reporting attributes a nested undefined variable to the wrong template and drops the source path from a missing-import error. Bisected: 3.1.4 fails, 3.1.5 passes.

Five development requirements had no lower bound at all, which made the floor resolution fail outright rather than merely resolve oddly: `yamllint` reached back to 0.1.0, whose sdist does not build. Bounds on development dependencies carry no promise to users; they only need to keep the resolution sane, so they are set near what the lock already holds.

## How to review

Start with the two jobs in `.github/workflows/ci.yml`, then the floor changes in `pyproject.toml`. The reasoning that most deserves scrutiny is in the comments on `dependency-lower-bounds`, covering why it is a single job on 3.10 rather than a matrix, and what it cannot reach.

Two limits stated there and worth repeating:

- Where a development dependency constrains a package more tightly than we do, it wins. pytest, httpx, graphql-core, pydantic, anyio and typing-extensions all resolve **above** their declared floors in this job. Reaching those needs an environment with no development dependencies in it, which this does not attempt.
- 9 floors are genuinely exercised: dulwich 0.24.7, jinja2 3.1.5, netutils 1.0.0, pyarrow 14.0.0, pyyaml 6.0, rich 12.0.0, typer 0.16.0, ujson 5.0.0, whenever 0.9.3.

## How to test

```bash
uv lock --resolution lowest-direct && uv sync --all-groups --all-extras && uv run pytest tests/unit/
git checkout uv.lock && uv lock   # restore afterwards
```

Verified locally before and after the floor corrections:

| Environment | Result |
| --- | --- |
| floors, before the corrections | 18 failed, 1843 passed |
| lock versions on 3.10 (control) | 8 failed, 1853 passed |
| floors, after the corrections | 7 failed, 1854 passed |

The failures common to the floor and control runs are artifacts of a scratch virtualenv, not of the floors. After the corrections, the floor run's failure set is a strict subset of the control's, so **no floor-specific failure remains**. The 11 that were floor-specific split cleanly into the two clusters above, 7 typer and 4 Jinja2.

Also verified: a fresh default-resolution install of `[all]` succeeds and imports every shipped module on 3.10, 3.12 and 3.14; ruff, ty and mypy clean; `yamllint -s .` clean; unit suite 1859 passed at lock versions with the 2 known macOS temp-path failures; lock consistent.

## Impact & rollout

- **Backward compatibility:** raising the two floors is resolver-visible. Nothing that worked stops working, since typer 0.15 could not produce a working CLI against the required click, and the Jinja2 versions dropped misreport template errors.
- **Performance:** adds one job plus a five-way matrix of install-only jobs. The matrix does no test running, so it is cheap.
- **Deployment notes:** `dependency-lower-bounds` is intended as a required check. It is deterministic, so it will not flap.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added

---
## Summary by cubic
Adds three CI jobs that exercise the declared dependency ranges in `pyproject.toml`, which the unit-test matrix never covered because it installs exactly what `uv.lock` pins. The lower-bound jobs exposed two wrong floors, now corrected: `typer>=0.16.0` (0.15 crashes on any `infrahubctl --help` with the `click>=8.3` the SDK already required) and `Jinja2>=3.1.5` (earlier versions misattribute template errors). Addresses IHS-224.

**CI jobs**
- `dependency-lower-bounds` re-resolves at the declared floors and runs the unit suite on Python 3.10, the only supported interpreter with floor-era wheels for pyarrow, pyyaml and ujson.
- `dependency-lower-bounds-no-harness` reaches the floors the first job cannot, since the test harness pulls pytest, pydantic, anyio and typing-extensions above their declared floors. It imports every shipped module, runs the CLI, and collects a test through the bundled pytest plugin.
- `install-matrix` builds the wheel and installs it with a fresh resolution on all five supported Python versions, importing every shipped module and running the CLI.
- All wheel checks run from outside the checkout so they test the installed wheel, not the source tree.
- Upper bounds are deliberately not tested; any upstream release could turn such a job red with no code change, so that work is tracked separately.

**Dependency metadata**
- Five development dependencies with no lower bound now have one near what the lock holds; without them the floor resolution fails outright.
- Pinning either corrected package below its new floor now fails at resolution time instead of breaking at runtime.

<sup>Written for commit 7b1a6179ac8988e1128e723f1858b44c6e5701ee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1319?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


.github/workflows/ci.yml
```diff
@@ -310,6 +310,127 @@ jobs:
...
+        uses: "actions/checkout@v7"
+      # Pinned to the oldest interpreter we support, because that is where floor-era wheels exist.
+      # The resolver selects the same floors on every supported version, but installing them does
+      # not work everywhere: pyarrow 14, pyyaml 6.0 and ujson 5.0 ship no wheels past cp312, so on
+      # 3.13 or 3.14 this job would compile them from source and fail for reasons unrelated to our
+      # code. A matrix would therefore add nothing. A floor states the oldest version we have
+      # validated, not one installable on every interpreter; a newer interpreter simply resolves
+      # higher, which `install-matrix` below confirms still works.
+      # Bump this when the oldest supported version changes. If it goes stale the job fails loudly,
+      # because `requires-python` will reject the interpreter.
```
gmazoyer: bot's probably overdone that comment

gmazoyer (approved):

PR #1320, opened 2026-09-03
Send the SDK release dispatch to every repository listening for it

BeArchiTek:
## Summary

When the SDK publishes a release, this workflow tells downstream repositories so they can open their own version-bump pull request. The matrix names four recipients. **Eleven** repositories carry a listener for `trigger-infrahub-sdk-python-update`, and only two of them are wired — `infrahub-demo-dc` and `infrahub-sync`. The other nine have bump workflows that have never once been triggered.

This adds the nine missing recipients, and flags one dispatch that goes nowhere.

## Key Changes

Added to the matrix:

- `opsmill/infrahub-ansible`
- `opsmill/infrahub-arista-avd`
- `opsmill/infrahub-demo-otn`
- `opsmill/infrahub-demo-service-catalog`
- `opsmill/infrahub-demo-sp`
- `opsmill/infrahub-mcp`
- `opsmill/infrahub-solution-ai-dc`
- `opsmill/infrahub-sync-lab`
- `opsmill/nornir-infrahub`

## One decision left for an owner: `opsmill/emma`

`emma` is dispatched to on every SDK release and has **no listener for this event at all**, so the dispatch is accepted and consumed by nothing.

## Related

Companion PR for the Infrahub-side matrix, which had the same defect: opsmill/infrahub#10510.

`opsmill/infrahub-demo-otn`'s listener for this event is in flight and lands before this is useful to it. A dispatch to a repository with no matching listener is a no-op rather than an error, so the ordering is safe either way.

## Base branch

Targeted at `stable` rather than `develop` on purpose: `release.yml` fires on `release: published`, so the file that executes is the one at the release tag, and the bot already syncs `stable` into `develop`. Please correct me if that is wrong for this repo — recent PRs here are split across `develop`, `infrahub-develop` and `stable`, so I reasoned it from the release trigger rather than from convention.

## Test Plan

Nothing to run: the change is a matrix list. Verification is on the next SDK release, where each newly added repository should show a run of its own bump workflow shortly after the release publishes.

---
## Summary by cubic
Wires the SDK release dispatch to every repository listening for it. Only two of eleven listener repositories were wired before, so nine version-bump workflows never ran.

- Adds the nine missing repositories to the dispatch matrix.
- Leaves `opsmill/emma` in the matrix; its dispatch is a no-op, and whether it gains a listener or loses the dispatch is its owner's call.
- Records the search query that lists listeners in the matrix comment, so the next reconciliation is a diff instead of an audit.

<sup>Written for commit b341679a0e2f63012566ff7c01351e30dfb06a5c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1320?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/repository-dispatch.yml
+++ b/.github/workflows/repository-dispatch.yml
@@ -36,8 +36,17 @@ jobs:
         # Either a literal path, or the name of a secret...
         repo:
           - "opsmill/emma"
+          - "opsmill/infrahub-ansible"
+          - "opsmill/infrahub-arista-avd"
           - "opsmill/infrahub-demo-dc"
+          - "opsmill/infrahub-demo-otn"
+          - "opsmill/infrahub-demo-service-catalog"
+          - "opsmill/infrahub-demo-sp"
+          - "opsmill/infrahub-mcp"
+          - "opsmill/infrahub-solution-ai-dc"
           - "opsmill/infrahub-sync"
+          - "opsmill/infrahub-sync-lab"
+          - "opsmill/nornir-infrahub"
           - "INFRAHUB_CUSTOMER1_REPOSITORY"
 
     steps:


BeArchiTek:
Closing this unmerged, for the same reason as its companion opsmill/infrahub#10510: the dispatch system is being replaced with bot-triggered fan-out, which engineering is piloting, so hand-reconciling the matrix would be thrown-away work.

Recording the survey so the replacement does not have to redo it. As of 2026-09-03, these repositories carry an `on: repository_dispatch: types: [trigger-infrahub-sdk-python-update]` listener:

| Repository | In this matrix today |
|---|---|
| `infrahub-demo-dc` | yes |
| `infrahub-sync` | yes |
| `infrahub-ansible` | **no** |
| `infrahub-arista-avd` | **no** |
| `infrahub-demo-otn` | **no** |
| `infrahub-demo-service-catalog` | **no** |
| `infrahub-demo-sp` | **no** |
| `infrahub-mcp` | **no** |
| `infrahub-solution-ai-dc` | **no** |
| `infrahub-sync-lab` | **no** |
| `nornir-infrahub` | **no** |

Two of eleven wired. `infrahub-demo-dc-fabric` also listens but is archived, and `INFRAHUB_CUSTOMER1_REPOSITORY` resolves to a secret I cannot read.

**The `opsmill/emma` question outlives this PR and is worth carrying forward.** `emma` is in this matrix and has no listener for this event in any workflow, so every SDK release dispatches into a repository that consumes nothing. This is already known — `infrahub-github-bot/config.yaml` says so in a comment, "`emma` is deliberately absent: it is dispatched to today but has no listener" — but this matrix was never corrected to match. Whoever builds the replacement should decide whether `emma` gains a listener or loses the dispatch, rather than porting the dead target across.

That case is also the argument for not rebuilding this as two hand-maintained lists: the drift runs in both directions and is silent either way. A listener with no sender never fires; a sender with no listener dispatches into the void. Deriving one side from the other would remove the whole class.

BeArchiTek:
**Correction: disregard my previous comment and the close.** I misread the instruction. The matrix additions are wanted; what was asked for was dropping the drift-detection tooling, not the reconciliation itself. Reopened, and the branch now carries only the added repositories.

Removed since the earlier review:

- The `gh api search/code` reconciliation query and the surrounding explanation. The original `# Either a literal path, or the name of a secret...` comment is restored unchanged, so the diff is now purely the new matrix entries.
- This also resolves the cubic-dev-ai finding, which was correct: that query matched this repository too, because the sender's own `event-type:` line contains the same string, so a naive diff would have reported the sender as a missing listener. The query is gone, so the problem is gone with it.

The branch was force-pushed to amend the single commit rather than stack a revert on top, because the original commit message described the query that no longer exists.

Longer term this is superseded by bot-triggered fan-out, which engineering is piloting. This PR is the interim step that gets the six unwired repositories receiving release notifications in the meantime.

fatih-acar (approved):

PR #1323, opened 2026-09-04
feat(client): retry transient failures, optionally indefinitely

fatih-acar:
## Summary

Generators and other long-running SDK jobs abort as soon as a single GraphQL call hits a transient infrastructure error, even though the task worker already runs the client with `retry_on_failure=True`. That switch only covered connection errors inside `execute_graphql`, gave up after 5 minutes, spun without any delay on HTTP 5xx, and never looked at REST calls or at GraphQL error envelopes. This PR turns `retry_on_failure` into a real transient-error policy and lets operators opt into retrying indefinitely, so a multi-hour generator survives a database failover or a network outage instead of failing.

## Key changes

- With `retry_on_failure` enabled, connection errors, timeouts, HTTP 500/502/503/504 responses and GraphQL errors the server flags with one of those statuses are retried. Everything else still fails fast, so a bad query or a schema error surfaces immediately even in unlimited mode.
- A connection dropped before any response arrives is retried too. httpx reports that as `RemoteProtocolError` rather than a network error, so it used to escape the retry handler entirely; it is now mapped to `ServerNotReachableError` like any other lost connection, on all six transport paths of both clients. See the section below for how this was found.
- Every httpx timeout is retried, not only `ReadTimeout`. `ConnectTimeout` is a handshake left hanging by a load balancer dropping packets mid-failover and maps to `ServerNotReachableError`; `WriteTimeout` (an upload the server stopped draining) and `PoolTimeout` map to `ServerNotResponsiveError` like a read timeout. Previously all three escaped the retry handler as raw httpx exceptions.
- Retries now cover every request path: GraphQL queries and mutations, the generator's `query_gql_query` data collection, schema loading and the other REST endpoints. The GraphQL-envelope retry shares one time budget and attempt counter with the transport layer.
- Exponential backoff with jitter replaces the fixed delay. `retry_delay` is the base and the new `retry_max_delay` caps it. This also fixes the 5xx busy loop.
- `max_retry_duration=0` now means retry indefinitely. On the task worker, `INFRAHUB_MAX_RETRY_DURATION=0` is enough to opt in because the worker already sets `retry_on_failure=True`.
- The new `retry_status_codes` setting controls which statuses count as transient. 500 is included because Infrahub reports some transient database errors (Neo4j transient or session errors exhausting the server-side retries) without further classification.
- Every retry is logged with the attempt number and elapsed time, escalating from WARNING to ERROR after five minutes so an indefinitely retrying job stays visible.
- Once the budget is exhausted the original error is raised instead of a generic "resp hasn't been initialized" error.
- `client.retry_on_failure` and `client.retry_delay` become properties so they can still be toggled at runtime, for example by a generator.

## Verified against a real failover

The policy was exercised against a real Infrahub deployment from `infrahub-testcontainers` (two API replicas behind HAProxy, plus Neo4j, RabbitMQ, Redis and Prefect) rather than against mocks alone. The server's `/api/response-delay` endpoint makes every GraphQL request sit for 10 seconds, which is a wide enough window to kill a container while a mutation is genuinely in flight.

That found the feature aborting on the exact scenario it exists for. With `retry_on_failure=True` and `max_retry_duration=0`, restarting HAProxy during a mutation killed the operation after a single attempt:

```text
MUTATION FAILED after 10.2s: httpx.RemoteProtocolError: Server disconnected without sending a response.
```

A load balancer that goes away mid-request closes the socket before sending a single byte of the response, and httpx reports that as `RemoteProtocolError`, a `ProtocolError` rather than a `NetworkError`. Only `NetworkError` and `ReadTimeout` were mapped to the SDK exceptions the retry handler treats as transient, so a failover bypassed the handler entirely: zero retries, unlimited budget notwithstanding. With the mapping added here, the same failover only delays the mutation:

```text
WARNING infrahub_sdk: Transient failure on .../graphql/main: Unable to connect to 'http://localhost:18000'.
                     Retry 1 in 0.9s (10s elapsed, no time limit)
INFO  httpx: HTTP Request: POST .../graphql/main "HTTP/1.1 200 OK"
```

A failover reaches the client in one of two shapes, depending on which side of the load balancer goes away. Both are now covered:

| Restarted | What the client sees | Retried by |
| --- | --- | --- |
| HAProxy | The connection dropped before any response (`RemoteProtocolError`) | The mapping added in this PR |
| The API servers behind it | HTTP 502 for the request in flight, 503 while they boot | `retry_status_codes`, which already worked |

The at-least-once caveat below turned out not to be theoretical. The first run used a plain create and the retry came back with `Violates uniqueness constraint 'name'`: the server keeps processing a request whose client has gone away, so attempt one had already written the node before the socket closed. The integration tests use `save(allow_upsert=True)` for that reason.

## Caveats

- A mutation whose connection was lost may have been applied by the server before the retry, as observed above. Upserts are idempotent. A bare create that already succeeded fails on retry with a non-transient error, which is raised.
- Because 500 is in the default set, a genuine bug that returns 500 is retried until the budget expires (5 minutes by default). Remove 500 from `retry_status_codes` to fail fast on it.
- A per-generator opt-in in `.infrahub.yml` needs an Infrahub-side change to forward the flag to the client and is left as a follow-up.

## Documentation

- `docs/docs/python-sdk/guides/client.mdx`: new "Retry transient failures" section under advanced use cases, listing a dropped connection among the transient failures.
- Regenerated `docs/docs/python-sdk/reference/config.mdx` and `docs/docs/python-sdk/sdk_ref/infrahub_sdk/client.mdx`.
- Changelog fragments `changelog/+transient-retry.added.md`, `changelog/+retry-5xx-busy-loop.fixed.md` and `changelog/+retry-connection-dropped-mid-request.fixed.md`.

## Test plan

- `uv run pytest tests/unit/sdk/test_retry.py`: 78 tests covering classification, backoff, budget and unlimited mode, the shared budget across layers, the opt-in default, runtime toggling, environment plumbing and async/sync parity. 12 of them are new and drive a lost connection through httpx itself, so the `RemoteProtocolError` mapping is covered without needing Docker.
- `uv run pytest tests/integration/test_retry_on_failover.py`: 4 tests against the deployment described above, 3m30s. Three restart HAProxy mid-mutation (retries disabled, async retry-forever, sync retry-forever) and all three fail without the mapping, so they are a real regression test rather than a demonstration. The fourth restarts the API servers instead and asserts the retry happened on a 502, not merely that the mutation eventually succeeded.
- Two details that reviewers may wonder about are documented in the test file: the load balancer is published on a pinned host port, because Docker assigns a new one on every container start and the client would otherwise retry against a dead address; and the response delay lives in the memory of each API worker process, so the test that restarts them re-applies it in teardown.
- `uv run invoke lint-code` passes. `uv run invoke lint-docs` passes rumdl; vale was not available locally.
- Full unit suite passes apart from the known pre-existing `tests/unit/ctl` failures.

---
## Summary by cubic
Turns `retry_on_failure` in the `infrahub_sdk` client into a real transient-error policy. Previously it retried only connection errors inside `execute_graphql`, capped at 5 minutes, and skipped REST calls, GraphQL error envelopes, the no-delay 5xx busy loop, and connections dropped before a response arrived. Now connection errors, dropped connections, every httpx timeout, HTTP 500/502/503/504 responses, and GraphQL envelopes flagged with those statuses are retried on every request path—multipart uploads and streamed downloads included—with exponential backoff and jitter, optionally forever.

**Key changes**
- New `retry_max_delay` caps the backoff; `retry_status_codes` tunes which statuses count as transient (500 is in the default set because Infrahub reports some transient DB errors without classification). A zero `retry_delay` or `retry_max_delay` is rejected, since either made every retry a tight loop.
- GraphQL-envelope retries share one time budget and attempt counter with transport-level retries, and each retry is logged, escalating from WARNING to ERROR after five minutes.
- Budget exhaustion raises the original error instead of a generic "resp hasn't been initialized" error.
- `retry_on_failure` and `retry_delay` can now be toggled at runtime.
- A connection dropped before any response arrives (the shape a failover takes) is now mapped to `ServerNotReachableError` and retried; `ConnectTimeout` maps there too and the remaining timeouts (write, pool) to `ServerNotResponsiveError`. Opt-in integration tests (`INFRAHUB_TESTING_FAILOVER=1`) restart HAProxy mid-mutation to cover it, with the sync case time-bounded.
- A multipart mutation answered with a transient GraphQL error envelope retries with the file rewound before each attempt; non-seekable streams (pipes, sockets) are copied to a temp file first (off the event loop in the async client) so a retried send carries the full body. Cancelling an async upload now returns at once, leaving the worker thread to close the temp file.
- A streamed download interrupted mid-body is reported as `ServerNotResponsiveError`, restarted on the shared retry budget, and the partial file is removed after a failed attempt. Streamed transient responses are only pre-read when a retry will actually happen, so a spent budget returns the response open instead of fetching its body. A destination that cannot be opened is left untouched, and a failed cleanup never replaces the transfer error.

**Caveats**
- A timed-out mutation may have been applied before the retry; upserts are idempotent, but a bare create that succeeded fails on retry with a non-transient error.
- Remove 500 from `retry_status_codes` if you want genuine bugs to fail fast instead of retrying until the budget expires.
- Per-generator opt-in via `.infrahub.yml` is left as a follow-up.

<sup>Written for commit a0a185d79440b7ce88a7c232115ad6cfb8982171. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1323?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


infrahub_sdk/client.py
```diff
@@ -1611,11 +1633,15 @@ async def _request(
...
+        retry_state: RetryState | None = None,
     ) -> httpx.Response:
         async def send() -> httpx.Response:
             return await self._request_method(url=url, method=method, headers=headers, timeout=timeout, payload=payload)
 
-        response = await self._rate_limit_handler.asend(send=send, url=url)
+        async def send_with_rate_limit() -> httpx.Response:
+            return await self._rate_limit_handler.asend(send=send, url=url)
+
+        response = await self._retry_handler.asend(send=send_with_rate_limit, url=url, state=retry_state)
```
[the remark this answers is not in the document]
fatih-acar: Valid point — the transient handler only wrapped `_request`, so multipart uploads and streamed downloads bypassed it. Now `_request_multipart` and stream initiation in `_get_streaming` go through the same handler on both clients, in d51b3bb: lost connections and timeouts are mapped inside the stream initiation so they can be retried, and a transient status closes the failed stream before the retry, the way the 429 handling already does. Covered by the `test_all_request_paths_retry_*` tests for the regular, multipart and streaming paths on both clients.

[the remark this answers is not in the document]
fatih-acar: Partly valid. The multipart mutation path (`_execute_graphql_with_file`) now applies the same transient-envelope classification and shared backoff as `execute_graphql`, rewinding the file before every attempt, in d51b3bb (`test_multipart_mutation_retries_transient_graphql_errors_on_the_shared_budget`).

Not changing `query_gql_query`: Infrahub's `/api/query/<name>` endpoint never returns a 200 envelope with `errors`. It raises `GraphQLQueryError` and answers HTTP 502 with a plain error list, so those failures are already handled by the transport-level status classification, and envelope handling there would be dead code against the server. The flip side is that the endpoint reports every GraphQL error as 502, transient or not, so a broken stored query is retried until the budget is exhausted; keeping the error classification on that endpoint is a server-side change outside this PR.

tests/integration/test_retry_on_failover.py
```diff
@@ -0,0 +1,215 @@
...
+        caplog.set_level("WARNING", logger="infrahub_sdk")
+        client = InfrahubClientSync(config=self.build_config(address, retry_on_failure=True))
+        client.schema.all()
+
+        node = client.create(kind="BuiltinTag", name="failover-sync")
+
+        restart = threading.Timer(RESTART_AFTER, restart_load_balancer, args=(load_balancer_container,))
+        restart.start()
+        try:
+            node.save(allow_upsert=True)
```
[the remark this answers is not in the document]
fatih-acar: Valid point — the sync test had no counterpart to the `asyncio.wait_for` bound of the async tests. The save now runs through `run_with_timeout`, a daemon-thread helper that fails the test after `RETRY_TIMEOUT` seconds and cannot block interpreter exit if the load balancer never comes back, in 7284ba7. The same commit also makes the whole module opt-in (`INFRAHUB_TESTING_FAILOVER=1`) so CI never runs these container-restarting tests.

tests/integration/test_retry_on_failover.py
```diff
@@ -0,0 +1,272 @@
...
+        The delay lives in the memory of each API worker process, set by a broadcast on the
+        message bus, so restarting those processes drops it. Restoring it here keeps the class
+        fixture's promise and leaves this test independent of the order tests run in.
+
+        Yields:
+            None: with the delay active, as ``slow_api`` leaves it.
+
+        """
+        yield
+        infrahub_compose.set_server_response_delay(RESPONSE_DELAY)
```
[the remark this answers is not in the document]
fatih-acar: Done in 7284ba7: the teardown now calls `wait_until_reachable(address)` before re-applying the delay, since the broadcast only reaches workers that are up. The fixture is kept even though the test currently runs last, so the class fixture's promise holds whatever order the tests run in.

infrahub_sdk/client.py
```diff
@@ -1411,24 +1436,35 @@ async def _execute_graphql_with_file(
...
-            query=query,
-            variables=variables,
-            file_content=file_content,
-            file_name=file_name or "upload",
-            headers=headers,
-            timeout=timeout,
-            operation_name=operation_name,
-        )
+        retry_state = self._retry_handler.new_state()
+        while True:
```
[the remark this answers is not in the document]
fatih-acar: Valid point — a stream that cannot seek was left consumed by the rewind before each attempt, so any retry (the pre-existing 429 one included) re-sent an empty file. `_execute_graphql_with_file` now copies a non-seekable `file_content` once to a temporary file before the first attempt and uploads from that copy, so every transport and envelope retry carries the full body, in c7ca018. Covered by `test_multipart_retry_re_sends_a_non_seekable_upload_in_full` on both clients.

infrahub_sdk/client.py
```diff
@@ -1575,14 +1625,19 @@ async def _get_streaming(
...
-                if response.status_code == 429:
+                try:
+                    response = await stack.enter_async_context(
+                        client.stream(method="GET", url=url, headers=headers, timeout=request_timeout)
+                    )
+                except CONNECTION_LOST_EXCEPTIONS as exc:
+                    raise ServerNotReachableError(address=self.address) from exc
+                except httpx.ReadTimeout as exc:
+                    raise ServerNotResponsiveError(url=url, timeout=request_timeout) from exc
+                if self._is_retried_stream_status(response.status_code):
```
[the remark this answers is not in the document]
fatih-acar: Valid point. `_get_streaming` now creates the retry state itself and `_is_retried_stream_status` asks `should_retry(state)` rather than `enabled`, so a transient status the handler will not retry (retries disabled or budget spent) reaches the caller open and unread, in fa0e4b7. `test_streaming_hands_over_an_open_transient_response_once_the_budget_is_spent` checks the stream is not consumed in that case.

tests/integration/test_retry_on_failover.py
```diff
@@ -0,0 +1,318 @@
...
+    def target() -> None:
+        try:
+            outcome.append(func())
+        except BaseException as exc:  # re-raised in the calling thread below
+            failure.append(exc)
+
+    worker = threading.Thread(target=target, daemon=True)
+    worker.start()
+    worker.join(timeout)
+    if worker.is_alive():
```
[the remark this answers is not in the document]
fatih-acar: Valid point. On timeout the guard now calls an `on_timeout` hook that turns the client's retries off (`client.retry_on_failure = False`), so the next failed attempt raises instead of sleeping again and the worker thread ends; the guard waits a grace period for that before failing the test, in 46b1ca4. If the load balancer happens to recover inside that window the save may still complete, which is harmless since the mutation is an idempotent upsert, and the thread stays a daemon as a last resort.

infrahub_sdk/client.py
```diff
@@ -80,6 +80,26 @@ class ProxyConfig(TypedDict):
...
+    stream cannot do: ``_rewind_multipart_files`` would leave it consumed and the retry would carry an
+    empty body. Such a stream is copied once, to a temporary file so the size does not matter, before
+    the first attempt.
+    """
+    seekable = getattr(file_content, "seekable", None)
+    if file_content is None or (callable(seekable) and seekable()):
+        yield file_content
+        return
+    with tempfile.TemporaryFile() as buffer:
+        shutil.copyfileobj(file_content, buffer)
```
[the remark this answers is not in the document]
fatih-acar: Valid point. The async upload path now runs the copy of a non-seekable stream in a worker thread via `asyncio.to_thread`, through an async counterpart of the helper, so draining a slow or large source no longer blocks the other tasks on the loop; the sync client keeps the plain copy, in 3a36222 (docstring and typing follow-ups in 3b942c4 and the commit after it). `test_async_upload_copies_a_non_seekable_stream_off_the_event_loop` checks that the copy is handed to the worker thread.

infrahub_sdk/client.py
```diff
@@ -80,6 +81,56 @@ class ProxyConfig(TypedDict):
...
+    if file_content is None or _is_rewindable(file_content):
+        yield file_content
+        return
+    source: BinaryIO = file_content
+    with tempfile.TemporaryFile() as buffer:
+
+        def copy() -> None:
+            shutil.copyfileobj(source, buffer)
+
+        await asyncio.to_thread(copy)
```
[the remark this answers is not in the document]
fatih-acar: Valid point. The copy now runs as a shielded task; when the awaiting upload is cancelled, `_aseekable_upload` waits for the worker thread to end before the temporary file is closed and the cancellation is re-raised, in 661133c. `test_cancelled_async_upload_waits_for_the_stream_copy_before_closing_the_buffer` blocks the copy mid-stream, cancels, checks the cancellation does not surface while the thread is still copying, then releases it and checks the stream was drained to EOF before the `CancelledError` arrives.

infrahub_sdk/client.py
```diff
@@ -80,6 +81,68 @@ class ProxyConfig(TypedDict):
...
+        def copy() -> None:
+            shutil.copyfileobj(source, buffer)
+
+        copy_task = asyncio.ensure_future(asyncio.to_thread(copy))
+        try:
+            await asyncio.shield(copy_task)
+        except asyncio.CancelledError:
+            await asyncio.wait({copy_task})
+            if not copy_task.cancelled():
+                copy_task.exception()  # the cancellation is what the caller sees; mark any copy error retrieved
```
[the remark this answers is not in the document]
fatih-acar: Valid point. The drain now runs in `_wait_despite_cancellation`, a loop that absorbs every cancellation aimed at the caller until the copy task is done, and only then is the buffer closed and the original `CancelledError` re-raised, in 34f6717. The cancellation test now cancels a second time while the drain is pending and checks it still does not surface before the stream has been drained to EOF.

infrahub_sdk/config.py
```diff
@@ -63,8 +63,35 @@ class ConfigBase(BaseSettings):
...
+        description=(
+            "Base delay in seconds before retrying a request that failed with a transient error. "
+            "The delay doubles after every attempt, with jitter, up to the maximum retry delay."
+        ),
+    )
+    retry_max_delay: int = Field(
+        default=60,
+        ge=0,
+        description="Maximum delay in seconds between two retries of a request that failed with a transient error.",
+    )
```
gmazoyer, line 78: both of these settings use `ge=0`, and compute_backoff returns `min(max_delay, ...)`, so either zero makes every retry sleep 0 seconds. Managed to make 20000+ attempts in one second against a persistent 502.

infrahub_sdk/client.py
```diff
@@ -80,6 +81,81 @@ class ProxyConfig(TypedDict):
...
+    """Wait for ``task`` to finish, absorbing every cancellation aimed at the caller meanwhile.
+
+    Used to drain a worker thread that cannot be interrupted: the resources it uses must outlive it,
+    and the caller re-raises the cancellation it already holds once this returns.
+    """
+    while not task.done():
+        try:
+            await asyncio.wait({task})
+        except asyncio.CancelledError:
+            continue
```
gmazoyer: What happens if the source stream never returns from `read()`? It feels like things could get stuck here even at default config, with retries off.

infrahub_sdk/retry.py
```diff
@@ -0,0 +1,274 @@
...
+DEFAULT_RETRY_STATUS_CODES: frozenset[int] = frozenset({500, 502, 503, 504})
+"""HTTP status codes treated as transient unless ``retry_status_codes`` says otherwise."""
+
+ESCALATE_AFTER_SECONDS = 300.0
+"""Once an operation has been retrying for this long, retry log lines switch from WARNING to ERROR."""
+
+TRANSIENT_EXCEPTIONS = (ServerNotReachableError, ServerNotResponsiveError)
+"""Client-side failures (connection error, read timeout) that are always considered transient."""
+
+CONNECTION_LOST_EXCEPTIONS = (httpx.NetworkError, httpx.RemoteProtocolError)
```
gmazoyer: Should `ConnectTimeout`, `WriteTimeout` and `PoolTimeout` be in that sequence to be retried too? `WriteTimeout` is used by the upload path this PR tries to harden.

infrahub_sdk/client.py
```diff
@@ -1591,15 +1731,18 @@ async def send() -> httpx.Response:
...
+                response = await self._retry_handler.asend(send=send_with_rate_limit, url=url, state=retry_state)
                 try:
                     yield response
                 finally:
                     stack = open_stream.get("stack")
                     if stack is not None:
                         await stack.aclose()
-            except httpx.NetworkError as exc:
+            except CONNECTION_LOST_EXCEPTIONS as exc:
                 raise ServerNotReachableError(address=self.address) from exc
```
gmazoyer: Is this the right exception for a connection lost mid-download?

The catch around the `yield` now includes `RemoteProtocolError`, so a transfer truncated while the caller reads the body surfaces as "Unable to connect to ..." with no retry, and `file_handler` leaves a partial file with no hint, even though the server was reachable in the first place.

changelog/+transient-retry.added.md
```diff
@@ -0,0 +1 @@
+With `retry_on_failure` enabled, the client now retries every transient failure instead of only connection errors: a connection dropped before any response arrives (the shape a load balancer failover or a server restart takes on an in-flight request), read timeouts, HTTP `500`/`502`/`503`/`504` responses and GraphQL errors the server flags with one of those statuses (for example a database that became unavailable mid-run). Retries apply to every request path, including multipart uploads, streamed downloads and REST endpoints such as `query_gql_query` used by generators, and use exponential backoff with jitter (`retry_delay` as the base, capped by the new `retry_max_delay`) instead of retrying HTTP 5xx responses in a tight loop without any delay. The new `retry_status_codes` setting tunes which statuses count as transient, and `max_retry_duration=0` retries indefinitely so long-running generators can survive an outage rather than abort. Retries are logged with the attempt number and elapsed time, escalating from `WARNING` to `ERROR` after five minutes, and once the budget is exhausted the original error is raised instead of a generic one.
```
gmazoyer: This can probably be shorten without losing its meaning.

infrahub_sdk/config.py
fatih-acar: Valid point, thanks for measuring it. Both settings are now `gt=0` in 9e4f8ce; as integers that makes 1 second the floor. `max_retry_duration` keeps `ge=0` since 0 means unlimited there.

infrahub_sdk/client.py
fatih-acar: Same as before this PR: httpx reads the source synchronously while sending the body, so a `read()` that never returns hung the upload already, and in the async client the event loop with it. The copy moves that read to a worker thread. It runs regardless of the retry settings because the 429 backoff and the 401 re-login also re-send the body. What did get worse was cancellation, which waited for the worker: in 9e4f8ce a cancel surfaces at once and the worker closes the temporary file itself once its pending read returns.

infrahub_sdk/retry.py
fatih-acar: Yes to all three, in 9e4f8ce. `ConnectTimeout` joins the connection-lost set (a handshake left hanging by a load balancer dropping packets), and every catch site now takes `httpx.TimeoutException`, so write and pool timeouts map to `ServerNotResponsiveError` like read timeouts. Covered for both clients.

infrahub_sdk/client.py
fatih-acar: Agreed, it was not. In 9e4f8ce a body breaking off raises `ServerNotResponsiveError` saying the connection was lost while reading the response, the file handler restarts the download on the shared retry budget when retries are on, and the partial file is removed either way.

changelog/+transient-retry.added.md
fatih-acar: Shortened to two sentences in 9e4f8ce.

infrahub_sdk/retry.py
```diff
@@ -0,0 +1,277 @@
...
+"""Once an operation has been retrying for this long, retry log lines switch from WARNING to ERROR."""
+
+TRANSIENT_EXCEPTIONS = (ServerNotReachableError, ServerNotResponsiveError)
+"""Client-side failures (connection error, timeout) that are always considered transient."""
+
+CONNECTION_LOST_EXCEPTIONS = (httpx.NetworkError, httpx.ConnectTimeout, httpx.RemoteProtocolError)
+"""httpx failures meaning the connection was lost or never established, mapped to ``ServerNotReachableError``.
+
+``ConnectTimeout`` is included because a load balancer that drops packets during a failover leaves the TCP
+handshake hanging instead of refusing it, and httpx files that under timeouts rather than network errors.
```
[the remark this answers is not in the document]
fatih-acar: Done in 623a85f, wording taken from the suggestion.

infrahub_sdk/file_handler.py
```diff
@@ -275,28 +281,44 @@ async def _stream_to_file(self, url: str, dest: Path) -> int:
...
                         await f.write(chunk)
                         bytes_written += len(chunk)
-                return bytes_written
-        except ServerNotReachableError:
-            self._client.log.error(f"Unable to connect to {self._client.address}")
-            raise
+            except BaseException:
+                await anyio.Path(dest).unlink(missing_ok=True)
+                raise
+            return bytes_written
```
gmazoyer, line 324: A failed `open("wb")` now also triggers the `unlink`, so a pre-existing dest gets deleted by an attempt that never wrote to it (to reproduce: dest with mode 0444, open raises `PermissionError`, the file is gone, on stable it survived). The file could be opened before the try, and the unlink wrapped in `suppress(OSError)` so a failed cleanup does not replace the real error. Same in the sync version of this code.

infrahub_sdk/file_handler.py
```diff
@@ -275,28 +281,44 @@ async def _stream_to_file(self, url: str, dest: Path) -> int:
...
-                    await resp.aread()
-                    self.handle_error_response(exc=exc)
-
-                bytes_written = 0
+        retry_handler = self._client._retry_handler
+        retry_state = retry_handler.new_state()
+        while True:
+            try:
+                return await self._download_to_file(url=url, dest=dest, retry_state=retry_state)
+            except (ServerNotReachableError, ServerNotResponsiveError) as exc:
```
gmazoyer, line 295: This tuple of error classes could be replaced by `TRANSIENT_EXCEPTIONS`

infrahub_sdk/file_handler.py
fatih-acar: Good catch, the open should not have been inside the try. In 980932f the file is opened first so a dest that cannot be opened is left untouched, and the unlink runs after the file is closed, inside `suppress(OSError)`, so the transfer error stays the one raised. Same in the sync handler, with tests on both clients for a pre-existing dest that cannot be opened and for a cleanup that fails.

infrahub_sdk/file_handler.py
fatih-acar: It was `TRANSIENT_EXCEPTIONS` at first, but ruff's DOC501 cannot resolve an imported tuple against the documented Raises entries and flags the bare `raise`, whereas the same pattern passes in `retry.py` because the tuple is local there. Listing the two classes keeps the except clause aligned with the docstring without a `noqa`.

gmazoyer (approved):

PR #1324, opened 2026-09-04
feat(client): add connect_timeout to bound the connect phase

fatih-acar:
## Summary

An Infrahub endpoint that does not answer the TCP or TLS handshake used to hold every SDK request for the full request timeout (60 seconds by default) before failing. This PR bounds the connect phase separately, so a client behind a DNS round robin or a failing-over load balancer moves on to its next attempt after a few seconds instead of waiting out the whole request budget.

## Key Changes

- A new `connect_timeout` setting (default 10 seconds, `INFRAHUB_CONNECT_TIMEOUT`) bounds the TCP/TLS handshake on every request path: regular requests, multipart uploads and streaming downloads, on both the async and sync clients. It never exceeds the per-request `timeout`.
- The `timeout` setting's description now states what it covers (read, write and pool phases) instead of calling itself a "connection timeout". No behaviour change.

The transient retry handler on `stable` already treats a connect timeout as a lost connection and retries it when `retry_on_failure` is enabled, so this PR only shortens how long each failed attempt takes. An earlier revision of this PR also carried that retry mapping; it was dropped during the rebase onto `stable`.

## Documentation Updates

- `docs/docs/python-sdk/reference/config.mdx` regenerated with the new setting.
- One changelog fragment in `changelog/`.

## Test Plan

- `uv run pytest tests/unit/sdk/test_connect_timeout.py` covers the config default and validation, the exact timeout httpx receives on every path for both clients, the cap by the request timeout, and that a connect timeout still surfaces as `ServerNotReachableError` and is retried through `execute_graphql`.
- `uv run invoke format lint-code` and `uv run invoke docs-validate` pass locally.
- The full `tests/unit` suite passes except three pre-existing `infrahubctl` table-rendering failures that fail identically on `stable`.

### DIFF
--- a/infrahub_sdk/client.py
+++ b/infrahub_sdk/client.py
@@ -380,6 +380,16 @@ class BaseClient:
             merged.update(headers)
         return merged
 
+    def _build_timeout(self, timeout: int) -> httpx.Timeout:
+        """Build the httpx timeout for a single request.
+
+        ``timeout`` bounds the read, write and pool phases. The connect phase (TCP and TLS
+        handshake) is bounded by ``config.connect_timeout`` instead, so an unreachable address
+        fails fast rather than consuming the whole request budget, and the caller's ``timeout``
+        remains an upper bound for it.
+        """
+        return httpx.Timeout(timeout=timeout, connect=min(self.config.connect_timeout, timeout))
+
     @property
     def request_context(self) -> RequestContext | None:
         return self._request_context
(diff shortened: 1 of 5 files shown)


infrahub_sdk/client.py
```diff
@@ -380,6 +380,16 @@ def _merge_request_headers(self, headers: dict | None) -> dict:
...
 
+    def _build_timeout(self, timeout: int) -> httpx.Timeout:
+        """Build the httpx timeout for a single request.
+
+        ``timeout`` bounds the read, write and pool phases. The connect phase (TCP and TLS
+        handshake) is bounded by ``config.connect_timeout`` instead, so an unreachable address
+        fails fast rather than consuming the whole request budget, and the caller's ``timeout``
+        remains an upper bound for it.
+        """
+        return httpx.Timeout(timeout, connect=min(self.config.connect_timeout, timeout))
```
ogenstad: For extra clarity since we have the keyword argument for the second param I think we can change this:

```diff
- return httpx.Timeout(timeout, connect=min(self.config.connect_timeout, timeout))
+ return httpx.Timeout(timeout=timeout, connect=min(self.config.connect_timeout, timeout))
```

This is assuming that "timeout" is the correct parameter name, but I think it is.

ogenstad (approved):

infrahub_sdk/client.py
fatih-acar: Done in 2966337. `timeout` is indeed the parameter name, on the current httpx as well as the 0.20 floor we declare, so the keyword form is safe across the supported range.

PR #1332, opened 2026-09-08
test: pin Rich colour and width so CLI-output tests stop failing outside CI

saltas888:
# Why

Tests that assert on CLI text compare against output whose colour and width Rich decides from the ambient environment. On a developer machine that exports `FORCE_COLOR` (or has a narrow terminal), Rich emits ANSI escapes and truncates its tables, so a large block of tests fails locally while staying green in CI — which trains contributors to explain away red test output.

This was reported independently by two contributors in the same week, in their PR descriptions: [#1227](https://github.com/opsmill/infrahub-sdk-python/pull/1227) recorded 3 pre-existing failures in `tests/unit/sdk/test_schema.py` that "assert on plain strings while Rich emits ANSI colour codes in this environment", and [#1233](https://github.com/opsmill/infrahub-sdk-python/pull/1233) recorded further unrelated-to-that-PR failures on `stable`.

**Goal:** make the suite's rendering environment deterministic, so CLI-output assertions give the same answer everywhere.

**Non-goals:** no production code changes, and no loosening of any assertion — the exact-output tests keep their value.

Closes INBOX-166

## What changed

- **`tests/conftest.py`** — a `pytest_configure` hook pins the rendering environment for the whole suite: unset `FORCE_COLOR`, set `NO_COLOR=1` and `COLUMNS=200`.
- **`tests/unit/sdk/test_schema.py`** — the three consoles the tests build themselves are now explicit: `Console(file=StringIO(), width=1000, no_color=True, force_terminal=False)`, so they do not depend on the environment at all.
- **`tests/unit/test_render_env.py`** *(new)* — pins the pinning: that the hook ran before this module was imported, that the width survives any `TERM`, and that `FORCE_COLOR` is what would actually clamp it. A future edit that loosens the hook fails here with an explanation, instead of as a scatter of puzzling output-comparison failures across the ctl tests.
- **`tests/AGENTS.md`** — documents the invariant and the two traps below, so the next contributor does not rediscover them.

### Implementation notes

Three details drove the design, all verified against `rich` 13.9.4 / `click` 8.3.3 rather than assumed:

1. **It has to be `pytest_configure`, not a fixture.** Rich snapshots `no_color` in `Console.__init__`, and many `infrahub_sdk.ctl` modules build a module-level `Console()` at import time — which happens while pytest collects the test modules, before any fixture can run. A session-scoped autouse fixture is already too late for those consoles.
2. **`FORCE_COLOR` must be removed, not blanked.** Rich treats *any* `FORCE_COLOR` value as proof it is writing to a terminal, empty string included (`if force_color is not None: self._force_terminal = True`).
3. **`TERM=dumb` is deliberately *not* set.** It looks like the obvious way to disable colour, but it is the trigger for Rich's dumb-terminal path, which pins the width to 80 and ignores `COLUMNS` — truncating the very wide tables the CLI-output fixtures record (e.g. the `repository_list` fixture is 139 columns).

   To be precise about the mechanism, since it was raised in review: `TERM=dumb` alone is harmless here, because `is_dumb_terminal` is `is_terminal and TERM in ("dumb", "unknown")` and captured test output is never a terminal. It is removing `FORCE_COLOR` (point 2) that keeps `is_terminal` `False` and so defuses the dumb path. Pinning `TERM` to a non-dumb value would therefore be a no-op for width while implying a knob that does nothing — whereas *adopting* `TERM=dumb`, as the card originally suggested, is what creates the failure if anything ever does force a terminal. `test_render_env.py` pins both halves of this.

One central hook covers all 24 `CliRunner()` call sites, so no per-test env plumbing was needed — passing `env=` to each runner would have been 24 edits that drift, and (per point 1) would not have fixed the module-level consoles anyway.

### What stayed the same

No production code touched — the diff is `tests/` plus a changelog fragment. No API, schema, dependency, or CI-workflow changes.

## How to review

`tests/conftest.py` is the whole change; the rest follows from it. The comment block there records the three findings above.

## How to test

The card's own acceptance gate, plus the environments that reproduce the original reports:

```bash
uv run pytest tests/unit/                                   # bare environment
FORCE_COLOR=1 COLUMNS=40 uv run pytest tests/unit/          # forced colour, narrow
NO_COLOR=1 uv run pytest tests/unit/                        # colour disabled
TERM=dumb uv run pytest tests/unit/                         # dumb terminal
FORCE_COLOR=3 TERM=xterm-256color COLUMNS=40 uv run pytest tests/unit/
FORCE_COLOR= TERM=dumb uv run pytest tests/unit/            # empty FORCE_COLOR + dumb
uv run ruff check . && uv run ruff format --check .
```

**Before** (on `develop`, with `FORCE_COLOR=3` exported — the reported dev-machine case): **37 failed, 1781 passed**, spread across `tests/unit/ctl/test_marketplace_app.py` (31), `tests/unit/sdk/test_schema.py` (3), `tests/unit/ctl/test_repository_app.py` (2) and `tests/unit/ctl/test_task_app.py` (1).

**After:** **1827 passed, 1 xfailed** — identical under all six environments above. `ruff check`, `ruff format --check`, `ty`, `mypy` (158 files) and `rumdl` (129 files) are all clean; `invoke lint` reports only `vale is not installed` locally.

Two notes for the reviewer:

- The card also named `test_gitrepo_init` in `tests/unit/sdk/test_repository.py`. It does not fail on current `develop` in any of the environments tested, so nothing was changed for it.
- `test_marketplace_app.py`'s 31 failures were not in the card's list but share the exact same root cause and are fixed by the same hook.

## Impact & rollout

- **Backward compatibility:** test-only; no runtime behaviour changes.
- **Config/env changes:** the test process now sets `NO_COLOR`/`COLUMNS` and unsets `FORCE_COLOR` for itself. A side effect is that pytest's own report renders at 200 columns.
- **Deployment notes:** safe to merge; nothing to coordinate.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added
- [ ] External docs updated (if user-facing or ops-facing change) — n/a, test-only
- [x] Internal .md docs updated (`tests/AGENTS.md`)

---

Filed from Engineering Inbox card [INBOX-166](https://opsmill.atlassian.net/browse/INBOX-166) by the platform-health agent. Opened for human review — the agent does not merge.

[INBOX-166]: https://opsmill.atlassian.net/browse/INBOX-166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
## Summary by cubic
Pins Rich's colour and width for the test suite so CLI-output assertions behave the same everywhere, fixing 37 tests that failed on developer machines but passed in CI. Closes INBOX-166.

- `tests/conftest.py` unsets `FORCE_COLOR` and sets `NO_COLOR=1` and `COLUMNS=200` from `pytest_configure` before any test module is imported.
- `tests/unit/sdk/test_schema.py` now uses a shared `schema_console` fixture that patches the console with `no_color=True` and `force_terminal=False`.
- `tests/unit/meta/test_render_env.py` guards the pinning: the env is applied, the width survives any `TERM`, and `FORCE_COLOR` set with `TERM=dumb` is pinned as the failure mode the hook prevents, for every value including the empty string.

<sup>Written for commit 8d901e899ded1586f798be202feb3b6f9cc8aa08. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1332?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tests/conftest.py
```diff
@@ -11,6 +11,31 @@
...
+# imports the test modules -- before any fixture can run. So the environment has to be pinned
+# from a hook that runs ahead of collection, which is what ``pytest_configure`` does.
+#
+# Without this, a developer whose shell exports ``FORCE_COLOR`` (or whose terminal is narrower
+# than the CLI-output fixtures) gets ANSI escapes and truncated Rich tables in captured output,
+# and every test that asserts on CLI text fails locally while staying green in CI.
+#
+# ``TERM`` is deliberately left alone: ``TERM=dumb`` puts Rich in its dumb-terminal path, which
+# pins the width to 80 and ignores ``COLUMNS``, truncating the wide tables these fixtures record.
+RENDER_ENV = {"NO_COLOR": "1", "COLUMNS": "200"}
```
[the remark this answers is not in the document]
saltas888: Checked this against `rich` 13.9.4 — it's a false positive, but a fair question, so I've pinned the answer in tests rather than just asserting it here.

`TERM=dumb` alone does not clamp the width. Rich's dumb-terminal path is gated on *also* believing it is on a terminal:

```python
@property
def is_dumb_terminal(self) -> bool:
    _term = self._environ.get("TERM", "")
    is_dumb = _term.lower() in ("dumb", "unknown")
    return self.is_terminal and is_dumb   # <-- both required
```

Captured test output is never a terminal, and the one thing that would make Rich claim otherwise is `FORCE_COLOR` (Rich reads *any* value, `""` included, as proof of a terminal) — which is exactly what the hook removes. Measured under the pinned env:

| `TERM` | `is_terminal` | `is_dumb_terminal` | width |
|---|---|---|---|
| `dumb` | False | False | **200** |
| `unknown` | False | False | **200** |
| `xterm-256color` | False | False | **200** |
| *(unset)* | False | False | **200** |

The clamp needs `FORCE_COLOR` to be present: with `FORCE_COLOR=1` **and** `TERM=dumb`, `is_terminal` becomes True, `is_dumb_terminal` True, and the width drops to 80. So removing `FORCE_COLOR` is what defuses the dumb path — pinning `TERM` to a non-dumb value would be a no-op for width, while implying a knob that does nothing.

There is also a reason not to pin it: `TERM=dumb` was the card's own suggestion, and adopting it is what *creates* the failure. If anything ever does force a terminal, `TERM=dumb` gives width 80 and truncates the wide tables these fixtures record (the `repository_list` fixture is 139 columns), turning one class of failure into another. The comment at `tests/conftest.py:24-25` records that.

Confirmed end-to-end: `TERM=dumb uv run pytest tests/unit/` is green, as are a bare environment, `NO_COLOR=1`, `FORCE_COLOR=1 COLUMNS=40`, and `FORCE_COLOR=3 TERM=xterm-256color COLUMNS=40` — 1825 passed in each.

Addressed in 5c3756b, which adds `tests/unit/test_render_env.py`: it asserts the hook's env is applied, that the width survives `TERM` in `dumb`/`unknown`/`xterm-256color`/`screen`/empty, and pins the `FORCE_COLOR` + `TERM=dumb` combination as the failure mode the hook exists to prevent. Your concern is now a regression guard instead of a comment.

tests/unit/test_render_env.py
```diff
@@ -0,0 +1,62 @@
...
+
+def test_force_color_is_what_would_clamp_the_width(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Pin the failure mode the hook exists to prevent.
+
+    With ``FORCE_COLOR`` set, Rich treats the captured output as a terminal; combined with
+    ``TERM=dumb`` that drops it to width 80, which truncates the wide tables the CLI-output
+    fixtures record. This is the state a developer's shell puts the suite in, and the reason
+    ``FORCE_COLOR`` is removed rather than merely overridden.
+    """
+    monkeypatch.setenv("FORCE_COLOR", "1")
```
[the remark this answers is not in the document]
saltas888: Valid — fixed in caada89.

You're right that this was the one form the test never exercised, and it's the form that matters most: the whole reason the hook *removes* `FORCE_COLOR` instead of overriding it with a falsy value is that Rich only tests presence —

```python
force_color = self._environ.get("FORCE_COLOR")
if force_color is not None:
    self._force_terminal = True
```

— so `export FORCE_COLOR=` forces a terminal exactly as `1` does. Asserting that in a docstring while testing only `"1"` left the actual claim unpinned.

Now parameterized over `"1"`, `"3"` and `""` (`"3"` because that's what a real shell exports). All three reach `is_dumb_terminal is True` and clamp to width 80:

```
test_force_color_is_what_would_clamp_the_width[1]  PASSED
test_force_color_is_what_would_clamp_the_width[3]  PASSED
test_force_color_is_what_would_clamp_the_width[]   PASSED
```

Also added `FORCE_COLOR= TERM=dumb` to the environment matrix the suite is verified under, which is the end-to-end version of your case: **1827 passed** there, matching a bare environment, `NO_COLOR=1`, `TERM=dumb`, `FORCE_COLOR=1 COLUMNS=40` and `FORCE_COLOR=3 TERM=xterm-256color COLUMNS=40`.

changelog/+pin-rich-render-env-in-tests.housekeeping.md
```diff
@@ -0,0 +1 @@
+Pinned Rich's colour and width for the test suite from `pytest_configure` in `tests/conftest.py`, which unsets `FORCE_COLOR` and sets `NO_COLOR=1` and `COLUMNS=200` before any test module is imported. Previously a developer whose shell exported `FORCE_COLOR` got ANSI escapes and truncated Rich tables in captured CLI output, so 37 tests across `tests/unit/ctl/test_marketplace_app.py`, `test_repository_app.py`, `test_task_app.py` and `tests/unit/sdk/test_schema.py` failed locally while staying green in CI. Test-only change.
```
polmichel: In my opinion, this one is kind of long and too much detailed for a release notes 
Maybe something like?
_Pinned Rich's colour and width for the test suite from `pytest_configure`, so CLI-output assertions render identically regardless of the developer's `FORCE_COLOR`/`COLUMNS` environment._

tests/unit/meta/test_render_env.py
polmichel, line 1: This file sits at the root of the `unit` package. Since it tests the test infrastructure rather than any functionality, I'd move it into a `unit/meta` or `unit/helpers` new subpackage, along with `test_packaging_metadata.py`.

tests/AGENTS.md
```diff
@@ -59,8 +59,33 @@ def test_cli_command():
...
+
+Practical consequences:
+
+- A plain `CliRunner()` is fine; it inherits the pinned environment. Pass `env=` only to widen
+  `COLUMNS` beyond 200 for a specific test.
+- When a test builds its own `Console`, make it explicit —
+  `Console(file=StringIO(), width=1000, no_color=True, force_terminal=False)` — so it does not
+  depend on the ambient environment at all.
+- Prefer fixing the environment over loosening an assertion, so exact-output tests keep their
+  value.
```
polmichel, line 95: Nice 👍

tests/unit/sdk/test_schema.py
```diff
@@ -479,7 +483,9 @@ async def test_display_schema_load_errors_details_when_error_is_in_attribute_or_
         ]
     }
 
-    with mock.patch("infrahub_sdk.ctl.schema.console", Console(file=StringIO(), width=1000)) as console:
+    with mock.patch(
+        "infrahub_sdk.ctl.schema.console", Console(file=StringIO(), width=1000, no_color=True, force_terminal=False)
+    ) as console:
```
polmichel: It looks like this could be encapsulated in a dedicated reusable fixture which can be called at the test instantiation
```
 @pytest.fixture
  def schema_console() -> Generator[Console, None, None]:
      """Replace the module-level console in infrahub_sdk.ctl.schema with one that captures output."""
      console = Console(file=StringIO(), width=1000, no_color=True, force_terminal=False)
      with mock.patch("infrahub_sdk.ctl.schema.console", console):
          yield console
```

polmichel (approved):
Looks good to me. Optional comment regarding code & package maintainability & changelog readability

changelog/+pin-rich-render-env-in-tests.housekeeping.md
saltas888: Agreed — trimmed to your wording in 8d901e8:

> Pinned Rich's colour and width for the test suite from `pytest_configure`, so CLI-output assertions render identically regardless of the developer's `FORCE_COLOR`/`COLUMNS` environment.

The detail it was carrying (which test modules failed, and why) belongs in the commit message and on the Jira card, not in release notes — both still have it.

tests/unit/meta/test_render_env.py
saltas888: Good call — moved to `tests/unit/meta/test_render_env.py` in 8d901e8, with an empty `__init__.py` to match every other `tests/unit` subpackage. `tests/AGENTS.md` now records what the package is for, so the next infrastructure test lands there too.

One snag on the companion file: `test_packaging_metadata.py` is not on `develop`. It was added by #1277 (`956c830`) which is on `infrahub-develop`, not this PR's base — so there is nothing to move alongside it here. Whoever lands it on `develop` can drop it straight into `tests/unit/meta/`.

tests/unit/sdk/test_schema.py
saltas888: Done in 8d901e8 — added `schema_console` essentially as you wrote it, and pointed the three `display_schema_load_errors` tests at it:

```python
@pytest.fixture
def schema_console() -> Generator[Console, None, None]:
    console = Console(file=StringIO(), width=1000, no_color=True, force_terminal=False)
    with mock.patch("infrahub_sdk.ctl.schema.console", console):
        yield console
```

Each test now takes `schema_console: Console` alongside its `mock_get_node`, which also drops a level of indentation from all three bodies and leaves one place to change if the console shape moves again. Kept it local to the module since nothing else needs it yet; easy to lift into `tests/unit/sdk/conftest.py` when something does.

saltas888:
All three review suggestions applied in `8d901e8` — thanks @polmichel.

| Suggestion | Done |
|---|---|
| Changelog too detailed for release notes | Trimmed to your one-sentence version |
| `test_render_env.py` should sit in a `unit/meta` subpackage | Moved to `tests/unit/meta/`, empty `__init__.py` matching the other subpackages |
| Encapsulate the console patching in a reusable fixture | `schema_console` added, all three `display_schema_load_errors` tests now use it |

`tests/AGENTS.md` also now points at the fixture as the preferred shape and records what `tests/unit/meta` is for.

One thing I could not do: you suggested moving `test_packaging_metadata.py` into the new package too, but it is not on `develop` — it came in via #1277 (`956c830`), which is on `infrahub-develop`. Nothing to move here; whoever lands it on `develop` can drop it straight into `tests/unit/meta/`.

Suite still green at **1827 passed** under a bare environment, `FORCE_COLOR=1 COLUMNS=40`, `NO_COLOR=1`, `TERM=dumb`, and `FORCE_COLOR=3 TERM=xterm-256color COLUMNS=40`. `ruff`, `ty`, `mypy` (158 files) and `rumdl` (129 files) all clean.

I have left your approval in place rather than re-requesting review, since all three were flagged as optional and the changes are confined to test layout and the changelog line — but happy to get another pass if you would prefer one before this goes in.

polmichel (approved):

PR #1348, opened 2026-09-10
feat: parse the server error envelope onto the base exception classes

ogenstad:
# Why

Part of the error enrichment that was included in 1.10.0 the overall epic aims to allow the SDK to raise more specific errors when there are problems such as node not found instead of the generic `GraphQLError`

---

When Infrahub rejects a request, the SDK throws away almost everything the server said about why. A caller who needs to react to a specific failure has only the message text to go on, so the only way to branch on "this was a uniqueness violation" rather than "something failed" is to string-match prose that is free to change between releases. The SDK does this to itself too, matching the literal `"Expired Signature"` to decide whether a 401 is worth a silent token refresh.

**Goal:** carry the server's error envelope onto the exception, so `exc.code` is the thing you branch on.

**Non-goals:** this is issue 1 of the four in [the spec's PR split](dev/specs/ifc-3034-error-catalogue/tasks.md). It deliberately does **not** add generated per-code bindings (issue 3), re-root the not-found classes or touch the CLI error ladder (issue 2), or change any error message. No class anyone catches changes shape.

Part of IFC-3034 / IHS-293.

## What changed

**Behavioural:**

- `AuthenticationError` and `GraphQLError` now carry `code`, `http_status`, `extensions`, and `errors` from the server's response, via a new shared `ApiError` base. Callers can branch on the catalogue code instead of the message.
- The silent token refresh is decided by `TOKEN_EXPIRED` in the envelope, with the legacy `"Expired Signature"` string kept only for servers that predate the catalogue. That literal now appears once in the SDK, down from twice.
- **Fixes a crash:** a 401 whose body was valid JSON but not an object (a bare array, string, or `null`, as a proxy or gateway can return) raised `AttributeError` out of the client. It now surfaces as `AuthenticationError` like every other unreadable 401.
- **Rejected file uploads now raise `AuthenticationError` too.** The multipart path called `raise_for_status()` bare, so a 401 or 403 on an upload surfaced a raw `httpx.HTTPStatusError` while every other request raised `AuthenticationError`. Pre-existing rather than introduced here, but squarely inside FR-015's "every failure the SDK observes as HTTP 401 or 403". The silent refresh already covered this path; only the error type was inconsistent.

**Implementation:**

- `infrahub_sdk/exceptions.py` becomes a package: `base.py` (the hand-written hierarchy), `factory.py` (two raise-time factories), `__init__.py` (a star-import façade). The import path `infrahub_sdk.exceptions` is unchanged, and a committed snapshot test pins that nothing dropped out of it.
- Fifteen call sites that each hand-rolled `" | ".join(messages)` or built a `GraphQLError` now funnel through the factories, so envelope parsing lives in one place.
- Both factories are total. Any shape the parser cannot read degrades to the generic exception that call site raised anyway, never to an SDK error of our own. That is the property the `malformed_*` fixtures exist to pin.

**What stayed the same:** no message text changed, no class was removed from any MRO, and `ApiError` is inserted strictly between `Error` and the two classes, so every existing `except` and `isinstance` still matches exactly what it matched. The CLI error ladder is byte-for-byte untouched.

### Suggested review order

1. `infrahub_sdk/exceptions/base.py` and `factory.py` for the new shape.
2. `infrahub_sdk/client.py` `_should_refresh_token` for the crash fix, and the two `_execute_graphql_with_file` methods for the upload path.
3. The call sites in `object_store.py` / `file_handler.py`, which are all the same edit.
4. Tests last.

## How to review

Four things worth a sceptical eye:

1. **Exception types on unreadable 401/403 bodies changed.** A non-JSON body previously raised `JsonDecodeError` (or a raw `json.JSONDecodeError` from the object store and file handler); a body with no `errors` key raised `TypeError`. Both now raise `AuthenticationError`. FR-015 requires `AuthenticationError` for every 401/403, so this fixes a pre-existing deviation, but it is visible to external consumers and has a changelog fragment.
2. **`GraphQLError.errors` drops entries that are not dicts** and maps a non-list `errors` to `[]`, so the declared `Sequence[dict[str, Any]]` is true at runtime. The payload still reaches the reader through the exception message. Consequence worth knowing: for a malformed envelope `infrahubctl` now prints nothing and exits 1, where it previously printed the payload. Issue 2's T040 fixes the renderer.
3. **`errors` is no longer declared `list`.** It inherits `Sequence[dict[str, Any]]`, which removes an unsound mutable-attribute override that pyright flags. `is_unknown_field_error` and `print_graphql_errors` were widened to match. Runtime is unaffected; an external consumer only breaks if it annotates `list` or calls `.append()` on `exc.errors`.
4. **Six docstrings carry a `# noqa: DOC501`.** Ruff reads `raise make_error(...)` as raising something named `make_error` and asks the docstring to document a function rather than an exception; the docstrings already name the real types. There is no config-level suppression, so the rule keeps working everywhere else. The directive sits after the docstring's closing quotes because that is the only placement ruff honours for this rule, and `RUF100` reports it as unused if ruff ever resolves a call to its return type.

The second commit annotates six tasks in `tasks.md` where the implementation diverged from the plan, each with the reasoning. The one that matters for the next issue: `GraphQLError`'s optional `message` parameter is deferred to its first caller rather than added unused, so T035 adds it.

## How to test

```bash
uv run pytest tests/unit/ -q
uv run invoke docs-generate docs-validate
```

Expected: unit suite green, ruff / ty / mypy clean across 160 files, docs in sync.

On macOS, 11 pre-existing failures are environmental and unrelated: 8 in `tests/unit/pytest_plugin/` are `psutil.cpu_freq()` raising `SystemError` inside the testcontainers plugin, and 3 in `tests/unit/ctl/` are Rich table assertions sensitive to terminal width.

## Impact & rollout

- **Backward compatibility:** no `except` clause in the SDK or CLI loses coverage. External consumers catching `JsonDecodeError` or `TypeError` around a 401/403 must catch `AuthenticationError` instead; documented in `changelog/+error-catalogue-envelope.changed.md`.
- **Performance:** none. Parsing happens only on an error path that already decoded the body.
- **Deployment notes:** safe to deploy on its own. Per the spec's landing order, the SDK merges before Infrahub bumps its submodule pointer, but that ordering constraint applies to issues 3 and 4, not this one.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added (two fragments: one `changed`, one `fixed`)
- [ ] External docs updated (if user-facing or ops-facing change) — the catalogue topic page belongs to issue 3, once per-code bindings exist
- [x] Internal .md docs updated (internal knowledge and AI code tools knowledge)


infrahub_sdk/client.py
```diff
@@ -104,17 +106,40 @@ class ProcessRelationsNodeSync(TypedDict):
...
 ) -> Callable[..., Coroutine[Any, Any, httpx.Response]]:
     @wraps(func)
     async def wrapper(client: InfrahubClient, *args: Any, **kwargs: Any) -> httpx.Response:
         response = await func(client, *args, **kwargs)
-        if response.status_code == 401:
-            errors = response.json().get("errors", [])
-            if "Expired Signature" in [error.get("message") for error in errors]:
-                await client.login(refresh=True)
-                return await func(client, *args, **kwargs)
+        if response.status_code == 401 and _should_refresh_token(response=response):
```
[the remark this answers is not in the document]
ogenstad: Valid, fixed in bbba2337 - thanks, this is a real FR-015 gap ("`AuthenticationError` ... MUST remain the class raised for every failure the SDK observes as HTTP 401 or 403"). Both `_execute_graphql_with_file` methods now wrap `raise_for_status()` and convert 401/403, with the docstrings updated and four tests covering both statuses on both clients, plus one pinning that a 500 still reaches the caller as `httpx.HTTPStatusError`.

One correction to the diagnosis: the silent refresh already worked on this path. `_post_multipart` carries `@handle_relogin` / `@handle_relogin_sync` (client.py:1453 and 2435), so an expired token was refreshed and retried. The gap was only the error type on a 401/403 that is not a refreshable expired token.

Worth noting this was pre-existing rather than introduced here - the multipart path never converted - so it strictly belongs to the same requirement rather than to this diff. It was small and squarely within FR-015, so it seemed better to close it than to defer it.

pyproject.toml
```diff
@@ -368,6 +372,7 @@ ignore = [
     "D301",    # Use `r"""` if any backslashes in a docstring — conflicts with Typer's `\b` no-wrap marker in CLI command docstrings, which must stay a real escape (not a raw string)
     "D417",    # Missing argument description in the docstring
     "DOC201",  # `return` is not documented in docstring
+    "DOC501",  # Raised exception missing from docstring (false positives for `raise make_error(...)`, where the factory's own name is read as the exception type)
```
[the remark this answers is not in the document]
ogenstad: Agreed on the principle, fixed in bbba2337, though the suggested location would not have worked.

The false positives are not raised in `infrahub_sdk/exceptions/**` - that is where the factories are *defined*. DOC501 fires at the *call sites*, because ruff reads the callee name in `raise authentication_error_from_response(...)` as the exception type. Confirmed by running the rule in isolation: 6 hits in `client.py`, 2 in `object_store.py`, 1 in `file_handler.py`, and 0 anywhere under `exceptions/`.

So the ignore is now scoped to those three modules instead of the global list. Verified that DOC501 is still enforced elsewhere in the SDK.

tests/fixtures/error_catalogue/README.md
```diff
@@ -0,0 +1,22 @@
+# Error catalogue fixtures
+
+Two envelope shapes appear here:
+
+- **GraphQL** (`/graphql`), the catalogue envelope. Data failures arrive as HTTP 200 with an `errors`
+  array; authentication failures arrive as a real 401 or 403. `extensions.code` is a catalogue code
+  string.
```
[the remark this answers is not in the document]
ogenstad: Valid, fixed in bbba2337. The `graphql_integer_code.json` fixture does contradict the taxonomy as written, and `factory.py` already had the accurate version of this ("a pre-catalogue envelope puts an integer here"), so the README was the odd one out.

Rewritten to separate the two axes rather than to add a parenthetical: transport (GraphQL vs REST) and what `extensions.code` holds (string catalogue code vs pre-catalogue integer) vary independently, so a `graphql_*` filename is not a promise that the code is a string.

infrahub_sdk/exceptions/base.py
```diff
@@ -57,12 +93,47 @@ def __init__(self, url: str, timeout: int | None = None, message: str | None = N
...
+        query: str | None = None,
+        variables: dict | None = None,
+    ) -> None:
         self.query = query
         self.variables = variables
-        self.errors = errors
-        self.message = f"An error occurred while executing the GraphQL Query {self.query}, {self.errors}"
+        # The message keeps the payload verbatim so a shape we cannot read still reaches the reader.
+        self.message = f"An error occurred while executing the GraphQL Query {query}, {errors}"
+        self.errors = as_error_list(errors)
```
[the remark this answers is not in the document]
ogenstad: Not taking this one, though happy to be overruled.

`Sequence` does not mean immutable - `list` satisfies it, so the contract is already enforced. What a tuple would add is runtime write protection on an exception attribute, and the cost is a consumer-visible break: `exc.errors` has been a `list` in every released version, so `exc.errors + [extra]`, `.append(...)`, and `assert exc.errors == [...]` in downstream tests would all start failing. That is a worse trade than the hygiene gain, in a PR whose whole compatibility claim (FR-018) is that nothing an existing caller does stops working.

The mutability that mattered was the unsound part, and that is already gone: `GraphQLError` no longer redeclares `errors: list[...]` over the base `Sequence[...]`, which is the mutable-attribute covariant override pyright flags as `reportIncompatibleVariableOverride`. Callers now get the read-only intent from the annotation while the runtime type stays what it always was.

On the base default being `()` while the factory produces a list: both satisfy `Sequence[dict[str, Any]]`, and anything that iterates, indexes, or takes `len` works with either. Making the class default a list instead would mean a mutable shared class attribute, which is worse.

pyproject.toml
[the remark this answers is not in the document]
ogenstad: Follow-up: went narrower still, in 2370c75c. There is now **no config-level DOC501 suppression at all** - not global, not per-file. Each of the six docstrings that actually triggers one carries its own directive.

Placement is not obvious, so recording what I measured. Ruff anchors this diagnostic on the *docstring*, not the raise, so only one position works:

| form | result |
|---|---|
| `# noqa: DOC501` on the `def` line | no effect |
| `# noqa: DOC501` on the `raise` line | no effect |
| `"""  # noqa: DOC501` after the closing quotes | suppresses |

I also tested whether restructuring the factories as alternate constructors would dodge the rule honestly, rather than suppressing it. It does not: `raise AuthError.from_response(r)` is still reported, as `from_response`. Ruff takes the last attribute segment, so there is no spelling of a factory call that satisfies the rule.

Two things narrowing bought that the per-file ignore would not have:

1. **It found a real bug in my own diff.** Both multipart methods re-raise `httpx.HTTPStatusError` for a non-401/403 status and neither docstring said so. The per-file ignore was hiding that behind the false positives. Both now document it - so the rule earned its keep within an hour of being restored.
2. **These directives self-clean.** `RUF100` is enabled repo-wide, so if ruff ever resolves a call to its return type, the unused directives get flagged and removed. A stale per-file ignore would sit there silently forever.

pyproject.toml
```diff
@@ -415,6 +419,21 @@ max-complexity = 14
     "S105", # Possible hardcoded password assigned to a variable
 ]
 
+# `file_handler` and `object_store` raise from the exception factories in
+# `infrahub_sdk/exceptions/factory.py`. Ruff reads the factory's own name as the raised type, so it asks
+# for a docstring entry naming a function rather than an exception; the docstrings already name the real
+# types. Scoped per file so DOC501 keeps working everywhere else. `client.py` needs the same, and gets it
+# in its own entry below.
+"infrahub_sdk/file_handler.py" = ["DOC501"]
```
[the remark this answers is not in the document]
ogenstad: Valid, and thanks for catching it - the description had gone stale against its own diff.

Worth noting your auto-resolution is half right. `2370c75` fixed the *configuration*, not the description, so the sentence you flagged was still wrong after that commit. Now updated, and it no longer describes either intermediate state: there is no config-level DOC501 suppression at all, global or per-file. Six docstrings carry their own directive.

Also added the multipart upload fix to the description, which landed in `bbba233` after the body was first written.

infrahub_sdk/client.py
```diff
@@ -2332,10 +2358,7 @@ def execute_graphql(
                     raise
```
polmichel, line 2358: The lower block in `_execute_graphql_with_file` which is doing the same job is raising the error when the code is not one of 401, 403 or 404. This is not directly related to this PR but should we also raise here when for instance the server is sending some 5XX error codes?

infrahub_sdk/client.py
```diff
@@ -1344,10 +1369,7 @@ async def execute_graphql(
...
             except httpx.HTTPStatusError as exc:
                 if exc.response.status_code in {401, 403}:
-                    response = decode_json(response=exc.response)
-                    errors = response.get("errors", [])
-                    messages = [error.get("message") for error in errors]
-                    raise AuthenticationError(" | ".join(messages)) from exc
+                    raise authentication_error_from_response(response=exc.response) from exc
                 if exc.response.status_code == 404:
                     raise URLNotFoundError(url=url) from exc
```
polmichel, line 1374: Same question about 5XX codes here (cf https://github.com/opsmill/infrahub-sdk-python/pull/1348/changes#r4007600733)

infrahub_sdk/client.py
```diff
@@ -1357,7 +1379,7 @@ async def execute_graphql(
         response = decode_json(response=resp)
 
         if "errors" in response:
-            raise GraphQLError(errors=response["errors"], query=query, variables=variables)
+            raise graphql_error_from_response(errors=response["errors"], query=query, variables=variables)
```
polmichel, line 1382: There is another untouched code location using this constructor: `infrahub_sdk.testing.schemas.animal.SchemaAnimal.load_schema`. Does this test fixture need to be updated to align with the new error building behavior?

infrahub_sdk/file_handler.py
```diff
@@ -156,16 +157,18 @@ def handle_error_response(exc: httpx.HTTPStatusError) -> None:
             NodeNotFoundError: If the file/node is not found (404).
             httpx.HTTPStatusError: For other HTTP errors.
 
-        """
+        """  # noqa: DOC501  # raises via a factory, whose name ruff reads as the exception's
         if exc.response.status_code in {401, 403}:
-            response = exc.response.json()
-            errors = response.get("errors", [])
```
polmichel, line 162: Code location flag for https://github.com/opsmill/infrahub-sdk-python/pull/1348/changes#r4007885222

changelog/+error-catalogue-envelope.changed.md
```diff
@@ -0,0 +1,5 @@
+Authentication failures now surface the server's error envelope. `AuthenticationError` and `GraphQLError` share a new `ApiError` base carrying `code`, `http_status`, `extensions`, and `errors`, so a caller can branch on the server's catalogue code instead of matching on message text.
+
+A 401 or 403 whose body the SDK cannot read as an error envelope now raises `AuthenticationError` carrying the best reason available: the REST API's bare `detail` string where the body has one, and otherwise the plain status. Previously the same responses raised `JsonDecodeError` (or, from the object store and file handler, a raw `json.JSONDecodeError`) when the body was not JSON, and `TypeError` when the body was JSON but carried no `errors` key. Code that catches those types around `object_store`, `file_handler`, or a client request should catch `AuthenticationError` instead.
```
polmichel: Nitpick and maybe not really important but all the locations where not raising a `TypeError`, for instance https://github.com/opsmill/infrahub-sdk-python/pull/1348/changes#r4007867249 was raising an empty `AuthenticationError`

polmichel (approved):
LGTM! Way cleaner.
I've just asked a few questions which may lead to valid comments.

infrahub_sdk/client.py
ogenstad: I'm not really sure about the end result of this, but for now I think we'll have to wait a while. The reason is that we already have refactored parts of the error handling for 50x errors in the `stable` and `infrahub-develop` branch with regards to retry operations there's a variable `DEFAULT_RETRY_STATUS_CODES` which controls this in those branches. I'll have to consider how the `pog-error-catalogue-IFC-3034` branch needs to be rebased and what that should look like before that one gets merged into `infrahub-develop` because of this. Luckily that retry logic is something that's just managed internally so a refactoring shouldn't impact end users. But for now I'll leave this as it is.

infrahub_sdk/client.py
ogenstad: Same comment as above.

infrahub_sdk/client.py
ogenstad: Checked, and it doesn't need updating. `SchemaLoadResponse.errors` is annotated `dict` (`infrahub_sdk/schema/__init__.py:976`), so `[resp.errors]` is already a `list[dict]`. That passes through the new `as_error_list()` untouched, which means `exc.errors` and the message are identical before and after this PR.

infrahub_sdk/file_handler.py
ogenstad: Thanks for the pointer you're right, and the changelog wording is fixed in the other thread.

changelog/+error-catalogue-envelope.changed.md
ogenstad: Not a nitpick the changelog was wrong, and it's now corrected.

PR #1363, opened 2026-09-15
feat: unify and re-root the not-found classes under GraphQLError

ogenstad:
# Why

`NodeNotFoundError`, `BranchNotFoundError`, and `SchemaNotFoundError` each represent a failure the server has a catalogue code for, but they sat outside the `ApiError` tree — so there was no single clause that catches a lookup miss regardless of whether the server reported it, the SDK decided it, or a REST 404 produced it. A catalogued failure's message also led with the GraphQL query text rather than with what actually went wrong.

Goal: make the hierarchy a tree under `ApiError`, have the three not-found classes adopt their catalogue codes, and make a server-reported failure's message say which failure it was.

Non-goals: no generated per-code classes and no `infrahub_sdk/exceptions/catalogue.py`. The typed per-code exceptions, the `error_handling.mdx` topic page, and the Infrahub-side generator are separate changes.

## What changed

Behavioral changes:

- `except GraphQLError` now also catches a node, branch, or schema lookup miss, including client-side ones and the REST 404 the file handler converts. `NodeInvalidError` inherits it.
- `exc.errors`, `exc.query`, and `exc.variables` are readable on those classes instead of raising `AttributeError`, since they now go through the constructor that owns that state.
- A catalogued failure reports as `UNIQUENESS_VIOLATION: Node of kind TestPerson already has name 'John'` instead of embedding the query. An uncatalogued one keeps today's message byte for byte.
- `infrahubctl` reports a catalogued failure by its code instead of labelling every 401 "Authentication failure" or printing a bare error list.

Implementation notes:

- The three classes declare a `CODE` attribute. Nothing here reads it — Infrahub's generator parses it out of `base.py` to know which codes are already represented — so a test pins the three strings.
- The CLI branch keys on `exc.code is not None` rather than on a class, so no future re-rooting can shadow it.
- The message names only the *governing* error (the first, which determines the class raised). The full list stays on `exc.errors`, and the CLI renders it beneath.

What stayed the same:

- Nothing renamed or removed, no constructor loses a signature. A snapshot test pins that every name importable from `infrahub_sdk.exceptions` still is, and now that each still descends from `Error`.
- An uncatalogued failure's message and CLI rendering are unchanged.

## How to review

Start with `infrahub_sdk/exceptions/base.py` — the re-rooting and the `CODE` declarations are the whole compatibility surface. Then `infrahub_sdk/ctl/utils.py`, where ladder order matters: the lookup-miss branch has to sit above `GraphQLError` or re-rooting makes it unreachable.

Where I'd want scrutiny:

- **The auth message now carries the code too.** The requirement is written about a server-reported catalogued failure rather than about a transport, so I applied it to both factories. That changed two assertions which landed in #1348. If you read it as GraphQL-only, it's a one-line revert plus those two assertions.
- **`infrahub_sdk/transfer/importer/json.py`** is outside this change's nominal file set. Re-rooting broke it: `isinstance(result, GraphQLError)` now matches classes whose `errors` is `[]`, so `--continue-on-error` reported `NodeNotFoundError: []` with no reason attached. Now guarded on `result.errors`. It is the one fix here with no direct test — that loop is not unit-testable without a refactor I judged disproportionate.

## How to test

```bash
uv run pytest tests/unit/sdk/test_exceptions.py tests/unit/ctl/test_utils.py -v
uv run pytest tests/unit/ -q
uv run invoke docs-generate && uv run invoke docs-validate
```

Unit suite passes (1981). On macOS, 11 failures are pre-existing and environmental — a `psutil.cpu_freq()` crash in the `infrahub-testcontainers` plugin kills the inner pytester subprocess in `tests/unit/pytest_plugin/`, and three ctl tests wrap Rich table output at the local terminal width — and all reproduce identically on the base commit.

## Impact & rollout

- **Backward compatibility:** three deliberate broadenings, each with a changelog fragment — `except GraphQLError` catches more, described message text changes shape, and a server-reported adopted code now raises its own class so `except NodeNotFoundError` sees server-side misses too. Nothing renamed or removed.
- **Performance:** none.
- **Deployment notes:** safe on its own. Blocks the Infrahub-side generator, which needs the `CODE` declarations to exist.

## Found along the way, not fixed here: #1366

Reviewing the one line this PR touches in `transfer/importer/json.py` turned up an unrelated two-year-old bug. `execute_batches` never unpacks the `(node, result)` pair its batch yields, so `isinstance(result, Exception)` always tests a tuple and the error branch is unreachable: `infrahubctl load` records a failed import task as a success, and `--continue-on-error` is a no-op.

Fixing it here made the branch reachable and turned six `test_export_import` integration tests red, exposing an HTTP 500 from `update_optional_relationships` that those tests had been passing through for as long as the bug existed. That is a separate fault, likely server-side, and it should not hold an error-catalogue PR red, so both were reverted in `cbb31943` and written up in #1366 with the fix, a standalone repro, and regression tests.

The only thing left here from that detour is a defensive `err.get("message", err)` where the error branch read `err["message"]`. It guards code that cannot currently run, which is why it ships without a test.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added
- [ ] External docs updated — the topic page belongs to the follow-up that adds the classes it documents
- [x] Internal .md docs updated

---
Assisted-by: speckit-implement unversioned
Assisted-by: code-review unversioned
Assisted-by: opsmill-dev-commit 0.1.0
Assisted-by: opsmill-dev-pr 0.1.0

---
## Summary by cubic
Re-roots the not-found classes under `GraphQLError` and makes a server-reported `NODE_NOT_FOUND`, `BRANCH_NOT_FOUND`, or `SCHEMA_NOT_FOUND` raise the matching class instead of a generic `GraphQLError`, so one clause catches a lookup miss however it arose. Implements IHS-294.

**Behavior**
- `except GraphQLError` now also catches node, branch, and schema lookup misses, including client-side ones and the REST 404; `NodeInvalidError` inherits the broadening but clears `CODE`.
- A described failure's message names the code and the server's message unless the governing error said nothing, in which case the message is what the call site would otherwise produce; `UNDEFINED_ERROR` and errors with no code keep today's message byte-identical.
- Authentication failures now also prefix the reported code when described, which changed two existing test assertions.
- `infrahubctl` renders a described failure by its code, keeps the path when the server sent one, lists the remaining server errors beneath it, and escapes every printed message, including the fallback traceback.
- `exc.errors`, `exc.query`, `exc.variables`, and `exc.code` are readable on all three re-rooted classes.

**Notes**
- `code_names_the_failure()` is the test for "the server described this": `UNDEFINED_ERROR` is coded but names nothing, so it never shapes the message or the CLI headline.
- The three classes declare `CODE` attributes and gain `from_payload` classmethods, which the factory dispatches server payloads through; `query_groups` catches `NodeNotFoundError` with a legacy-message fallback for servers older than the catalogue.
- The `infrahubctl load` importer fix was reverted, so failed import tasks are still counted as successes; only the defensive `.get` guard remains.
- Nothing is renamed or removed; constructors keep their signatures, and the exceptions facade re-exports only names that have a caller.

<sup>Written for commit 9a4649ad33d506884069602a42628e9155ec15e1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub-sdk-python/pull/1363?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


infrahub_sdk/ctl/utils.py
```diff
@@ -55,6 +57,15 @@ def handle_exception(exc: Exception, console: Console, exit_code: int) -> NoRetu
     """Handle exception in a different fashion based on its type."""
     if isinstance(exc, typer.Exit):
         raise exc
+    if isinstance(exc, ApiError) and exc.code is not None:
+        # A failure the server named. Its message already carries the code and the governing error's
+        # words, which every branch below would either mislabel or replace with the raw error list.
+        # Keying on the code rather than on a class also keeps the branch immune to re-rooting.
+        console.print(f"[red]{escape(str(exc))}")
+        # The code names only the first error, so the rest still have to be rendered on their own.
+        if len(exc.errors) > 1:
```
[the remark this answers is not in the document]
ogenstad: Valid, and fixed in e9e13c3 — thank you. Confirmed against the ladder before changing anything:

```
before this PR: ['TestPersonCreate'] Node of kind TestPerson already has name 'John'
on 360aeca:     UNIQUENESS_VIOLATION: Node of kind TestPerson already has name 'John'
```

The `> 1` guard was aimed at not losing errors two onward, and I missed that the *first* error carries a `path` the coded line has nowhere to put.

I diverged from the literal suggestion, because applying `if exc.errors:` on its own has a side effect worth flagging: `print_graphql_errors` only formats nicely when an entry has **both** `message` and `path`, and otherwise dumps the raw entry. Authentication envelopes carry no `path`, and before this PR they never reached that renderer at all — they took the `Authentication failure:` branch. So the suggestion as written produces:

```
TOKEN_EXPIRED: Expired Signature
{'message': 'Expired Signature', 'extensions': {'code': 'TOKEN_EXPIRED', 'http_status': 401, 'data': {}}}
```

So the fix does two things instead:

1. Render the error list whenever `exc.errors` is non-empty (your point), and let the coded line name just the code — the list already carries the message, so nothing repeats.
2. `print_graphql_errors` falls back to an entry's `message` rather than the dict when there is no `path`.

Result:

```
UNIQUENESS_VIOLATION
['TestPersonCreate'] Node of kind TestPerson already has name 'John'

TOKEN_EXPIRED
Expired Signature

SCHEMA_NOT_FOUND
['A'] first failure
['B'] second failure
```

Path retained, no duplicated message, no raw dicts. The exception's own `message` is unchanged, so a library caller still reads `CODE: message`. Covered by three new cases in `tests/unit/ctl/test_utils.py`: the single-error path is kept, the governing message appears exactly once, and an envelope without a path does not render as a dict.

infrahub_sdk/ctl/utils.py
```diff
@@ -149,10 +173,16 @@ def print_graphql_errors(console: Console, errors: Sequence[dict[str, Any]], fal
...
+        if not isinstance(error, dict) or "message" not in error:
             console.print(f"[red]{escape(str(error))}")
+            continue
+        message = escape(str(error["message"]))
+        # The path names the operation that failed. An envelope without one still has its message,
+        # which beats rendering the whole entry as a dict.
+        if "path" in error:
+            console.print(f"[red]{escape(str(error['path']))} {message}")
+        else:
+            console.print(f"[red]{message}")
```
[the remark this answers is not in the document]
ogenstad: Valid, and fixed exactly as you suggested in d520ac3 — the message-only fallback now applies to the catalogued path alone, and `print_graphql_errors` is back to printing the raw entry.

Worth recording why this one mattered more than the confidence score suggests. The shape it breaks is not hypothetical: a GraphQL **validation** error fails before execution, so it carries `locations` and no `path` — precisely the case my fallback swallowed.

```
pre-PR:  {'message': "Cannot query field 'nope' on type 'Query'.", 'locations': [{'line': 1, 'column': 9}]}
on e9e13c3: Cannot query field 'nope' on type 'Query'.
```

The line and column pointing at the offending part of the query were dropped, which is the most useful thing in that entry.

It also broke a guarantee this change is supposed to make. The design note behind the catalogued branch says that because the branch only claims errors carrying a code, *every uncatalogued failure keeps today's rendering byte-identical*. Reaching into the shared renderer violated that, and I'd reached for it to avoid a raw dict on the catalogued authentication path — a catalogued problem solved in the wrong place.

So `print_catalogued_errors` now owns the message-only fallback, and a test pins that an uncatalogued entry without a path still renders whole, `locations` included. Both of your findings came from the same root cause on my side: I kept treating the coded line as if it could stand in for the entry, when it only ever names the failure.

infrahub_sdk/exceptions/__init__.py
```diff
@@ -7,7 +7,15 @@
...
-The raise-time factories stay importable by name, since what an end user catches is the classes.
+The raise-time factories and `code_names_the_failure` stay importable by name, since what an end user
+catches is the classes. Nothing else is re-exported: a name here is a stability promise, so it earns
+its place by having a caller rather than by being plausibly useful one day.
+
+Those three are imported below as `X as X`, which is not a typo. The alias is what marks a name as
+re-exported rather than merely imported, and is the spelling ruff's F401 accepts for an import
+nothing in this file uses; a plain `from .base import X` for a name outside `__all__` fails the lint.
+One statement each is isort's doing, which splits aliased imports apart unless `combine-as-imports`
+is set.
```
ajtmccarty: do we really need a comment describing the linting?

infrahub_sdk/exceptions/base.py
```diff
@@ -128,11 +148,16 @@ def __init__(
...
+        message: str | None = None,
     ) -> None:
         self.query = query
         self.variables = variables
-        # The message keeps the payload verbatim so a shape we cannot read still reaches the reader.
-        self.message = f"An error occurred while executing the GraphQL Query {query}, {errors}"
+        # The default keeps the payload verbatim so a shape we cannot read still reaches the reader.
+        # A described failure supplies a message naming the code instead, and carries no query text.
+        # Only `None` asks for the default: a subclass raised with a deliberately empty message keeps
+        # it, rather than having the GraphQL placeholder put words in its mouth.
```
ajtmccarty: Agreed, cut to the one thing the code cannot say, in `9a4649ad`:

```python
# `is not None` rather than `or`: an empty message is a deliberate one, not a request for
# the default.
default = f"An error occurred while executing the GraphQL Query {query}, {errors}"
self.message = message if message is not None else default
```

Of the four lines, three were restating the line beneath them or describing caller behaviour that is already documented in `factory.py`. The surviving one is worth keeping because `or` is the idiomatic spelling and every sibling class in this file uses it, so the narrower test reads like an oversight rather than the point.

I took the note as a pattern rather than a one-off and applied it to the three other comments in this branch with the same problem: the adopted-class block in `factory.py`, and two in `query_groups.py`. Net 20 lines of comment out, 6 in.

ajtmccarty (approved):

infrahub_sdk/exceptions/__init__.py
ogenstad: No I guess that's not needed. I just thought that those alias imports looked ugly, I've removed the comments.

